PR #86, opened 2017-11-08
Tests/met 141 tests metis authentication

stzanakis:


metis-authentication/metis-authentication-service/src/test/java/eu/europeana/metis/authentication/dao/TestPsqlMetisUserDao.java
```diff
@@ -0,0 +1,351 @@
...
+  public void cleanUp() {
+    Mockito.reset(session);
+    Mockito.reset(transaction);
+    Mockito.reset(query);
+  }
+
+  @Test
+  public void createMetisUser() {
+    when(sessionFactory.openSession()).thenReturn(session);
+    when(session.beginTransaction()).thenReturn(transaction);
```
helinek: I would move those to a separate method or even to prepare because they are called in the beginning of every test

stzanakis: when(session.beginTransaction()).thenReturn(transaction); is not in all methods.
I will move when(sessionFactory.openSession()).thenReturn(session); to an before method.

PR #96, opened 2018-01-04
Bug/met 179 do exact match of returned list of organizations from zoho

stzanakis:


metis-authentication/metis-authentication-service/src/main/java/eu/europeana/metis/authentication/dao/ZohoAccessClientDao.java
```diff
@@ -36,40 +44,79 @@ public ZohoAccessClientDao(String zohoBaseUrl, String zohoAuthenticationToken) {
...
+  }
+
+  private JsonNode findExactMatchOfOrganization(JsonNode jsonOrgizationsResponse,
+      String organizationName) {
+    if (jsonOrgizationsResponse.get(RESPONSE_STRING).get(RESULT_STRING) == null) {
+      return null;
+    }
+    JsonNode organizationJsonNode = jsonOrgizationsResponse.get(RESPONSE_STRING).get(RESULT_STRING)
+        .get(ACCOUNTS_MODULE_STRING).get(ROW_STRING).get(FIELDS_LABEL);
+    if (organizationJsonNode == null) {
```
jochen-vermeulen: I suggest to use 
jsonOrgizationsResponse.....get(ROW_STRING).isArray() instead of a null comparison to determine whether we are dealing with an object or an array.

metis-authentication/metis-authentication-service/src/main/java/eu/europeana/metis/authentication/dao/ZohoAccessClientDao.java
```diff
@@ -36,40 +44,79 @@ public ZohoAccessClientDao(String zohoBaseUrl, String zohoAuthenticationToken) {
...
-    if (jsonContactResponse.get(RESPONSE_STRING).get(RESULT_STRING) == null)
+    if (jsonContactResponse.get(RESPONSE_STRING).get(RESULT_STRING) == null) {
       return null;
-    return jsonContactResponse.get(RESPONSE_STRING).get(RESULT_STRING).get(CONTACTS_MODULE_STRING).get("row").get("FL");
+    }
+    return jsonContactResponse.get(RESPONSE_STRING).get(RESULT_STRING).get(CONTACTS_MODULE_STRING)
+        .get(ROW_STRING).get(FIELDS_LABEL);
   }
 
   public JsonNode getOrganizationByOrganizationName(String organizationName) throws IOException {
```
jochen-vermeulen: I suggest returning only the organization ID as this is the only thing that is needed and do the check for the presence of an organization role (see MetisUser.setAndCheckOrganizationIdFromJsonNode()) somewhere in this class instead.

PR #100, opened 2018-01-12
Feat/met 146 add support for plugin monitoring second integration

stzanakis:


metis-core/metis-core-common/src/main/java/eu/europeana/metis/core/workflow/plugins/OaipmhHarvestPlugin.java
```diff
@@ -32,6 +33,7 @@
   private String id;
   private PluginStatus pluginStatus = PluginStatus.INQUEUE;
   private static final PluginType pluginType = PluginType.OAIPMH_HARVEST;
```
jochen-vermeulen: Name of this constant should be all caps.

metis-core/metis-core-common/src/main/java/eu/europeana/metis/core/workflow/plugins/OaipmhHarvestPlugin.java
```diff
@@ -130,14 +133,18 @@ public void setExternalTaskId(long externalTaskId) {
...
   @Override
-  public void setExecutionRecordsStatistics(
-      ExecutionRecordsStatistics executionRecordsStatistics) {
-    this.executionRecordsStatistics = executionRecordsStatistics;
+  public void setExecutionProgress(
+      ExecutionProgress executionProgress) {
+    this.executionProgress = executionProgress;
+  }
+
+  public String getTopologyName() {
```
jochen-vermeulen: I don't think this method is used?

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/execution/WorkflowExecutor.java
```diff
@@ -148,13 +149,46 @@ private Date runMetisPlugin(AbstractMetisPlugin abstractMetisPlugin) {
...
         .execute(dpsClient, ecloudBaseUrl, ecloudProvider, workflowExecution.getEcloudDatasetId());
+
+    if (!abstractMetisPlugin.getPluginMetadata().isMocked()) {
+      return periodicCheckingLoop(sleepTime, abstractMetisPlugin);
+    } else {
+      return periodicCheckingLoopMocked(sleepTime, iterationsToFake, abstractMetisPlugin);
+    }
+  }
+
+  private Date periodicCheckingLoop(int sleepTime, AbstractMetisPlugin abstractMetisPlugin) {
```
jochen-vermeulen: This method is very similar to periodicCheckingLoopMocked... is there no way to combine these two?

metis-core/metis-core-common/src/main/java/eu/europeana/metis/core/workflow/plugins/AbstractMetisPlugin.java
```diff
@@ -18,8 +18,8 @@
 @JsonSubTypes({
     @JsonSubTypes.Type(value=OaipmhHarvestPlugin.class, name="OAIPMH_HARVEST"),
     @JsonSubTypes.Type(value=HTTPHarvestPlugin.class, name="HTTP_HARVEST"),
-    @JsonSubTypes.Type(value=DereferencePlugin.class, name="DEREFERENCE"),
-    @JsonSubTypes.Type(value=VoidMetisPlugin.class, name="VOID")
+    @JsonSubTypes.Type(value=EnrichmentPlugin.class, name="ENRICHMENT"),
+    @JsonSubTypes.Type(value=ValidationPlugin.class, name="VALIDATION")
 })
 public interface AbstractMetisPlugin {
```
jochen-vermeulen: Implementing classes are very similar! Maybe we can introduce an abstract class that implements most of the functionality? Also this interface could be renamed, but I am not sure if this is feasible (it is implemented in the metis-core-workflow-qa dependency).

metis-core/metis-core-common/src/main/java/eu/europeana/metis/core/workflow/plugins/AbstractMetisPluginMetadata.java
```diff
@@ -15,8 +15,8 @@
 @JsonSubTypes({
     @JsonSubTypes.Type(value=OaipmhHarvestPluginMetadata.class, name="OAIPMH_HARVEST"),
     @JsonSubTypes.Type(value=HTTPHarvestPluginMetadata.class, name="HTTP_HARVEST"),
-    @JsonSubTypes.Type(value=DereferencePluginMetadata.class, name="DEREFERENCE"),
-    @JsonSubTypes.Type(value=VoidMetisPluginMetadata.class, name="VOID")
+    @JsonSubTypes.Type(value=EnrichmentPluginMetadata.class, name="ENRICHMENT"),
+    @JsonSubTypes.Type(value=ValidationPluginMetadata.class, name="VALIDATION")
 })
 public interface AbstractMetisPluginMetadata {
```
jochen-vermeulen: Same comment as for class AbstractMetisPlugin: Implementing classes are very similar! Maybe we can introduce an abstract class that implements most of the functionality? Also this interface should be renamed, but I am not sure if this is feasible (it is implemented in the metis-core-workflow-qa dependency).

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/execution/WorkflowExecutor.java
stzanakis: I thought about that as well but the problem is the nature of the loops. I was going to capture in a separate method the code inside the loop but the problem is that the null values returned refer to Date and the taskState which is required for the external loop is TaskState so that is not possible. The other option is to complicate the loop in one method and take into consideration both scenarios, which I tried but I got complaints from Sonar of loop conditions complexity plus I felt it was less readable, so I decided readability over complexity. We can revise that if needed in the future.

metis-core/metis-core-common/src/main/java/eu/europeana/metis/core/workflow/plugins/OaipmhHarvestPlugin.java
stzanakis: It is required by the serialization to json. I will add a comment.

metis-core/metis-core-common/src/main/java/eu/europeana/metis/core/workflow/plugins/OaipmhHarvestPlugin.java
stzanakis: That indeed is true. Again json serialization is important, just checked and it works as before because the getter is in place.

metis-core/metis-core-common/src/main/java/eu/europeana/metis/core/workflow/plugins/AbstractMetisPlugin.java
stzanakis: Yes indeed I have in mind, just waiting until we have some more plugins ready to see how that looks and probably an abstract class is the way to go.

metis-core/metis-core-common/src/main/java/eu/europeana/metis/core/workflow/plugins/AbstractMetisPluginMetadata.java
stzanakis: Yes indeed I have in mind, just waiting until we have some more plugins ready to see how that looks and probably an abstract class is the way to go.

PR #112, opened 2018-02-06
Feat/met 475 create metis publishing implementation

SidSatbhai:
Contains an implementation of the indexing functionality and an example usage of the client for use in eCloud.


pom.xml
```diff
@@ -169,6 +169,7 @@
   <!--</reporting>-->
   <modules>
     <module>metis-core</module>
+    <module>metis-indexing</module>
```
stzanakis: This is duplicated, the build fails currently

metis-indexing/metis-indexing-client/src/main/java/eu/europeana/indexing/client/IndexingWorker.java
```diff
@@ -0,0 +1,119 @@
+package eu.europeana.indexing.client;
+
+import eu.europeana.indexing.client.IndexingClient;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class IndexingWorker {
+	private static final Logger LOGGER = LoggerFactory.getLogger(IndexingWorker.class);
+	
+	public static final String SAMPLE_INPUT_2 =
```
stzanakis: Hardcoded value should be removed

metis-indexing/metis-indexing-client/src/main/java/eu/europeana/indexing/client/config/Application.java
```diff
@@ -0,0 +1,180 @@
+/*
```
stzanakis: There shouldn't be a Spring Application Configuration file.

metis-indexing/metis-indexing-client/src/test/java/eu/europeana/indexing/client/test/IndexingClientTestWithMain.java
```diff
@@ -0,0 +1,121 @@
+package eu.europeana.indexing.client.test;
```
stzanakis: This has to be re-written to accommodate the respective class commented above.

metis-indexing/metis-indexing-client/src/main/java/eu/europeana/indexing/client/IndexingException.java
```diff
@@ -0,0 +1,9 @@
+package eu.europeana.indexing.client;
```
stzanakis: This Exception class can be moved to the metis-common module under the exception package

metis-indexing/metis-indexing-service/src/main/java/eu/europeana/indexing/model/IndexingMongoServer.java
```diff
@@ -0,0 +1,22 @@
+package eu.europeana.indexing.model;
```
stzanakis: Ok so I see that in the FullBeanDao the connection to the database is initialized.
The connection to the database should be initialized in a different class, maybe IndexingMongoServer. (Similar way to MorphiaDatastoreProvider, search on the project is in the metis-core)

And the FullBeanDao which is a DAO should be provided with a connection and that connection should be used for any DAO operation(DatasetDao).

metis-indexing/metis-indexing-service/src/main/java/eu/europeana/indexing/service/dao/FullBeanDao.java
```diff
@@ -0,0 +1,126 @@
+/*
```
stzanakis: Ok so I see that in the FullBeanDao the connection to the database is initialized.
The connection to the database should be initialized in a different class, maybe IndexingMongoServer. (Similar way to MorphiaDatastoreProvider, search on the project is in the metis-core)

And the FullBeanDao which is a DAO should be provided with a connection and that connection should be used for any DAO operation(DatasetDao).

metis-indexing/metis-indexing-service/src/test/java/eu/europeana/indexing/test/PublishingServiceTestWithMain.java
```diff
@@ -0,0 +1,161 @@
+package eu.europeana.indexing.test;
```
stzanakis: Same concept here. There shouldn't be a test with main class and the urls should not be public.
Sufficient testing can be done with embedded databases or mocks.

metis-indexing/metis-indexing-service/src/main/java/eu/europeana/indexing/service/PublishingService.java
```diff
@@ -0,0 +1,404 @@
...
+
+    public static String convertRDFtoString(RDF rdf) throws JiBXException, UnsupportedEncodingException {
+    	IMarshallingContext context = getRdfBindingFactory().createMarshallingContext();
+    	context.setIndent(2);
+    	ByteArrayOutputStream out  = new ByteArrayOutputStream();
+    	context.marshalDocument(rdf, UTF8, null, out);
+    	return out.toString(UTF8);
+    }
+    
+    private static void saveEdmClasses(FullBeanImpl fullBean, boolean isFirstSave, FullBeanDao fullBeanDao) throws NoSuchMethodException, IllegalAccessException, InvocationTargetException {
```
stzanakis: This method has a Cyclomatic complexity of 30 out of 15 allowed. It's ready huge and should be split into more constructed sub methods.

metis-indexing/metis-indexing-service/src/main/java/eu/europeana/indexing/service/PublishingService.java
```diff
@@ -0,0 +1,404 @@
...
+import org.jibx.runtime.IMarshallingContext;
+import org.jibx.runtime.IUnmarshallingContext;
+import org.jibx.runtime.JiBXException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+
+@Component
+public class PublishingService {
```
stzanakis: This quite a big class. Sonar gives me over 30 Critical and Major issues.

metis-indexing/metis-indexing-client/src/main/resources/indexing.properties.example
```diff
@@ -0,0 +1,12 @@
+mongo.host=host1,...,hostn
```
stzanakis: Please separate the blocks with comments, by grouping the configuration e.g. Mongo, Solr etc.

stzanakis:
Some general comments:
- The build currently fails so I couldn't run it because of that pom issue.
- The tests are not really junit tests, they just have a main. So basically current coverage of new code is 0% 
- Every new code should have as much as possible to 100% unit tests(either embedded dbs or mocked).
- Code published to github should not include sensitive information like domains/ip addresses.
- For new code we should add as much as possible javadoc but in this case this could potentially wait for the second phase where we will upgrade the versions of solr and mongo.
- Please try to resolve issues reported from Sonar, just opening a few files I saw lots of issues. This is old code sure but it should be cleaned with this story.

SidSatbhai:
Most of the issues mentioned after the initial code review have been addressed now.

Performing unit tests for the publishing service is proving to be tricky and time consuming using mocks because of errors (null pointers, etc.) emerging from corelib code that is being used.

The service and other functionality was tested earlier in a hard-coded fashion (using live systems).

metis-indexing/metis-indexing-service/src/main/java/eu/europeana/indexing/model/IndexingMongoServer.java
jochen-vermeulen: I don't think this class is used properly at all! I think it should just be removed as it is not necessary. See my comments in PublishingService (the only place where it is being used).

metis-indexing/metis-indexing-service/src/main/java/eu/europeana/indexing/service/PublishingService.java
```diff
@@ -0,0 +1,390 @@
...
+        fullBean.setTimespans(timespans);
+        fullBean.setLicenses(licenses);
+        fullBean.setServices(services);
+    }
+    
+    // Modified and appended legacy code
+    public static FullBeanImpl updateFullBean(FullBeanImpl fullBean, FullBeanDao fullBeanDao) throws NoSuchMethodException, IllegalAccessException, InvocationTargetException, MongoDBException {
+        saveEdmClasses(fullBean, false, fullBeanDao);
+        
+        IndexingMongoServer mongoServer = new IndexingMongoServer(fullBeanDao.getDS());
```
jochen-vermeulen: This object is never used but as a wrapper for a mongo datastore. I suggest removing the use of this object, instead work with the datastore directly. Or even better, not work on the datastore but create methods in FullBeanDaon to perform the tasks you want to perform here.

metis-indexing/metis-indexing-service/src/main/java/eu/europeana/indexing/service/dao/FullBeanDao.java
```diff
@@ -0,0 +1,74 @@
...
+    private final Datastore ds;
+
+    public FullBeanDao(MorphiaDatastoreProvider morphiaDatastoreProvider) {
+    	this.morphiaDatastoreProvider = morphiaDatastoreProvider;
+    	ds = getDS();
+    }
+
+    public  <T> T searchByAbout(Class<T> clazz, String about) {
+
+		return ds.find(clazz).filter("about", about).get();
```
jochen-vermeulen: Create a constant for the "about" string. It is used for the same purpose in multiple locations.

metis-indexing/metis-indexing-client/src/test/java/eu/europeana/indexing/client/test/PublishingServiceTest.java
```diff
@@ -0,0 +1,144 @@
+package eu.europeana.indexing.client.test;
```
jochen-vermeulen: Test classes are typically in the same package as the classes that they are supposed to test. In fact this class tests something in a completely different project! 

I think the solution here (as mentioned before) is to merge the two projects.

metis-indexing/metis-indexing-client/src/test/java/eu/europeana/indexing/client/test/PublishingServiceTest.java
```diff
@@ -0,0 +1,144 @@
...
+
+import org.apache.solr.client.solrj.impl.CloudSolrServer;
+import org.apache.solr.client.solrj.impl.LBHttpSolrServer;
+import org.junit.Before;
+import org.junit.Test;
+import org.mockito.Mockito;
+
+import com.mongodb.MongoClient;
+
+public class PublishingServiceTest {
```
jochen-vermeulen: This class doesn't seem to test much at all! 

Also, I am not a fan of testing large string inputs, I prefer to test more specifically constructed values that I can then verify. But I suppose that this could be seen as a matter of personal opinion.

metis-indexing/metis-indexing-service/src/main/java/eu/europeana/indexing/service/PublishingService.java
```diff
@@ -0,0 +1,390 @@
...
+    }
+
+    public PublishingService(FullBeanDao fullBeanDao, LBHttpSolrServer solrServer, CloudSolrServer cloudSolrServer) {
+  	  this.fullBeanDao = fullBeanDao;
+  	  this.solrServer = solrServer;
+  	  this.cloudSolrServer = cloudSolrServer;
+    }
+    
+    public boolean process(String record) throws IndexingException
+    {
```
jochen-vermeulen: Method is too complex (according to SonarQube... complexity of 16 instead of 15! ;)

metis-indexing/metis-indexing-service/src/main/java/eu/europeana/indexing/service/PublishingService.java
```diff
@@ -0,0 +1,390 @@
...
+    public static String convertRDFtoString(RDF rdf) throws JiBXException, UnsupportedEncodingException {
+    	IMarshallingContext context = getRdfBindingFactory().createMarshallingContext();
+    	context.setIndent(2);
+    	ByteArrayOutputStream out  = new ByteArrayOutputStream();
+    	context.marshalDocument(rdf, UTF8, null, out);
+    	return out.toString(UTF8);
+    }
+    
+    // Modified and appended legacy code
+    private static void saveEdmClasses(FullBeanImpl fullBean, boolean isFirstSave, FullBeanDao fullBeanDao) throws NoSuchMethodException, IllegalAccessException, InvocationTargetException, MongoDBException {
```
jochen-vermeulen: This method is way too long and complex. It should be split up.

metis-indexing/metis-indexing-client/src/test/java/eu/europeana/indexing/client/test/IndexingWorkerTestWithMain.java
```diff
@@ -0,0 +1,111 @@
+package eu.europeana.indexing.client.test;
+
+import eu.europeana.indexing.client.IndexingClient;
+import eu.europeana.indexing.client.IndexingWorker;
+import eu.europeana.metis.exception.IndexingException;
+
+public class IndexingWorkerTestWithMain {
```
jochen-vermeulen: This is not a proper test class (it has a main method).

metis-indexing/metis-indexing-client/src/main/resources/indexing.properties.example
jochen-vermeulen: This is not solved yet.

jochen-vermeulen: Does this mean that we need an entry in the .gitignore of this project for the file indexing.properties?

metis-indexing/metis-indexing-client/src/main/java/eu/europeana/indexing/client/config/IndexingConfig.java
```diff
@@ -0,0 +1,121 @@
...
+		try(InputStream resourceStream = loader.getResourceAsStream(resourceName)) {
+		    props.load(resourceStream);
+		} catch (IOException e) {
+			throw new IndexingException("Could not load indexing configuration.", e.getCause());
+		}
+
+		// Mongo
+		String mongoHost = props.getProperty("mongo.host");
+		String mongoPort = props.getProperty("mongo.port");
+		String mongoUsername = props.getProperty("mongo.usernaname");
```
jochen-vermeulen: This variable is never used! And neither are some other ones in the lines below!

metis-indexing/metis-indexing-client/src/main/java/eu/europeana/indexing/client/config/IndexingConfig.java
```diff
@@ -0,0 +1,121 @@
...
+import java.io.InputStream;
+import java.net.MalformedURLException;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Properties;
+import org.apache.commons.lang.StringUtils;
+import org.apache.solr.client.solrj.impl.CloudSolrServer;
+import org.apache.solr.client.solrj.impl.LBHttpSolrServer;
+
+public class IndexingConfig {
```
jochen-vermeulen: This class uses only code from the service project. It would seem that there is not much reason to have two different projects here (client vs service): there is no service, just client. I would suggest merging these projects into one indexing project.

metis-indexing/metis-indexing-client/src/main/java/eu/europeana/indexing/client/config/IndexingConfig.java
```diff
@@ -0,0 +1,121 @@
...
+	
+	public PublishingService getPublishingService() {
+		return publishingService;
+	}
+	
+	public FullBeanDao getFullBeanDao() {
+		return fullBeanDao;
+	}
+	
+	private void init() throws IndexingException {
```
jochen-vermeulen: As this is code that is likely to go wrong at some point (i.e. wrong configuration in config file) I think it would benefit from some logging.

metis-indexing/metis-indexing-client/src/main/java/eu/europeana/indexing/client/IndexingWorker.java
```diff
@@ -0,0 +1,45 @@
...
+	}
+	
+	 /**
+	   * Publishes a record
+	   * 
+	   * @param record The record to be published
+	   * @return true (success) or false (failure)
+	   * @throws IndexingException
+	   */
+	public boolean publish(String record) throws IndexingException {
```
jochen-vermeulen: All this entire class does is call the indexing client. Maybe integrate this method with the indexing client (which is also quite simple) and then remove the class.

metis-indexing/metis-indexing-client/src/main/java/eu/europeana/indexing/client/IndexingClient.java
```diff
@@ -0,0 +1,33 @@
...
+	    	    
+		fullBeanDao = config.getFullBeanDao();		
+		publishingService = config.getPublishingService();
+	}
+	
+	public boolean publish(String record) throws IndexingException {
+		return publishingService.process(record);
+	}
+	
+	public List<FullBeanImpl> getAll() {
```
jochen-vermeulen: This method is not used at all. I suggest removing it and then checking if this means that some code in the FullBeanDao can also go.

metis-indexing/metis-indexing-client/src/main/java/eu/europeana/indexing/client/IndexingClient.java
```diff
@@ -0,0 +1,33 @@
...
+		return fullBeanDao.getAll();
+	}
+	
+	public FullBeanImpl getFullBean(String id) {
```
jochen-vermeulen: This method is not used at all. I suggest removing it and then checking if this means that some code in the FullBeanDao can also go.

metis-indexing/metis-indexing-client/pom.xml
```diff
@@ -0,0 +1,60 @@
...
+            <version>${project.version}</version>
+        </dependency>
+                <dependency>
+            <groupId>org.apache.commons</groupId>
+            <artifactId>commons-lang3</artifactId>
+            <version>${version.commons.lang}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.springframework</groupId>
+            <artifactId>spring-web</artifactId>
```
jochen-vermeulen: I don't think spring is used in this project! There are three dependencies that should be removed here.

metis-indexing/metis-indexing-service/pom.xml
```diff
@@ -0,0 +1,50 @@
...
+            <version>${project.version}</version>
+        </dependency>
+                <dependency>
+            <groupId>org.apache.commons</groupId>
+            <artifactId>commons-lang3</artifactId>
+            <version>${version.commons.lang}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.springframework</groupId>
+            <artifactId>spring-web</artifactId>
```
jochen-vermeulen: I don't think this project uses spring! These three dependencies should be removed.

metis-indexing/metis-indexing-service/src/main/java/eu/europeana/indexing/service/dao/FullBeanDao.java
```diff
@@ -0,0 +1,74 @@
...
+import eu.europeana.corelib.solr.bean.impl.FullBeanImpl;
+import eu.europeana.indexing.service.mongo.MorphiaDatastoreProvider;
+import java.util.List;
+import org.mongodb.morphia.Datastore;
+import org.mongodb.morphia.query.Query;
+import org.mongodb.morphia.query.UpdateOperations;
+
+public class FullBeanDao {
+    private final MorphiaDatastoreProvider morphiaDatastoreProvider;
+    private final Datastore ds;
```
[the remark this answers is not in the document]
SidSatbhai: Agreed. Fixed.

metis-indexing/metis-indexing-service/src/main/java/eu/europeana/indexing/service/dao/FullBeanDao.java
```diff
@@ -0,0 +1,74 @@
...
+					.get();
+		}
+		return null;
+	}
+
+	public void save(FullBeanImpl fBean) {
+		ds.save(fBean);
+	}
+	
+	public Datastore getDS() {
```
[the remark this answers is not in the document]
SidSatbhai: Fixed.

metis-indexing/metis-indexing-client/src/main/resources/indexing.properties.example
SidSatbhai: Fixed.

metis-indexing/metis-indexing-service/src/main/java/eu/europeana/indexing/service/dao/FullBeanDao.java
SidSatbhai: Fixed.

metis-indexing/metis-indexing-client/src/test/java/eu/europeana/indexing/client/test/PublishingServiceTest.java
SidSatbhai: Agreed. Fixed.

metis-indexing/metis-indexing-client/src/test/java/eu/europeana/indexing/client/test/PublishingServiceTest.java
SidSatbhai: I prefer to do meaningful unit tests using mocks instead of superficial ones. In this case, usage of corelib is preventing the writing of meaningful unit tests with mocks. We'll have to decide as a team how we want to approach mocking.

metis-indexing/metis-indexing-client/src/test/java/eu/europeana/indexing/client/test/IndexingWorkerTestWithMain.java
SidSatbhai: As you've seen earlier with enrichment/dereferencing, the "Test with Mains" classes are used to run integration-type tests and test overall flows. They are not unit tests.

metis-indexing/metis-indexing-client/src/main/resources/indexing.properties.example
SidSatbhai: Correct. Fixed.

metis-indexing/metis-indexing-client/src/main/java/eu/europeana/indexing/client/config/IndexingConfig.java
SidSatbhai: Need to check with Simon which fields exactly are required in production and how exactly the init logic should work, e.g. when to use Cloud Solr and when not, etc.

metis-indexing/metis-indexing-client/src/main/java/eu/europeana/indexing/client/config/IndexingConfig.java
SidSatbhai: Agreed. Fixed.

metis-indexing/metis-indexing-client/src/main/java/eu/europeana/indexing/client/config/IndexingConfig.java
SidSatbhai: Fixed.

metis-indexing/metis-indexing-client/src/main/java/eu/europeana/indexing/client/IndexingWorker.java
SidSatbhai: Fixed.

metis-indexing/metis-indexing-client/src/main/java/eu/europeana/indexing/client/IndexingClient.java
SidSatbhai: This method might come in handy during further refactoring. Once we're completely done and if it's still not used, I'll toss it.

metis-indexing/metis-indexing-client/pom.xml
```diff
@@ -0,0 +1,51 @@
...
+    <version>1.0-SNAPSHOT</version>
+  </parent>
+  <modelVersion>4.0.0</modelVersion>
+
+  <artifactId>metis-indexing-client</artifactId>
+
+	<dependencies>
+    <dependency>
+            <groupId>eu.europeana.metis</groupId>
+            <artifactId>metis-indexing-service</artifactId>
```
jochen-vermeulen: Project does not build! This dependency no longer exists!!

metis-indexing/metis-indexing-client/src/main/java/eu/europeana/indexing/client/IndexingWorker.java
jochen-vermeulen: Can we then at least give them a constructor that needs no argument? Let's give them a very clean interface! That gives us more freedom to refactor in the future.

PR #116, opened 2018-02-15
Bug/met 611 fix certifactes problem when registering user

stzanakis:


metis-authentication/metis-authentication-service/src/main/java/eu/europeana/metis/authentication/service/AuthenticationService.java
```diff
@@ -164,15 +160,17 @@ private boolean isPasswordValid(MetisUser metisUser, String passwordToTry) {
    * </p>
    * @return a String array containing the email and password decoded, in that order
    * @throws GenericMetisException which can be one of:
-   * {@link BadContentException if the content of the authorization String is un-parsable
+   * <ul>
+   * <li>{@link UserUnauthorizedException if the content of the authorization String is un-parsable</li>
+   * </ul>
```
jochen-vermeulen: Missing closing bracket '}'

metis-authentication/metis-authentication-service/src/main/java/eu/europeana/metis/authentication/dao/ZohoAccessClientDao.java
```diff
@@ -141,8 +162,9 @@ private JsonNode findOrganizationFromListOfJsonNodes(JsonNode jsonOrgizationsRes
         JsonNode organizationField = organizationFields.next();
         JsonNode val = organizationField.get(VALUE_LABEL);
         JsonNode content = organizationField.get(CONTENT_LABEL);
-        if (val.textValue().equals(ORGANIZATION_NAME_FIELD) && content.textValue()
-            .equals(organizationName)) {
+        if (StringUtils.isNotEmpty(val.textValue()) && val.textValue()
+            .equals(ORGANIZATION_NAME_FIELD) && StringUtils.isNotEmpty(content.textValue())
+            && content.textValue().equals(organizationName)) {
```
jochen-vermeulen: Maybe use StringUtils.equals as this also handles null values: you don't have to use StringUtils.isNotEmpty that way and there will only be two calls instead of four in this if statement.

metis-authentication/metis-authentication-rest/src/main/java/eu/europeana/metis/authentication/rest/AuthenticationController.java
```diff
@@ -147,7 +153,7 @@ public void deleteUser(@RequestHeader("Authorization") String authorization,
       throw new UserUnauthorizedException("Action allowed only from admin users");
     }
     authenticationService.deleteUser(userEmailToDelete);
-    LOGGER.info("User with email: {} deleted", userEmailToDelete);
+    LOGGER.info("User with email: {} deleted", StringEscapeUtils.escapeJava(userEmailToDelete));
```
jochen-vermeulen: My eclipse complains that the expression is not done conditionally: (even if the log level is warn, the escape is performed).

Same in two log lines below.

metis-authentication/metis-authentication-service/src/main/java/eu/europeana/metis/authentication/service/AuthenticationService.java
stzanakis: Fixed

metis-authentication/metis-authentication-service/src/main/java/eu/europeana/metis/authentication/dao/ZohoAccessClientDao.java
stzanakis: Indeed. Fixed

metis-authentication/metis-authentication-rest/src/main/java/eu/europeana/metis/authentication/rest/AuthenticationController.java
stzanakis: Not changing as discussed, the impact is limited in this case.

PR #120, opened 2018-02-22
Feat/met 640 update metis to java 8

SidSatbhai:


pom.xml
```diff
@@ -109,6 +109,12 @@
           <groupId>org.apache.maven.plugins</groupId>
           <artifactId>maven-compiler-plugin</artifactId>
           <version>${version.maven.compiler.plugin}</version>
+          <configuration>
```
jochen-vermeulen: Remove the configuration tag (and all content) from the plugin declaration above (line 54)! This is now no longer needed.

Also, are you sure you need to specify this plugin in all poms if you already declare it in the parent pom?

SidSatbhai: This configuration tag is required in the parent POM to specify the compiler configuration. POMs further down the hierarchy can then indeed simply specify this plugin without configuration details (or they can modify configuration details at that level, i.e. if they want to use an older compiler version, etc.)

This is how the Maven documentation suggests working with compiler configurations.

jochen-vermeulen: But you are now specifying the java versions twice in the parent pom: they are found both at lines 54-59 and at lines 112-117.

SidSatbhai: I get what you're saying now. Fixed!

PR #122, opened 2018-02-22
Feat/met 577 integrate validation internal and add new fields for validation external

stzanakis:


metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/execution/WorkflowExecutor.java
```diff
@@ -73,6 +74,7 @@ public WorkflowExecution call() {
       lock.unlock();
       return workflowExecution;
     }
+    lock.unlock();
```
jochen-vermeulen: Would it be better to wrap the block above in a try ... finally so that the unlock statement is ALWAYS executed (even if an exception occurs)?
Then you can also remove the other unlock statement in line 74.

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/execution/WorkflowExecutor.java
```diff
@@ -178,17 +177,19 @@ private Date runMetisPlugin(AbstractMetisPlugin previousAbstractMetisPlugin,
       }
     }
 
+    final int MONITOR_ITERATIONS_TO_FAKE = 2;
```
jochen-vermeulen: This constant should go outside this method.

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/execution/WorkflowExecutor.java
```diff
@@ -178,17 +177,19 @@ private Date runMetisPlugin(AbstractMetisPlugin previousAbstractMetisPlugin,
...
+      return periodicCheckingLoopMocked(sleepTime, MONITOR_ITERATIONS_TO_FAKE, abstractMetisPlugin);
     }
   }
 
-  private Date periodicCheckingLoop(int sleepTime, AbstractMetisPlugin abstractMetisPlugin) {
+  private Date periodicCheckingLoop(long sleepTime, AbstractMetisPlugin abstractMetisPlugin) {
     TaskState taskState = null;
     int consecutiveMonitorFailures = 0;
-    int maxMonitorFailures = 3;
+    final int MAX_MONITOR_FAILURES = 3;
```
jochen-vermeulen: This constant should go outside this method.

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/execution/WorkflowExecutor.java
```diff
@@ -226,15 +228,17 @@ private Date periodicCheckingLoop(int sleepTime, AbstractMetisPlugin abstractMet
     return abstractMetisPlugin.getFinishedDate();
   }
 
-  private Date periodicCheckingLoopMocked(int sleepTime, int iterationsToFake,
+  private Date periodicCheckingLoopMocked(long sleepTime, int iterationsToFake,
       AbstractMetisPlugin abstractMetisPlugin) {
+    final int FAKE_RECORDS_PER_ITERATION = 100;
```
jochen-vermeulen: This constant should go outside this method to the top of the class.

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/execution/WorkflowExecutor.java
stzanakis: makes sense. fixed

PR #136, opened 2018-03-27
Feat/met 737 create library for media service

mmatela:
@stzanakis  @jochen-vermeulen


metis-media-service/src/main/java/eu/europeana/metis/mediaservice/MediaProcessor.java
```diff
@@ -0,0 +1,441 @@
...
+	}
+	
+	private void processText(String url, String mimeType, File contents) throws IOException {
+		if (!shouldExtractMetadata(urlTypes.get(url)))
+			return;
+		
+		boolean containsText = mimeType.startsWith("text/");
+		Integer resolution = null;
+		
+		if (mimeType.equals("application/pdf")) {
```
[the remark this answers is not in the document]
stzanakis: Also this if has too many nested flow statements if, for, try etc. We can split it down I think. 
Please refactor code to remove the label breaks as well.

metis-media-service/src/main/java/eu/europeana/metis/mediaservice/MediaProcessor.java
```diff
@@ -0,0 +1,441 @@
...
+			
+			// load colormap
+			try (InputStream is = Thread.currentThread().getContextClassLoader().getResourceAsStream("colormap.png")) {
+				colormapFile = File.createTempFile("colormap", ".png");
+				colormapFile.deleteOnExit();
+				try (FileOutputStream out = new FileOutputStream(colormapFile)) {
+					IOUtils.copy(is, out);
+				}
+			} catch (IOException e) {
+				throw new RuntimeException("colormap.png can't be loaded", e);
```
stzanakis: Use specific exception instead.

metis-media-service/src/main/java/eu/europeana/metis/mediaservice/MediaProcessor.java
```diff
@@ -0,0 +1,441 @@
...
+			}
+			
+			// find ffprobe
+			try {
+				List<String> lines = runCommand(Arrays.asList("ffprobe"), true);
+				if (lines.isEmpty() || !lines.get(0).startsWith("ffprobe version 2"))
+					throw new RuntimeException("ffprobe 2.x not found");
```
stzanakis: Use specific exception instead.

metis-media-service/src/main/java/eu/europeana/metis/mediaservice/MediaProcessor.java
```diff
@@ -0,0 +1,441 @@
...
+				ffprobeCmd = "ffprobe";
+			} catch (IOException e) {
+				throw new RuntimeException("Error while looking for ffprobe tools", e);
```
stzanakis: Use specific exception instead.

metis-media-service/src/main/java/eu/europeana/metis/mediaservice/MediaProcessor.java
```diff
@@ -0,0 +1,441 @@
...
+					for (String path : paths) {
+						lines = runCommand(Arrays.asList(path, "-version"), true);
+						if (!lines.isEmpty() && lines.get(0).startsWith("Version: ImageMagick 6")) {
+							magickCmd = path;
+							break;
+						}
+					}
+				}
+				if (magickCmd == null)
+					throw new RuntimeException("ImageMagick version 6/7 not found");
```
stzanakis: Use specific exception instead.

metis-media-service/src/main/java/eu/europeana/metis/mediaservice/MediaProcessor.java
```diff
@@ -0,0 +1,441 @@
...
+			} catch (IOException e) {
+				throw new RuntimeException("Error while looking for ImageMagick tools", e);
```
stzanakis: Use specific exception instead.

metis-media-service/src/main/java/eu/europeana/metis/mediaservice/MediaProcessor.java
```diff
@@ -0,0 +1,441 @@
...
+	 * @return if true, resources of given type don't need to be downloaded before
+	 *         processing.
+	 */
+	public static boolean supportsLinkProcessing(String mimeType) {
+		return isAudioVideo(mimeType);
+	}
+	
+	public static boolean isImage(String mimeType) {
+		// TODO pdf is not an image? https://europeana.atlassian.net/browse/MMS-34
+		return mimeType.equals("application/pdf") || mimeType.startsWith("image/");
```
stzanakis: Same here literal should optimally be on the left side.

metis-media-service/src/main/java/eu/europeana/metis/mediaservice/MediaProcessor.java
```diff
@@ -0,0 +1,441 @@
...
+import org.apache.pdfbox.text.PDFTextStripper;
+import org.apache.tika.Tika;
+import org.json.JSONObject;
+import org.json.JSONTokener;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class MediaProcessor implements Closeable {
+	
+	private static final Logger log = LoggerFactory.getLogger(MediaProcessor.class);
```
stzanakis: Rename to LOGGER please to follow same use on naming.

metis-media-service/src/main/java/eu/europeana/metis/mediaservice/MediaProcessor.java
```diff
@@ -0,0 +1,441 @@
...
+	private static final int[] THUMB_SIZE = { 200, 400 };
+	private static final String[] THUMB_SUFFIX = { "-MEDIUM", "-LARGE" };
+	
+	private static File tempDir;
+	private static File colormapFile;
+	private static String magickCmd;
+	private static String ffprobeCmd;
+	private static Tika tika;
+	
+	private ExecutorService commandIOThreadPool = Executors.newFixedThreadPool(2);
```
stzanakis: Number 2 can be a constant field.

metis-media-service/src/main/java/eu/europeana/metis/mediaservice/MediaProcessor.java
```diff
@@ -0,0 +1,441 @@
...
+	public void close() {
+		commandIOThreadPool.shutdown();
+	}
+	
+	private void processImage(String url, String mimeType, File content) throws MediaException, IOException {
+		List<File> thumbs = prepareThumbnailFiles(url, mimeType);
+		
+		int sizes = THUMB_SIZE.length;
+		ArrayList<String> command = new ArrayList<>(Arrays.asList(
+				magickCmd, content.getPath() + "[0]", "-format", "%w\n%h\n%[colorspace]\n", "-write", "info:"));
```
stzanakis: Create constants with all the literals, especially the ones that indicate parameter names and/or are used more than once. Like "-write" or "application/pdf" for example.

metis-media-service/src/main/java/eu/europeana/metis/mediaservice/MediaProcessor.java
```diff
@@ -0,0 +1,441 @@
...
+			width = Integer.parseInt(results.get(0));
+			int height = Integer.parseInt(results.get(1));
+			if (shouldExtractMetadata(urlTypes.get(url))) {
+				WebResource resource = edm.getWebResource(url);
+				resource.setMimeType(mimeType);
+				resource.setFileSize(content.length());
+				resource.setWidth(width);
+				resource.setHeight(height);
+				resource.setOrientation(width > height);
+				resource.setColorspace(results.get(2));
```
stzanakis: Constant for number 2.

metis-media-service/src/main/java/eu/europeana/metis/mediaservice/MediaProcessor.java
```diff
@@ -0,0 +1,441 @@
...
+				throw new IOException("Too many duplicates of url " + url);
+			}
+		}
+		return thumbs;
+	}
+	
+	private List<String> extractDominantColors(List<String> results) {
+		final Pattern pattern = Pattern.compile("#([0-9A-F]{6})");
+		return results.stream()
+				.skip(3)
```
stzanakis: Constants for number 3 and 6. It's easier to understand if the reader know what these numbers refer to and why they are used.

metis-media-service/src/main/java/eu/europeana/metis/mediaservice/MediaProcessor.java
```diff
@@ -0,0 +1,441 @@
...
+	private List<String> runCommand(List<String> command, boolean mergeError) throws IOException {
+		return runCommand(command, mergeError, null);
+	}
+	
+	private List<String> runCommand(List<String> command, boolean mergeError, byte[] inputBytes) throws IOException {
+		Process process = new ProcessBuilder(command).redirectErrorStream(mergeError).start();
+		if (!mergeError) {
+			commandIOThreadPool.execute(() -> {
+				try (InputStream errorStream = process.getErrorStream()) {
+					String output = IOUtils.toString(errorStream);
```
stzanakis: Deprecated toString method. Please use the toString that has as second parameter the string encoding.
e.g. IOUtils.toString(errorStream, "UTF-8");

metis-media-service/src/main/java/eu/europeana/metis/mediaservice/MediaProcessor.java
```diff
@@ -0,0 +1,441 @@
...
+			commandIOThreadPool.execute(() -> {
+				try (OutputStream processInput = process.getOutputStream()) {
+					processInput.write(inputBytes);
+				} catch (IOException e) {
+					log.error("Pushing data to process input stream failed for command " + command, e);
+				}
+			});
+		}
+		try (InputStream in = process.getInputStream()) {
+			return IOUtils.readLines(in);
```
stzanakis: Same for readLines, it's deprecated.

metis-media-service/src/main/java/eu/europeana/metis/mediaservice/MediaProcessor.java
```diff
@@ -0,0 +1,441 @@
...
+				throw new RuntimeException("Error while looking for ffprobe tools", e);
+			}
+			
+			// find image magick
+			try {
+				List<String> lines = runCommand(Arrays.asList("magick", "-version"), true);
+				if (!lines.isEmpty() && lines.get(0).startsWith("Version: ImageMagick 7")) {
+					magickCmd = "magick";
+				} else { // try convert, but careful about conflict with a windows tool
+					boolean isWindows = System.getProperty("os.name").toLowerCase().contains("win");
```
stzanakis: Specify Locale on lower case to be sure the conversion is correct?

metis-media-service/src/main/java/eu/europeana/metis/mediaservice/MediaProcessor.java
```diff
@@ -0,0 +1,441 @@
...
+	 * be removed. So duplicate thumbnails are stored in subdirectories with
+	 * increasing numbers.
+	 */
+	private List<File> prepareThumbnailFiles(String url, String mimeType) throws IOException {
+		List<File> thumbs = new ArrayList<>();
+		String md5 = DigestUtils.md5Hex(url);
+		String ext = Arrays.asList("application/pdf", "image/png").contains(mimeType) ? ".png" : ".jpeg";
+		for (int i = 0; i < THUMB_SUFFIX.length; i++) {
+			File temp = File.createTempFile("thumb", null);
+			for (int j = 0; j < 1000; j++) {
```
stzanakis: 1000 should be a constant.

metis-media-service/src/main/java/eu/europeana/metis/mediaservice/WebResource.java
```diff
@@ -0,0 +1,160 @@
...
+	
+	public void setColorspace(String colorspace) {
+		if (!"grayscale".equals(colorspace) && !"sRGB".equals(colorspace))
+			throw new IllegalArgumentException("Unrecognized color space: " + colorspace);
+		HasColorSpace hasColorSpace = new HasColorSpace();
+		hasColorSpace.setHasColorSpace("sRGB".equals(colorspace) ? ColorSpaceType.S_RGB : ColorSpaceType.GRAYSCALE);
+		resource.setHasColorSpace(hasColorSpace);
+	}
+	
+	public void setOrientation(boolean landscape) {
```
stzanakis: Create two methods instead of using the boolean as parameter. It is usually cleaner to the user of the method what the method does.

metis-media-service/src/main/java/eu/europeana/metis/mediaservice/WebResource.java
```diff
@@ -0,0 +1,160 @@
...
+	
+	public void setSampleRate(int sampleRate) {
+		resource.setSampleRate(intVal(new SampleRate(), sampleRate));
+	}
+	
+	public void setSampleSize(int sampleSize) {
+		resource.setSampleSize(intVal(new SampleSize(), sampleSize));
+	}
+	
+	public void setContainsText(boolean containsText) {
```
stzanakis: Set here where two methods would be cleaner than using one method with a boolean parameter.

metis-media-service/src/main/java/eu/europeana/metis/mediaservice/WebResource.java
```diff
@@ -0,0 +1,160 @@
...
+					hex.setDatatype("http://www.w3.org/2001/XMLSchema#hexBinary");
+					return hex;
+				})
+				.collect(Collectors.toList()));
+	}
+	
+	/** @param duration duration in seconds */
+	public void setDuration(double duration) {
+		Duration duration2 = new Duration();
+		duration2.setDuration(Integer.toString((int) Math.round(duration * 1000)));
```
stzanakis: TimeUnit can be used here or create constant of 1000.

mmatela:
Do we really have to blindly follow all these rules?

Some examples where I strongly disagree:
1. Upper case log variable - see arguments here: https://stackoverflow.com/questions/1417190/should-a-static-final-logger-be-declared-in-upper-case
2. serialVersionUID: this class is serialized to be sent between storm bolts. If two bolts in the same cluster use a different version of this class, it's actually dangerous and an exception should be thrown.
3. RuntimeException - in some cases it's disputable, but I try to use it when it indicates a problem in runtime environment: when "colormap.png" is not available, the jar was not built properly; if magick command is not found, it must be installed - otherwise there's no point running the topology. It's not something that should be caught in code and handled some way, so why create a custom exception?
4. String literals - in most cases constants would seriously decrease readability, unless the constant name was identical to literal contents - but then, what's the purpose?

About test coverage - if the library consisted only of simple POJO classes, would you insist on writing tests for all the getters and setters? That's an starting question to determine if there's any kind of threshold we can discuss :)
It's not that I'm inherently against automatic testing - I've worked on projects with very complex algorithms where there was several times more test code than working code and it was great. Compared to that, there is nothing going on here - it's just simple integration of external tools and libraries, so integration tests sound as the best approach.

stzanakis:
Ok so let me reply to your comment above.

**For your bullet points:**
1. This is a naming convention, very arguable(or personal preference) and I have seen both cases indeed, we just have the upper case almost everywhere that's why I asked to change it and follow the same convention in our project.
2. serialVersionUID, it's a common use on Serializable interface. If Storm requires this to not be present then we shouldn't add it of course, please communicate that with the other team members in PSNC since I am not very familiar with this kind of issue in Storm.
3. You are right these kind of exceptions are indeed unrecoverable exceptions, and can remain as you suggest although we will have to mark them in our analysis later on, there are some cases thought that we could use checked exceptions.
4. For String literals, having constants with the same name as the literal is more clear and easier to maintain, especially if it's used in multiple places.

**Code coverage**
Correct for trivial methods, like getters, setters, constructors we are also not writing tests. And yes more often than not test code is more than working code :D. I can give you, or you can see some examples in our already committed code, where something might seem trivial but is always a good thing to have. For example a Controller class, where everything that is used under the Controller is mocked.
e.g. link https://github.com/europeana/metis-framework/blob/develop/metis-core/metis-core-rest/src/test/java/eu/europeana/metis/core/rest/TestOrchestratorController.java
This is just to give you an idea on how we handle this type of tests, and I think the media service can highly benefit of tests like that, even if that means we'll need more time to write such tests.

And the last part but actually the first sentence on your comment :)

> Do we really have to blindly follow all these rules?  

Of course we don't blindly follow all these rules. Most of the times though they are correct and they are standards a community follows, which we try to do the same, even if sometimes they might seem like not really required(Btw rules sometimes change from the community and then we adapt to the new rules). So if I could put a percentage I would say 99% of the reported issues we try to solve, and 1% we mark them as invalid, if we have a good reason that we think it is an invalid reported issue, with a comment of the reason.

I hope this comment explains, a bit more clearer :), how and why we try to follow these kind of principles in our code.

mmatela:
To continue our back-and-forth:
1. I hardly see it as personal preference, the link I provided contains precise and highly up-voted explanations why the "upper case for static final fields" convention shouldn't apply to loggers. Only a few answers, much less popular, say the opposite. Do you have a source where the proportion of opinions is opposite?
2. It's not that storm requires it. I just think serialVersionUID in this case (as in many others) is an unnecessary clutter that can potentially cause more headache than it's supposed to save.
3. We have an agreement, yey :)
4. Seems like it requires a longer discussion and comparing concrete examples.

> Btw rules sometimes change from the community and then we adapt to the new rules

So who is this community? Do I have a vote? :)

> I would say 99% of the reported issues we try to solve

Depends on how you look at it, maybe I've already avoided 98% of potential issues by not writing sloppy code ;) The approach I'm familiar with is to treat sonar issues as merely suggestions - it's a reviewer's role to judge whether they make sense in each case. And if they don't, we just leave them in, they're not that distracting.

I'll apply suggestions that I agree with, but what shall we do about the rest? Do you want me to explain every case or will you do a second review and re-raise the issues you have strong opinions about? Do we do back-and-forth here or more directly via slack? And what if we don't reach agreement - will you force me to spoil my precious code? ;)

stzanakis:
ok I will try to make it short this time, it's Friday.. :)

1. The rule is here https://sbforge.org/sonar/rules/show/squid:S1312?layout=false. We have been following that rule for the code we handle.
2. Ok I thought it was Storm specific. Then here is the official javadoc explaining why the serialVersionUID should be explicitly defined. https://docs.oracle.com/javase/7/docs/api/java/io/Serializable.html. And here is the link to the rule
http://ci.ops4j.org/sonar/coding_rules#rule_key=squid%3AS2057
3. --
4. It's not a critical issue in general but in these cases we try to do similar thing to other places in our code base, personally I find it very useful.

> So who is this community? Do I have a vote? :)

The community is here if you want to contribute to the rules https://www.sonarqube.org/community/

For our review yes. Every case that you think is not applicable and the reason you think so should be commented at the corresponding location. Then a second review will take place either by the same or another developer.

>  And what if we don't reach agreement - will you force me to spoil my precious code?

xD
That's one of the reasons we try to follow the standards, except for specific cases that we have agreed upon and have a strong argument, that should explicitly not follow the rule.

metis-media-service/src/main/java/eu/europeana/metis/mediaservice/WebResource.java
mmatela: This rule doesn't make sense here. `setContainsText(true)` and `setContainsText(false)` are obvious to read. Furthermore, where I call it I already have a boolean variable and I don't want to introduce unnecessary if-else.

metis-media-service/src/main/java/eu/europeana/metis/mediaservice/MediaProcessor.java
```diff
@@ -0,0 +1,441 @@
...
+		}
+		
+		if (!mimeType.equals(providedMimeType))
+			log.info("Invalid mime type provided (should be {}, was {}): {}", mimeType, providedMimeType, url);
+		if (contents == null && !supportsLinkProcessing(mimeType))
+			throw new IllegalArgumentException("Contents file is required for mime type " + mimeType);
+		
+		try {
+			if (isImage(mimeType))
+				processImage(url, mimeType, contents);
```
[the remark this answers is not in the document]
mmatela: No, it's checked earlier: for images and texts `supportsLinkProcessing()` will return false and an `IllegalArgumentException` is thrown.

metis-media-service/src/main/java/eu/europeana/metis/mediaservice/EdmObject.java
```diff
@@ -0,0 +1,130 @@
...
+	
+	static {
+		try {
+			rdfBindingFactory = BindingDirectory.getFactory(RDF.class);
+		} catch (JiBXException e) {
+			throw new RuntimeException("Unable to create binding factory", e);
+		}
+	}
+	
+	private final RDF rdf;
```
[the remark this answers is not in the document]
mmatela: Since `RDF` is not serializable, I need to look up and use other serialization methods with Storm. This renders the serialVersionUID discussion irrelevant :)

metis-media-service/src/main/java/eu/europeana/metis/mediaservice/EdmObject.java
```diff
@@ -0,0 +1,130 @@
...
+	
+	private EdmObject(RDF rdf) {
+		this.rdf = rdf;
+	}
+	
+	public Map<String, List<UrlType>> getResourceUrls(Collection<UrlType> urlTypes) {
```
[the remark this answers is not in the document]
mmatela: Sonar would shut up if I turned the internal `for` into `stream().foreach()`. But I don't think it would improve readability.

metis-media-service/src/main/java/eu/europeana/metis/mediaservice/MediaProcessor.java
mmatela: If "-write" happened to be one character shorter or used one time less, this rule would not trigger.
It wouldn't make sense to make this literal stand out by turning it into a constant. And adding a dozen constants would look even worse.

metis-media-service/src/main/java/eu/europeana/metis/mediaservice/MediaProcessor.java
mmatela: Have it your way, but it looks so ugly it makes me want to avoid logging stuff as much as possible :P

metis-media-service/src/main/java/eu/europeana/metis/mediaservice/MediaProcessor.java
mmatela: I'd prefer to keep the list of fields clean. Defining that 2 is the THREAD_COUNT would not make anything clearer.

mmatela:
I've finished fixing sonar warnings, replied to specific comments where relevant. Time to look at junit testing possibilities.

metis-media-service/src/main/java/eu/europeana/metis/mediaservice/EdmObject.java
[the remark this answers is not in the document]
jochen-vermeulen: How about making the inside of the for-loop a separate method? Or find some other way to split this method?

metis-media-service/src/main/java/eu/europeana/metis/mediaservice/MediaProcessor.java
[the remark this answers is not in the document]
jochen-vermeulen: OK, but suppose the method supportsLinkProcessing() should change, then suddenly and quite unexpectedly we have this problem. So why not do some simple sanity checking at the beginning of each of the other two methods (isImage and processText)? Something as simple as
if (content == null) { throw new IllegalStateException(<<proper explanation>>); }
would already be nice, but maybe it is not that difficult to make it actually deal nicely with null content?

metis-media-service/src/main/java/eu/europeana/metis/mediaservice/MediaProcessor.java
```diff
@@ -0,0 +1,466 @@
...
+		}
+		try (InputStream in = process.getInputStream()) {
+			return IOUtils.readLines(in, Charset.defaultCharset());
+		}
+	}
+	
+	/**
+	 * This is a non-static method because it needs access to
+	 * {@link #runCommand(List, boolean)}.
+	 */
```
jochen-vermeulen: Why not make it static and pass an instance of MediaProcessor as argument? Seems a lot less hacky! Or even access to a thread pool, because that is the only thing why it needs an instance of MediaProcessor.

metis-media-service/src/main/java/eu/europeana/metis/mediaservice/MediaProcessor.java
```diff
@@ -0,0 +1,441 @@
...
+				.limit(6)
+				.map(line -> {
+					Matcher m = pattern.matcher(line);
+					m.find();
+					return m.group(1); // throw exception if not found
+				})
+				.collect(Collectors.toList());
+	}
+	
+	private void processAudioVideo(String url, String mimeType, File contents) throws MediaException, IOException {
```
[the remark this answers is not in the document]
jochen-vermeulen: I agree. It would seem this class has at least three responsibilities:

1. Maintaining a thread pool so that tasks can be executed (runCommand methods)
2. Providing utils-like functionality for the management of media files (e.g. prepareThumbnailFiles, PdfListener, but also parts of the process.... methods, for instance the part that extracts dominant colors from images)
3. Analyzing EDM WebResource objects and adding technical data to them.

Is there no way to split this up?

metis-media-service/src/main/java/eu/europeana/metis/mediaservice/EdmObject.java
```diff
@@ -0,0 +1,130 @@
...
+import org.jibx.runtime.IMarshallingContext;
+import org.jibx.runtime.IUnmarshallingContext;
+import org.jibx.runtime.JiBXException;
+
+import eu.europeana.corelib.definitions.jibx.Aggregation;
+import eu.europeana.corelib.definitions.jibx.HasView;
+import eu.europeana.corelib.definitions.jibx.RDF;
+import eu.europeana.corelib.definitions.jibx.WebResourceType;
+
+public final class EdmObject {
```
jochen-vermeulen: Could you provide some more documentation in this class? At least the object itself and the getResourceUrls method deserve some documentation.

metis-media-service/src/main/java/eu/europeana/metis/mediaservice/WebResource.java
```diff
@@ -0,0 +1,166 @@
...
+import eu.europeana.corelib.definitions.jibx.OrientationType;
+import eu.europeana.corelib.definitions.jibx.SampleRate;
+import eu.europeana.corelib.definitions.jibx.SampleSize;
+import eu.europeana.corelib.definitions.jibx.SpatialResolution;
+import eu.europeana.corelib.definitions.jibx.StringType;
+import eu.europeana.corelib.definitions.jibx.Type1;
+import eu.europeana.corelib.definitions.jibx.WebResourceType;
+import eu.europeana.corelib.definitions.jibx.Width;
+
+public class WebResource {
```
jochen-vermeulen: I think some documentation in this class would be nice? Just explaining why this class is here (instead of just using WebResourceType directly).

metis-media-service/src/main/java/eu/europeana/metis/mediaservice/MediaProcessor.java
```diff
@@ -0,0 +1,466 @@
...
+
+public class MediaProcessor implements Closeable {
+	
+	private static final Logger LOGGER = LoggerFactory.getLogger(MediaProcessor.class);
+	
+	private static final int[] THUMB_SIZE = { 200, 400 };
+	private static final String[] THUMB_SUFFIX = { "-MEDIUM", "-LARGE" };
+	
+	private static final File tempDir = new File(System.getProperty("java.io.tmpdir"));
+	@VisibleForTesting
```
jochen-vermeulen: Can't you make it private and provide a getter for testing? I think that would be much nicer. Same for the other three properties that are marked with this annotation. That way these four properties cannot be changed by outside code.

metis-media-service/src/main/java/eu/europeana/metis/mediaservice/EdmObject.java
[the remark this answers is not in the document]
mmatela: Come on, it's just an arbitrary metric scoring one point above an arbitrary limit :P
The method is much more readable than the "Cognitive Complexity" suggests, because all the if-statements follow the same pattern.

PR #140, opened 2018-04-03
Feat/met 455 retrieve dynamic dataset level information

stzanakis:


SidSatbhai:
I've reviewed the code. Apart from several methods which are very long, everything looks fine. Regarding method length - are we following any specific rules?

PR #144, opened 2018-04-04
MET-675: Updated ZipService for PreviewController

SidSatbhai:


metis-preview/metis-preview-service/src/main/java/eu/europeana/metis/preview/service/ZipService.java
```diff
@@ -25,32 +24,31 @@
...
-  public List<String> readFileToStringList(InputStream inputStream) throws ZipFileException {
+  public List<String> readFileToStringList(MultipartFile file) throws ZipFileException {
 
-    //TODO use a different zip library based on streams so we don't have to create files
-    //and delete them afterward
-    List<String> xmls = new ArrayList<>();
+    List<String> records = new ArrayList<>();
+    
     try {
-      String prefix = String.valueOf(new Date().getTime());
```
stzanakis: Why was this part modified? I understand that it was probably copied from validation module but here it was actually cleaner by using the default implementation of java for temp files.
Also the logging statements were removed. Please revert the handling of temp files to the previous version.

metis-preview/metis-preview-service/src/main/java/eu/europeana/metis/preview/service/ZipService.java
```diff
@@ -25,32 +24,31 @@
...
+        FileUtils.deleteQuietly(new File(fileName + ".zip"));
+        File[] files = new File(fileName).listFiles();
+        
+        for (File input : files) {
+            if(!input.isDirectory()){
+                FileInputStream stream = new FileInputStream(input);
+                records.add(IOUtils.toString(stream));
+                stream.close();
+            }
+        }
```
stzanakis: The unzippedFiles are not removed as it was performed previously.

metis-preview/metis-preview-service/src/test/java/eu/europeana/metis/preview/service/ZipServiceTest.java
```diff
@@ -16,24 +25,37 @@
   @Test
   public void readFileToStringList_withOkzipfile_returnsListOfStrings() throws Exception {
       ZipService service = new ZipService();
-      Resource resource = new ClassPathResource("ValidExternalOk.zip");
```
stzanakis: Why these changes on the methods of this class were needed?

metis-preview/metis-preview-service/src/main/java/eu/europeana/metis/preview/service/ZipService.java
SidSatbhai: The files are simply created in another manner now, so the prefix is not required. The missing logging statements have been fixed.

metis-preview/metis-preview-service/src/main/java/eu/europeana/metis/preview/service/ZipService.java
SidSatbhai: See line 40 in ZipService: FileUtils.deleteQuietly(new File(fileName + ".zip"));

metis-preview/metis-preview-service/src/test/java/eu/europeana/metis/preview/service/ZipServiceTest.java
SidSatbhai: This is because the readFileToStringList method in ZipService has been modified to take a MultipartFile as an input parameter instead of an InputStream.

metis-preview/metis-preview-service/src/main/java/eu/europeana/metis/preview/service/ZipService.java
stzanakis: Why do you think this is a better approach? There is a default implementation from java, we don't have to reinvent it if possible. Also the methods from java io have the benefit of avoiding same file name collisions.

metis-preview/metis-preview-service/src/main/java/eu/europeana/metis/preview/service/ZipService.java
stzanakis: The line you are referring to deletes the .zip file and not the unzipped files in the new directory created.

metis-preview/metis-preview-service/src/test/java/eu/europeana/metis/preview/service/ZipServiceTest.java
stzanakis: Sure that I understand, but I meant why not use same code as before for getting the resource e.g.:
Resource resource = new ClassPathResource("ValidExternalOk.zip");
resource.getInputStream()

Just seems cleaner like that.
In any case not very important.

metis-preview/metis-preview-service/src/main/java/eu/europeana/metis/preview/service/ZipService.java
SidSatbhai: I agree with what you're saying. I've modified the implementation to work with the standard Java IO classes again. We should consider changing this in the validation service too.

metis-preview/metis-preview-service/src/main/java/eu/europeana/metis/preview/service/ZipService.java
SidSatbhai: My bad. I've fixed this.

metis-preview/metis-preview-service/src/test/java/eu/europeana/metis/preview/service/ZipServiceTest.java
SidSatbhai: I agree. I've cleaned up the tests.

metis-preview/metis-preview-service/src/main/java/eu/europeana/metis/preview/service/ZipService.java
stzanakis: Yes indeed we  should modify that part too on another ticket though.

PR #158, opened 2018-04-26
Increased code coverage for enrichment module

SidSatbhai:


metis-enrichment/metis-enrichment-client/src/test/java/eu/europeana/enrichment/rest/client/EnrichmentWorkerTest.java
```diff
@@ -83,13 +83,23 @@
...
+		  testEnrichmentWorkerHappyFlow(mode);
+	  }
+	  
+	  Configurator.setLevel(logger.getName(), Level.INFO);
+	  for (Mode mode : Mode.values()) {
+		  testEnrichmentWorkerHappyFlow(mode);
+	  }
+	  	  
+	  Configurator.setLevel(logger.getName(), origLevel);
   }
```
jochen-vermeulen: I don't think we should do the exact same tests for different log levels unless we also check the logging behavior somewhere in this method. This means prescribing what logging should take place at different levels. 

However, I think that this goes a bit too far and the method should be reverted to it's original state. Any thoughts?

metis-enrichment/metis-enrichment-client/src/test/java/eu/europeana/enrichment/rest/client/EnrichmentWorkerTest.java
```diff
@@ -35,10 +39,6 @@
 import eu.europeana.enrichment.utils.EntityMergeEngine;
 import eu.europeana.enrichment.utils.InputValue;
 
-// TODO Tests that could be added:
-// 1. What if there is nothing to process (empty or null lists are returned when extracting or when
-// enriching/dereferencing).
-// 2. Throwing exceptions
 @RunWith(MockitoJUnitRunner.class)
```
jochen-vermeulen: This class is deprecated. Maybe switch to the alternative suggested in that class's javadoc.

metis-enrichment/metis-enrichment-client/src/test/java/eu/europeana/enrichment/rest/client/EnrichmentWorkerTest.java
```diff
@@ -35,10 +39,6 @@
 import eu.europeana.enrichment.utils.EntityMergeEngine;
 import eu.europeana.enrichment.utils.InputValue;
 
-// TODO Tests that could be added:
-// 1. What if there is nothing to process (empty or null lists are returned when extracting or when
-// enriching/dereferencing).
-// 2. Throwing exceptions
```
jochen-vermeulen: Why were these comments removed? I don't think the suggestions have been followed, but maybe there is another reason this is no longer necessary?

metis-enrichment/metis-enrichment-client/pom.xml
```diff
@@ -54,16 +54,7 @@
...
-		<dependency>
-		    <groupId>org.slf4j</groupId>
-		    <artifactId>slf4j-simple</artifactId>
-		    <version>${version.slf4j}</version>
-		</dependency>
-		<dependency>
-		    <groupId>org.slf4j</groupId>
-		    <artifactId>slf4j-jdk14</artifactId>
-		    <version>${version.slf4j}</version>
-		</dependency>
```
jochen-vermeulen: We shouldn't be moving to log4j, so these dependencies should stay.

metis-enrichment/metis-enrichment-client/pom.xml
```diff
@@ -138,11 +129,7 @@
                 </exclusion>
             </exclusions>
         </dependency>
-		<dependency>
-		    <groupId>org.slf4j</groupId>
-		    <artifactId>slf4j-api</artifactId>
-		    <version>${version.slf4j}</version>
-		</dependency>
+
```
jochen-vermeulen: We shouldn't be moving to log4j, so this dependency should stay.

metis-enrichment/metis-enrichment-client/src/main/java/eu/europeana/enrichment/rest/client/EnrichmentWorker.java
```diff
@@ -27,8 +27,9 @@
  */
 public class EnrichmentWorker {
 
-  private static final Logger LOGGER = LoggerFactory.getLogger(EnrichmentWorker.class);
-
+  private static final Logger LOGGER = LogManager.getLogger(EnrichmentWorker.class);
```
jochen-vermeulen: We should not replace slf4j with log4j. The old version of the statement was the right one.

metis-enrichment/metis-enrichment-client/src/main/java/eu/europeana/enrichment/rest/client/EnrichmentClient.java
```diff
@@ -63,4 +60,8 @@ public EnrichmentBase getByUri(String uri) {
 
 		return response.getBody();
     }
+    
+    public void setRestTemplate(RestTemplate template) {
```
jochen-vermeulen: Make this method default (instead of public)? Only the test code needs to have access.

metis-enrichment/metis-enrichment-client/src/main/java/eu/europeana/enrichment/rest/client/DereferenceClient.java
```diff
@@ -157,4 +157,8 @@ public EnrichmentResultList dereference(String resourceId) {
     return restTemplate.exchange(dereferenceUrl, HttpMethod.GET, entity, EnrichmentResultList.class)
         .getBody();
   }
+  
+  public void setRestTemplate(RestTemplate restTemplate) {
```
jochen-vermeulen: Make this method default (instead of public)? Only the test code needs to have access.

metis-enrichment/metis-enrichment-client/src/test/java/eu/europeana/enrichment/rest/client/EnrichmentClientTest.java
```diff
@@ -0,0 +1,73 @@
...
+public class EnrichmentClientTest {
+	
+	@Test
+	public void testEnrich() {
+		EnrichmentResultList result = new EnrichmentResultList();
+		
+	    final RestTemplate restTemplate = mock(RestTemplate.class);
+	    doReturn(result).when(restTemplate).postForObject(anyString(), any(), any());
+	    
+	    final EnrichmentClient enrichmentClient = spy(new EnrichmentClient(""));
```
jochen-vermeulen: No need to spy this object? It is never verified.

metis-enrichment/metis-enrichment-client/src/test/java/eu/europeana/enrichment/rest/client/EnrichmentClientTest.java
```diff
@@ -0,0 +1,73 @@
...
+	    enrichmentClient.setRestTemplate(restTemplate);
+
+	    List<InputValue> values = new ArrayList<InputValue>();
+	    EnrichmentResultList res = enrichmentClient.enrich(values);
+	    
+	    verify(restTemplate, times(1)).postForObject(anyString(), any(), any());
+	    Assert.assertEquals(res, result);
```
jochen-vermeulen: Maybe also verify that restTemplate.postForObject is called with the right arguments?

And, if you really want to be thorough, that an exception is caught properly?

metis-enrichment/metis-enrichment-client/src/test/java/eu/europeana/enrichment/rest/client/EnrichmentClientTest.java
```diff
@@ -0,0 +1,73 @@
...
+		ResponseEntity<EnrichmentBase> result = new ResponseEntity<EnrichmentBase>(HttpStatus.OK);
+		
+	    final RestTemplate restTemplate = mock(RestTemplate.class);
+
+	    doReturn(result).when(restTemplate).exchange(any(URI.class),
+                any(HttpMethod.class),
+                any(HttpEntity.class),
+                eq(EnrichmentBase.class));
+	    
+	    final EnrichmentClient enrichmentClient = spy(new EnrichmentClient("http://dummy"));
```
jochen-vermeulen: No need to spy this object.

metis-enrichment/metis-enrichment-client/src/test/java/eu/europeana/enrichment/rest/client/EnrichmentClientTest.java
```diff
@@ -0,0 +1,73 @@
...
+	    enrichmentClient.setRestTemplate(restTemplate);
+
+	    EnrichmentBase res = enrichmentClient.getByUri("http://test");
+	    
+	    verify(restTemplate, times(1)).exchange(any(URI.class),
+                eq(HttpMethod.GET),
+                any(HttpEntity.class),
+                eq(EnrichmentBase.class));
+	    
+	    Assert.assertEquals(res, result.getBody());
```
jochen-vermeulen: Both values are null. So if these are equal it doesn't guarantee that the method works properly. I would suggest comparing non-null EnrichmentBase instances.

metis-enrichment/metis-enrichment-client/src/test/java/eu/europeana/enrichment/rest/client/DereferenceClientTest.java
```diff
@@ -0,0 +1,56 @@
...
+import eu.europeana.enrichment.api.external.model.EnrichmentBase;
+import eu.europeana.enrichment.api.external.model.EnrichmentResultList;
+import eu.europeana.enrichment.utils.InputValue;
+import org.junit.Assert;
+
+@RunWith(MockitoJUnitRunner.class)
+public class DereferenceClientTest {
+	
+	@Test
+	public void testDereference() {
```
jochen-vermeulen: See EnrichmentClientTest.testGetByUri() for remarks: this code is almost the same.

metis-enrichment/metis-enrichment-client/src/test/java/eu/europeana/enrichment/rest/client/EnrichmentWorkerTest.java
SidSatbhai: EnrichmentWorker writes logs specifically based on the set logging level. My main goal here was to increase code coverage by testing for all used log levels (debug and info in this case).

Log4J2 allows this during runtime using the Configurator.setLevel method and I have not been able to get runtime logging level changes to work using pure SLF4J.

metis-enrichment/metis-enrichment-client/src/main/java/eu/europeana/enrichment/rest/client/EnrichmentWorker.java
SidSatbhai: This is required for the reason mentioned previously: to change log levels during runtime.

metis-enrichment/metis-enrichment-client/src/main/java/eu/europeana/enrichment/rest/client/EnrichmentClient.java
SidSatbhai: Fixed.

metis-enrichment/metis-enrichment-client/src/main/java/eu/europeana/enrichment/rest/client/DereferenceClient.java
SidSatbhai: Fixed.

metis-enrichment/metis-enrichment-client/src/test/java/eu/europeana/enrichment/rest/client/EnrichmentClientTest.java
SidSatbhai: Fixed.

metis-enrichment/metis-enrichment-client/src/test/java/eu/europeana/enrichment/rest/client/EnrichmentClientTest.java
SidSatbhai: Fixed.

metis-enrichment/metis-enrichment-client/src/test/java/eu/europeana/enrichment/rest/client/EnrichmentClientTest.java
SidSatbhai: Fixed.

metis-enrichment/metis-enrichment-client/src/test/java/eu/europeana/enrichment/rest/client/EnrichmentClientTest.java
SidSatbhai: Fixed.

metis-enrichment/metis-enrichment-client/src/test/java/eu/europeana/enrichment/rest/client/DereferenceClientTest.java
SidSatbhai: Fixed.

metis-enrichment/metis-enrichment-client/pom.xml
SidSatbhai: Is there a specific reason why we've decided to stick with SLF4J while simultaneously using Log4J2 as the logging implementation? Log4J2 offers so much more functionality that we can't use if we stick with SLF4J.

Is this particularly because we are using dependencies throughout Metis that rely on the interface provided by SLF4J?

metis-enrichment/metis-enrichment-client/src/test/java/eu/europeana/enrichment/rest/client/EnrichmentClientTest.java
```diff
@@ -0,0 +1,107 @@
...
+	    
+	    final EnrichmentClient enrichmentClient = spy(new EnrichmentClient("http://dummy"));
+	    enrichmentClient.setRestTemplate(restTemplate);	    
+	    
+	    List<InputValue> values = new ArrayList<InputValue>();
+	    enrichmentClient.enrich(values);	
+	    
+	    verify(enrichmentClient).setRestTemplate(restTemplate);
+	    verify(enrichmentClient).enrich(values);
+	}
```
jochen-vermeulen: These two verify statements will not be executed because the method will exit after enrichmentClient.enrich throws the UnknownException. In this case you may wish to use the default try... fail... catch... verify approach.

metis-enrichment/metis-enrichment-client/pom.xml
jochen-vermeulen: Well, if we do decide to switch from one to the other, it should be a decision that we talked about, agreed on and subsequently implemented for all tools (not just enrichment). There are most likely also arguments to be made for sticking with SLF4J.

Regardless, though, I would say that this is outside the scope of this ticket, therefore I think this should be reverted to the way it was (using SLF4J).

metis-enrichment/metis-enrichment-client/src/test/java/eu/europeana/enrichment/rest/client/EnrichmentWorkerTest.java
jochen-vermeulen: This issue is still open.

metis-enrichment/metis-enrichment-client/src/test/java/eu/europeana/enrichment/rest/client/EnrichmentWorkerTest.java
jochen-vermeulen: Well the coverage gain must be practically nothing... negligible indeed, so to me not worth the trouble. And if you just want the coverage, why not just test it with the DEBUG level? Why also with INFO? On top of all this, I think it is not the right way to increase the coverage anyway as we are not actually testing the extra code that is covered. Lastly, I think it certainly is not a valid reason for such a drastic move as switching logging frameworks. 

For all these reasons I would still propose that this code be reverted to the original state.

SidSatbhai: Done.

metis-enrichment/metis-enrichment-client/src/test/java/eu/europeana/enrichment/rest/client/EnrichmentWorkerTest.java
SidSatbhai: Done.

metis-enrichment/metis-enrichment-client/pom.xml
SidSatbhai: Done.

metis-enrichment/metis-enrichment-client/pom.xml
SidSatbhai: Reverted.

metis-enrichment/metis-enrichment-client/src/test/java/eu/europeana/enrichment/rest/client/EnrichmentClientTest.java
SidSatbhai: Removed the unreachable lines.

metis-enrichment/metis-enrichment-client/src/test/java/eu/europeana/enrichment/rest/client/EnrichmentWorkerTest.java
```diff
@@ -1,15 +1,32 @@
 package eu.europeana.enrichment.rest.client;
 
+import static eu.europeana.metis.RestEndpoints.ENRICHMENT_ENRICH;
```
jochen-vermeulen: Lots of unused imports in this class.

metis-enrichment/metis-enrichment-client/src/test/java/eu/europeana/enrichment/rest/client/EnrichmentWorkerTest.java
```diff
@@ -23,23 +40,10 @@
...
-import eu.europeana.enrichment.utils.InputValue;
+import org.mockito.junit.MockitoJUnitRunner;
+import org.springframework.web.client.RestTemplate;
 
-// TODO Tests that could be added:
-// 1. What if there is nothing to process (empty or null lists are returned when extracting or when
-// enriching/dereferencing).
-// 2. Throwing exceptions
-@RunWith(MockitoJUnitRunner.class)
+@RunWith(MockitoJUnitRunner.Silent.class)
```
jochen-vermeulen: Usage of the silent mode is NOT recommended. Why not use the default MockitoJUnitRunner behaviour?

metis-enrichment/metis-enrichment-client/src/test/java/eu/europeana/enrichment/rest/client/EnrichmentWorkerTest.java
SidSatbhai: Fixed.

metis-enrichment/metis-enrichment-client/src/test/java/eu/europeana/enrichment/rest/client/EnrichmentWorkerTest.java
SidSatbhai: Fixed.

PR #159, opened 2018-05-02
Bug/met 936 preview service should use new indexing library

jochen-vermeulen:


metis-common/src/main/java/eu/europeana/metis/utils/ZipFileUtils.java
```diff
@@ -0,0 +1,72 @@
...
+import net.lingala.zip4j.core.ZipFile;
+import net.lingala.zip4j.exception.ZipException;
+
+/**
+ * This class provides utility methods for managing zip files.
+ * 
+ * @author jochen
+ *
+ */
+public class ZipFileUtils {
```
stzanakis: Make class final since private constructor.

metis-indexing/src/main/java/eu/europeana/indexing/ClientsConnectionProvider.java
```diff
@@ -0,0 +1,51 @@
...
+    return solrClient;
+  }
+
+  @Override
+  protected EdmMongoServer getMongoClient() {
+    return mongoClient;
+  }
+
+  @Override
+  public void close() throws IOException {}
```
stzanakis: Why is this empty?

metis-indexing/src/main/java/eu/europeana/indexing/ConnectionProvider.java
```diff
@@ -0,0 +1,46 @@
...
+ * Full Beans to be accessed by external agents.
+ * </p>
+ * <p>
+ * Please note that this class is {@link Closeable} and must be closed to release it's resources.
+ * </p>
+ * 
+ * @author jochen
+ *
+ */
+abstract class ConnectionProvider implements Closeable {
```
stzanakis: Let's start using this rule for naming abstract classes?
http://grepcode.com/file/repo1.maven.org/maven2/org.codehaus.sonar-plugins.java/java-checks/1.3/org/sonar/l10n/squidjava/rules/squid/S00118.html

metis-common/src/main/java/eu/europeana/metis/utils/ZipFileUtils.java
jochen-vermeulen: Fixed.

metis-indexing/src/main/java/eu/europeana/indexing/ConnectionProvider.java
jochen-vermeulen: Are we strict about it? I changed the name to AbstractConnectionProvider, but this name would not be valid according to the link you provided because two words follow after Abstract instead of one.

jochen-vermeulen:
The two issues you described are bugs in the Indexer, and have nothing to do with this pull request. I will make a new ticket for this in Jira.

stzanakis:
As discussed the mentioned issues will be handled on a separate ticket.

PR #162, opened 2018-05-07
Feat/met 974 integrate ecloud cancellation functionality

stzanakis:


metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/execution/WorkflowExecutor.java
```diff
@@ -81,41 +78,31 @@ public WorkflowExecution call() {
...
     //Run if the workflowExecution was retrieved from the queue in RUNNING state. Another process released it and came back into the queue
-    workflowExecution.setUpdatedDate(new Date());
+    if (workflowExecution.getWorkflowStatus() == WorkflowStatus.RUNNING) {
+      workflowExecution.setUpdatedDate(new Date());
+    } else {
+      workflowExecution.setStartedDate(new Date());
+    }
     workflowExecution.setWorkflowStatus(WorkflowStatus.RUNNING);
     workflowExecutionDao.updateMonitorInformation(workflowExecution);
     lock.unlock(); //Unlock as soon as possible
```
jochen-vermeulen: This seems like something that might go wrong in the future. It only works because nothing else happens after it before this is unlocked anyway in the calling method. May I suggest splitting this method in a locked part and an unlocked part? Then this statement can go and everything becomes a bit cleaner.

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/execution/WorkflowExecutor.java
```diff
@@ -62,14 +62,11 @@ public WorkflowExecution call() {
...
           .isExecutionActive(this.workflowExecution, monitorCheckIntervalInSecs)) {
-        runInQueueStateWorkflowExecution(lock);
-      } else if (workflowExecution.getWorkflowStatus() == WorkflowStatus.RUNNING
-          && !workflowExecutionDao
-          .isExecutionActive(this.workflowExecution, monitorCheckIntervalInSecs)) {
-        runRunningStateWorkflowExecution(lock);
+        runInqueueOrRunningStateWorkflowExecution(lock);
       } else {
         LOGGER.info(
             "Discarding user workflow execution with id: {}, it's not INQUEUE or RUNNNING and not active",
```
jochen-vermeulen: This log statement seems false... it could be already active?

PR #170, opened 2018-05-30
#EA-1125 disable Wikidata Address dereferenciation 

gsergiu:
disabled the retrieval of wikidata addresses, added unit test for dereferenciation of BNF, NISV and BL, added default constructor for
WikidataAccessDao

### DIFF
--- a/metis-enrichment/metis-enrichment-service/src/main/java/eu/europeana/enrichment/service/WikidataAccessService.java
+++ b/metis-enrichment/metis-enrichment-service/src/main/java/eu/europeana/enrichment/service/WikidataAccessService.java
@@ -37,7 +37,6 @@ import eu.europeana.enrichment.service.exception.WikidataAccessException;
 public class WikidataAccessService {
 
   public static final String WIKIDATA_BASE_URL = "http://www.wikidata.org/entity/Q";
-  public static final String WIKIDATA_ORGANIZATION_XSL_FILE = "/wkd2org.xsl";
   private static final Logger LOGGER = LoggerFactory.getLogger(WikidataAccessService.class);
 
   private final WikidataAccessDao wikidataAccessDao;
@@ -190,6 +189,8 @@ public class WikidataAccessService {
    */
   public void saveXmlToFile(String xml, File contentFile) throws WikidataAccessException {
     try {
+      if(!contentFile.exists())
+        contentFile.createNewFile();
       FileUtils.write(contentFile, xml, StandardCharsets.UTF_8.name());
     } catch (IOException e) {
       throw new WikidataAccessException(WikidataAccessException.XML_COULD_NOT_BE_WRITTEN_TO_FILE_ERROR, e);
--- a/metis-enrichment/metis-enrichment-service/src/main/java/eu/europeana/enrichment/service/WikidataAccessService.java
+++ b/metis-enrichment/metis-enrichment-service/src/main/java/eu/europeana/enrichment/service/WikidataAccessService.java
@@ -189,8 +189,8 @@ public class WikidataAccessService {
    */
   public void saveXmlToFile(String xml, File contentFile) throws WikidataAccessException {
     try {
-      if(!contentFile.exists())
-        contentFile.createNewFile();
+      //create content file if needed
+      contentFile.createNewFile();
       FileUtils.write(contentFile, xml, StandardCharsets.UTF_8.name());
     } catch (IOException e) {
       throw new WikidataAccessException(WikidataAccessException.XML_COULD_NOT_BE_WRITTEN_TO_FILE_ERROR, e);
--- a/metis-enrichment/metis-enrichment-service/src/main/java/eu/europeana/enrichment/service/dao/WikidataAccessDao.java
+++ b/metis-enrichment/metis-enrichment-service/src/main/java/eu/europeana/enrichment/service/dao/WikidataAccessDao.java
@@ -2,6 +2,8 @@ package eu.europeana.enrichment.service.dao;
 
 import java.io.ByteArrayInputStream;
 import java.io.File;
+import java.io.FileInputStream;
+import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.InputStream;
 import javax.xml.bind.JAXBContext;
@@ -35,6 +37,7 @@ import eu.europeana.enrichment.service.exception.WikidataAccessException;
  */
 public class WikidataAccessDao {
 
+  public static final String WIKIDATA_ORGANIZATION_XSL_FILE = "/wkd2org.xsl";
   private static final String SPARQL = "https://query.wikidata.org/sparql";
   private static final int SIZE = 1024 * 1024;
 
@@ -42,12 +45,22 @@ public class WikidataAccessDao {
 
   private Transformer transformer;
 
-  private File templateFile;
+  private InputStream xslTemplateImputStream;
 
   EntityConverterUtils entityConverterUtils = new EntityConverterUtils();
 
-  public WikidataAccessDao(File templateFile) throws WikidataAccessException {
-    this.templateFile = templateFile;
+  public WikidataAccessDao(File templateFile)
+      throws WikidataAccessException, FileNotFoundException {
+    this(new FileInputStream(templateFile));
+  }
+
+  public WikidataAccessDao(InputStream xslTemplate) throws WikidataAccessException {
+    this.xslTemplateImputStream = xslTemplate;
+    init();
+  }
+
+  public WikidataAccessDao() throws WikidataAccessException {
+    this.xslTemplateImputStream = getClass().getResourceAsStream(WIKIDATA_ORGANIZATION_XSL_FILE);
     init();
   }
 
@@ -64,7 +77,7 @@ public class WikidataAccessDao {
   public final void init() throws WikidataAccessException {
     TransformerFactory transformerFactory = TransformerFactory.newInstance();
     try {
-      Source xslt = new StreamSource(templateFile);
+      Source xslt = new StreamSource(xslTemplateImputStream);
       transformer = transformerFactory.newTransformer(xslt);
       transformer.setOutputProperty("{http://xml.apache.org/xslt}indent-amount", "4");
     } catch (TransformerConfigurationException e) {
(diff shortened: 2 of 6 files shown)


metis-enrichment/metis-enrichment-service/src/main/java/eu/europeana/enrichment/service/dao/WikidataAccessDao.java
```diff
@@ -64,7 +77,7 @@ public EntityConverterUtils getEntityConverterUtils() {
   public final void init() throws WikidataAccessException {
```
jochen-vermeulen: Adjust java doc: there is no more file parameter. The same goes for the parameter outputFile in the function getEntity below.

metis-enrichment/metis-enrichment-service/src/main/java/eu/europeana/enrichment/service/dao/WikidataAccessDao.java
```diff
@@ -35,19 +37,30 @@
...
+  public static final String WIKIDATA_ORGANIZATION_XSL_FILE = "/wkd2org.xsl";
   private static final String SPARQL = "https://query.wikidata.org/sparql";
   private static final int SIZE = 1024 * 1024;
 
   private static final Logger LOGGER = LoggerFactory.getLogger(WikidataAccessDao.class);
 
   private Transformer transformer;
 
-  private File templateFile;
+  private InputStream xslTemplateImputStream;
```
jochen-vermeulen: I think it is a bad idea to have an input stream as a class variable of a class that is supposed to be a long living object. Also the input stream is never closed and removed. Why not perform the init method immediately upon construction so that you don't need to keep the input stream? Or read the input stream upon construction into a string or byte array?

metis-enrichment/metis-enrichment-service/src/main/java/eu/europeana/enrichment/service/WikidataAccessService.java
```diff
@@ -190,6 +189,8 @@ public Organization toOrganizationImpl(WikidataOrganization wikidataOrganization
    */
   public void saveXmlToFile(String xml, File contentFile) throws WikidataAccessException {
     try {
+      if(!contentFile.exists())
```
jochen-vermeulen: Don't need this if statement. File.createNewFile will only create a file if it doesn't already exists.

metis-enrichment/metis-enrichment-service/src/test/java/eu/europeana/enrichment/service/wikidata/WikidataAccessDaoTest.java
```diff
@@ -41,41 +47,129 @@ public void tearDown() throws Exception {}
 
   // TODO JV it is not a good idea to test on real-time data that comes from a external source. This
   // can break our tests if the data is changed.
+  // Answer SG: we can try to split the tests in unit tests and integration tests. However the data structure on wikidata is stable, 
+  // adding new fields or values must not break the code. Otherwise it is clear that we have a bug in the implementation. 
+  //It is good to see such bugs as soon as possible. We may consider to have the same tests implemented both as unit (using local files), and integration (using http requests)tests  
+  /**
```
jochen-vermeulen: Like we discussed, we had a data change recently that broke our unit tests. Can we make a task for making sure that these tests are marked as integration tests and skipped by our build? 

At any rate, let's not have an entire chat session in our code. This should be removed as soon as we have a solution.

metis-enrichment/metis-enrichment-service/src/test/java/eu/europeana/enrichment/service/wikidata/WikidataAccessDaoTest.java
```diff
@@ -41,41 +47,129 @@ public void tearDown() throws Exception {}
...
-    String savedWikidataXml = FileUtils.readFileToString(wikidataTestOutputFile, "UTF-8"); 
-    assertEquals(wikidataXml, savedWikidataXml);
-
-    //convert Wikidata organization to OrganizationImpl object
-    Organization organizationImpl = wikidataAccessService.toOrganizationImpl(wikidataOrganization);
-    assertNotNull(organizationImpl);
-    assertEquals(38, organizationImpl.getPrefLabel().values().size());
-    assertEquals(organizationImpl.getAbout(), TEST_WIKIDATA_URL);
-    assertEquals(TEST_ACRONYM, organizationImpl.getEdmAcronym().get(Locale.FRENCH.getLanguage()).get(0));
+    //File wikidataTestOutputFile = getClasspathFile(WIKIDATA_TEST_OUTPUT_FILE);
```
jochen-vermeulen: Why is this line commented out? Can we remove it?

metis-enrichment/metis-enrichment-service/src/main/java/eu/europeana/enrichment/service/dao/WikidataAccessDao.java
gsergiu: Actually the init() method is called by the constructors. And the stream is closed automatically by the transformer factory class:  transformerFactory.newTransformer(xslt) within the scope of init method. This happens somewhere in the low level implementation of XML (I already verified this in debugger).

metis-enrichment/metis-enrichment-service/src/main/java/eu/europeana/enrichment/service/WikidataAccessService.java
gsergiu: Correct, but the next line of code FileUtils.write() requires that the file exists. I think it is better to create the file in this method, eather than in all other methods that call the serialization to xml file.

metis-enrichment/metis-enrichment-service/src/test/java/eu/europeana/enrichment/service/wikidata/WikidataAccessDaoTest.java
gsergiu: actually I updated the assertions in this pull request and change == to >= comparisons. So adding new labels in wikidata will not break the tests anymore. It is very unlikely that the labels will be removed from wikidata, even if this is theoretically possible. ... For now I left the comments with the test code as I aggree that this is not the perfect solution. The separation of unit and integration tests should be addressed in another ticket.

gsergiu: PS: It would be fine to create a new issue by copy-pasting the chat from the code comments.

metis-enrichment/metis-enrichment-service/src/test/java/eu/europeana/enrichment/service/wikidata/WikidataAccessDaoTest.java
gsergiu: yes ..  I'll remove it

metis-enrichment/metis-enrichment-service/src/main/java/eu/europeana/enrichment/service/dao/WikidataAccessDao.java
gsergiu: done

metis-enrichment/metis-enrichment-service/src/test/java/eu/europeana/enrichment/service/wikidata/WikidataAccessDaoTest.java
gsergiu: done

metis-enrichment/metis-enrichment-service/src/main/java/eu/europeana/enrichment/service/WikidataAccessService.java
jochen-vermeulen: No, I mean keep the line below (contentFile.createNewFile();) but just remove the line with the if statement. 

contentFile.createNewFile will not create a new file if one already exists. So it does it's own check - no need to add ours.There is no point in checking the disk twice (and generating the extra IO that comes with that).

metis-enrichment/metis-enrichment-service/src/main/java/eu/europeana/enrichment/service/dao/WikidataAccessDao.java
jochen-vermeulen: Firstly, I couldn't find anywhere in the specifications of transformerFactory.newTransformer() that it will close the input stream (or source). So maybe they do, maybe they only do sometimes or under certain conditions, and maybe they will stop doing it in a new version. Please let's just be safe and make sure the input stream is closed.

Secondly: I don't know what is inside the input stream that is passed. There may be a byte array that is no longer needed and that will only be freed once the stream is garbage collected. 

I still don't see why the input stream cannot just be passed directly to the init() method. But if that really cannot be done, may I suggest the following alternative? 

Instead of passing an input stream during construction, we pass an input stream supplier: java.util.function.Supplier<InputStream>. Then in the init() method we create the input stream (using the supplier), read it and close it, and nothing is kept in memory. The callers are then completely responsible for using it properly (i.e. they should not simply pass a wrapper around an input stream but actually specify how the stream is to be created).

Any thoughts?

metis-enrichment/metis-enrichment-service/src/main/java/eu/europeana/enrichment/service/WikidataAccessService.java
gsergiu: I see ..  I was reading too fast the comment. I remove the line now.

metis-enrichment/metis-enrichment-service/src/main/java/eu/europeana/enrichment/service/dao/WikidataAccessDao.java
gsergiu: Well I don't expect the JDK rt.jar will be changed so that they don't close streams anymore in the future versions. 
Now the default constructur is used by the code, 
this.xslTemplateImputStream = getClass().getResourceAsStream(WIKIDATA_ORGANIZATION_XSL_FILE);
So .. the input stream is the connected to the /wkd2org.xsl which is available in resources.

I can pass the inputStream to the init() method if you consider this to be better.

PR #172, opened 2018-06-01
Bug/met 1102 make xsl transformation thread safe

jochen-vermeulen:


metis-transformation/metis-transformation-service/src/main/java/eu/europeana/metis/transformation/service/XsltTransformer.java
```diff
@@ -1,84 +1,137 @@
...
-  private StringWriter executeTransformation(Transformer transformer, InputStream contentStream)
-      throws TransformerException {
-    Source xmlDoc = new StreamSource(contentStream);
-    StringWriter writer = new StringWriter();
-    transformer.transform(xmlDoc, new StreamResult(writer));
-    return writer;
-  }
+    // If it doesn't, resolve the URL and create the compiled transformation.
+    final TransformerFactory transformerFactory = new TransformerFactoryImpl();
+    final Templates templates;
```
stzanakis: If the final variable is not set immediately then this usually is a code smell on SonarQube.

metis-transformation/metis-transformation-service/src/main/java/eu/europeana/metis/transformation/service/XsltTransformer.java
```diff
@@ -1,84 +1,137 @@
...
+ * For each thread a new instance needs to be created, but, due to the caching mechanism of the XSLT
+ * compilation, this operation is not very expensive.
  */
 public class XsltTransformer {
 
   private static final Logger LOGGER = LoggerFactory.getLogger(XsltTransformer.class);
 
   private static final int CACHE_SIZE = 50;
-  private static LRUCache<String, Transformer> cache = new LRUCache<>(CACHE_SIZE);
+  private static LRUCache<String, Templates> cache = new LRUCache<>(CACHE_SIZE);
```
stzanakis: Should be final.

metis-transformation/metis-transformation-service/src/main/java/eu/europeana/metis/transformation/service/XsltTransformer.java
```diff
@@ -1,84 +1,137 @@
...
+
+  /**
+   * Constructor.
+   * 
+   * @param xsltUrl The URL of the XSLT file.
+   * @param datasetId The value that will be injected to the datasetId field in the XSLT. Can be
+   *        null.
+   * @throws TransformationException In case there was a problem with setting up the transformation.
+   */
+  public XsltTransformer(String xsltUrl, String datasetId) throws TransformationException {
```
stzanakis: Previous implementation changed which looks nice but we were expected more fields to be injected on the xslt transformation. On the other hand another argument I told you recently is that we might want to extract the code of these kind of fields to a java based transformation instead.(e.g. the generation of europeana identifiers which is tackled during this sprint.)
Since it's not clear yet of how we'll proceed with what I just explained, I think we can keep it as you implemented it.

PR #173, opened 2018-06-04
MET-1117 Validate and trim harvesting parameters on workflow create/u…

jochen-vermeulen:
…dpate

### DIFF
--- a/metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/service/OrchestratorService.java
+++ b/metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/service/OrchestratorService.java
@@ -1,22 +1,5 @@
 package eu.europeana.metis.core.service;
 
-import java.io.IOException;
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.Date;
-import java.util.EnumSet;
-import java.util.List;
-import java.util.Set;
-import java.util.stream.Collectors;
-import java.util.stream.Stream;
-import org.apache.commons.lang3.StringUtils;
-import org.bson.types.ObjectId;
-import org.redisson.api.RLock;
-import org.redisson.api.RedissonClient;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-import org.springframework.beans.factory.annotation.Autowired;
-import org.springframework.stereotype.Service;
 import eu.europeana.metis.CommonStringValues;
 import eu.europeana.metis.RestEndpoints;
 import eu.europeana.metis.authentication.user.AccountRole;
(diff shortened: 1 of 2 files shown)


metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/service/OrchestratorService.java
```diff
@@ -726,6 +743,35 @@ public DatasetExecutionInformation getDatasetExecutionInformation(MetisUser meti
...
+        .getPluginMetadata(PluginType.OAIPMH_HARVEST);
+    if (oaipmhPluginMetadata != null) {
+      URL u = new URL(oaipmhPluginMetadata.getUrl().trim()); // this would check for the protocol
+      u.toURI(); // does the extra checking required for validation of URI
+      //Remove all the query parameters
+      int queryIndexStart = oaipmhPluginMetadata.getUrl().indexOf('?');
+      String urlWithoutQueryParameters = queryIndexStart >= 0 ? oaipmhPluginMetadata.getUrl()
+          .substring(0, oaipmhPluginMetadata.getUrl().indexOf('?'))
+          : oaipmhPluginMetadata.getUrl();
+      oaipmhPluginMetadata.setUrl(urlWithoutQueryParameters.trim());
```
jochen-vermeulen: I feel it's typically better to use libraries to mess with URIs. In this case, the Java API itself can remove the query (and supposedly the fragment too):
`String strippedUri = new URI(uri.getScheme(), uri.getAuthority(), uri.getPath(), null, null).toString();`

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/service/OrchestratorService.java
```diff
@@ -726,6 +743,35 @@ public DatasetExecutionInformation getDatasetExecutionInformation(MetisUser meti
...
+          : oaipmhPluginMetadata.getMetadataFormat().trim());
+      oaipmhPluginMetadata.setSetSpec(oaipmhPluginMetadata.getSetSpec() == null ? null
+          : oaipmhPluginMetadata.getSetSpec().trim());
+    }
+
+    HTTPHarvestPluginMetadata httpHarvestPluginMetadata = (HTTPHarvestPluginMetadata) workflow
+        .getPluginMetadata(PluginType.HTTP_HARVEST);
+    if (httpHarvestPluginMetadata != null) {
+      URL u = new URL(
+          httpHarvestPluginMetadata.getUrl().trim()); // this would check for the protocol
```
jochen-vermeulen: Can't you do URI.parse() directly? Is the detour through URL really necessary?

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/service/OrchestratorService.java
stzanakis: okay that looks better indeed.

PR #177, opened 2018-06-06
MET-1083 Require old password check when requesting an update of the …

stzanakis:
…password.

### DIFF
--- a/metis-authentication/metis-authentication-rest/src/main/java/eu/europeana/metis/authentication/rest/AuthenticationController.java
+++ b/metis-authentication/metis-authentication-rest/src/main/java/eu/europeana/metis/authentication/rest/AuthenticationController.java
@@ -13,13 +13,13 @@ import eu.europeana.metis.exception.UserUnauthorizedException;
 import java.util.List;
 import javax.ws.rs.QueryParam;
 import org.apache.commons.lang3.StringEscapeUtils;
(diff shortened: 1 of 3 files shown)


metis-authentication/metis-authentication-rest/src/main/java/eu/europeana/metis/authentication/rest/AuthenticationController.java
```diff
@@ -115,14 +118,18 @@ public MetisUser loginUser(@RequestHeader("Authorization") String authorization)
...
   @ResponseStatus(HttpStatus.NO_CONTENT)
   public void updateUserPassword(@RequestHeader("Authorization") String authorization,
+      @QueryParam("oldPassword") String oldPassword,
       @QueryParam("newPassword") String newPassword)
       throws GenericMetisException {
-    String accessToken = authenticationService
-        .validateAuthorizationHeaderWithAccessToken(authorization);//Before any action, validate token
-    if (StringUtils.isEmpty(newPassword)) {
+    if (StringUtils.isBlank(oldPassword) || StringUtils.isBlank(newPassword)) {
       throw new BadContentException("newPassword not provided");
```
jochen-vermeulen: This is not an appropriate error warning if the old password is blank. Could be very confusing while debugging!

PR #183, opened 2018-06-19
MET-1186 Add extra parameters on XSLT plugin.

stzanakis:


metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/service/DatasetService.java
```diff
@@ -436,29 +439,30 @@ public DatasetXslt getLatestXsltForDatasetId(String datasetId) throws GenericMet
...
-    }
+  private List<Record> transformRecords(Dataset dataset, List<Record> records, String xsltUrl) {
 
     // Transform the records.
     return records.stream().map(record -> {
       try {
-        return new Record(record.getEcloudId(), transformer
-            .transform(record.getXmlRecord().getBytes(StandardCharsets.UTF_8)).toString());
+        String recordString = new String(record.getXmlRecord().getBytes(StandardCharsets.UTF_8));
+        EuropeanaIdCreator europeanIdCreator = new EuropeanaIdCreator();
```
jochen-vermeulen: Create this object outside the 'loop'.

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/service/DatasetService.java
```diff
@@ -436,29 +439,30 @@ public DatasetXslt getLatestXsltForDatasetId(String datasetId) throws GenericMet
...
+        EuropeanaGeneratedIdsMap europeanaGeneratedIdsMap = europeanIdCreator
+            .constructEuropeanaId(recordString, dataset.getDatasetId());
+        final XsltTransformer transformer = new XsltTransformer(xsltUrl, europeanaGeneratedIdsMap, dataset.getDatasetName(),
```
jochen-vermeulen: Create this object outside the 'loop' too (see my remarks in XsltTransformer).

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/service/DatasetService.java
```diff
@@ -436,29 +439,30 @@ public DatasetXslt getLatestXsltForDatasetId(String datasetId) throws GenericMet
...
-      throw new XsltSetupException("Could not setup XSL transformation.", e);
-    }
+  private List<Record> transformRecords(Dataset dataset, List<Record> records, String xsltUrl) {
 
     // Transform the records.
     return records.stream().map(record -> {
       try {
-        return new Record(record.getEcloudId(), transformer
-            .transform(record.getXmlRecord().getBytes(StandardCharsets.UTF_8)).toString());
+        String recordString = new String(record.getXmlRecord().getBytes(StandardCharsets.UTF_8));
```
jochen-vermeulen: This I don't understand: record.getXmlRecord() already gives a String.

metis-transformation/metis-transformation-service/src/main/java/eu/europeana/metis/transformation/service/XsltTransformer.java
```diff
@@ -32,37 +32,60 @@
...
-   * @param datasetId The value that will be injected to the datasetId field in the XSLT. Can be
-   *        null.
+   * @param europeanaGeneratedIdsMap all the identifiers related to europeana RDF elements
+   * @param datasetName the dataset name related to the dataset
+   * @param edmCountry the Country related to the dataset
+   * @param edmLanguage the language related to the dataset
    * @throws TransformationException In case there was a problem with setting up the transformation.
    */
-  public XsltTransformer(String xsltUrl, String datasetId) throws TransformationException {
+  public XsltTransformer(String xsltUrl, EuropeanaGeneratedIdsMap europeanaGeneratedIdsMap,
```
jochen-vermeulen: This class was designed to be created once for the transformation of multiple records, so that the (expensive) compilation of the XSLT could be done only once. The inclusion of the EuropeanaGeneratedIdsMap in the constructor here however forces a new instance to be created for every record. Maybe it is better to pass that object to the transform method instead (and set the parameter values there instead of in the constructor). 

Alternatively the object could be changed so that it can on-the-fly compute the parameters given the record ID, but still the values should then be set in the transform method, not in the constructor.

PR #186, opened 2018-06-25
Debt/met 1193 move creation of full bean to metis

jochen-vermeulen:
I'm aware that there are some SonarQube issues (mainly methods that are too complex) and that the coverage decreased a bit. We could schedule more maintenance tickets next sprint to deal with this.

### DIFF
--- a/metis-indexing/src/main/java/eu/europeana/indexing/mongo/property/LicenseUpdater.java
+++ b/metis-indexing/src/main/java/eu/europeana/indexing/mongo/property/LicenseUpdater.java
@@ -1,0 +1,48 @@
+package eu.europeana.indexing.mongo.property;
+
+import eu.europeana.corelib.edm.exceptions.MongoUpdateException;
+import org.apache.commons.lang.StringUtils;
+import org.mongodb.morphia.query.Query;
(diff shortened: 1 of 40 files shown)


metis-indexing/src/main/java/eu/europeana/indexing/mongo/property/LicenseUpdater.java
```diff
@@ -0,0 +1,50 @@
...
+
+  @Override
+  public LicenseImpl update(LicenseImpl mongoEntity, LicenseImpl newEntity,
+      MongoServer mongoServer) {
+    Query<LicenseImpl> updateQuery = mongoServer.getDatastore().createQuery(LicenseImpl.class)
+        .field("about").equal(mongoEntity.getAbout());
+    UpdateOperations<LicenseImpl> ops =
+        mongoServer.getDatastore().createUpdateOperations(LicenseImpl.class);
+    final UpdateTrigger updateTrigger = new UpdateTrigger();
+    if (mongoEntity.getCcDeprecatedOn() != newEntity.getCcDeprecatedOn()) {
```
stzanakis: Can we check this one? Is the Date value to be checked actually, instead of the Object?

PR #192, opened 2018-07-06
Debt/met 1160 data checker will use transformation library

jochen-vermeulen:


metis-transformation/metis-transformation-service/src/main/java/eu/europeana/metis/transformation/service/CacheWithExpirationTime.java
```diff
@@ -0,0 +1,219 @@
...
+   * 
+   * @param since The interval length of the period we want to check (which ends now). A negative
+   *        duration cleans everything.
+   */
+  public void removeItemsNotAccessedSince(Duration since) {
+    lock.writeLock().lock();
+    try {
+      final Iterator<Entry<K, CacheItemWithExpirationTime<V>>> iterator =
+          cache.entrySet().iterator();
+      while (iterator.hasNext()) {
```
stzanakis: Have you heard of removeIf method?
this block could be simply for example:
cache.entrySet().removeIf(entry -> !entry.getValue().valueWasAccessedRecently(since));

PR #198, opened 2018-07-26
Feat/met 1266 mongo retryable calls

stzanakis:


metis-common/src/main/java/eu/europeana/metis/utils/MongoRequestUtil.java
```diff
@@ -0,0 +1,89 @@
...
+   * It's a {@link java.util.function.Supplier} functional interface that throws a {@link
+   * MongoSocketException}.
+   * <p>This supplier is to be used for requests to Mongo, and was created to avoid the
+   * issues that are faced with the Application and Mongo in Cloud Foundry where the underlying
+   * system would throw a {@link java.net.SocketException}</p>
+   *
+   * @param <R> the object that should be returned.
+   */
+  @FunctionalInterface
+  public interface SupplierWithMongoSocketException<R> {
```
jochen-vermeulen: Since get() throws only a RuntimeException, this whole interface is not necessary, and a regular Supplier could be used.

metis-common/src/main/java/eu/europeana/metis/utils/MongoRequestUtil.java
```diff
@@ -0,0 +1,89 @@
...
+   */
+  public static <R> R retryableMongoRequest(
+      SupplierWithMongoSocketException<R> supplierWithMongoSocketException) {
+    int retryCounter = 0;
+
+    R result = null;
+    boolean success = false;
+    do {
+      try {
+        result = supplierWithMongoSocketException.get();
```
jochen-vermeulen: Just do:
return supplierWithMongoSocketException.get();
No need to keep a boolean as well as the result in a variable.
The loop should be do ... while (true).

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/dao/ScheduledWorkflowDao.java
```diff
@@ -93,9 +96,9 @@ public ScheduledWorkflow getScheduledWorkflow(String datasetId) {
...
   public ScheduledWorkflow getScheduledWorkflowByDatasetId(String datasetId) {
-    return morphiaDatastoreProvider.getDatastore()
-        .find(ScheduledWorkflow.class).field(DATASET_ID)
-        .equal(datasetId).get();
+    return MongoRequestUtil.retryableMongoRequest(
+        () -> morphiaDatastoreProvider.getDatastore().find(ScheduledWorkflow.class)
+            .field(DATASET_ID).equal(datasetId).get());
   }
 
   /**
```
jochen-vermeulen: Should the method 'exists' below not also be changed?

metis-common/src/main/java/eu/europeana/metis/utils/MongoRequestUtil.java
stzanakis: Yes! Indeed!

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/dao/ScheduledWorkflowDao.java
stzanakis: I missed that, yes it should.

PR #211, opened 2018-10-31
Feat/met 1468 minute cap for cancelation of hanging executions

stzanakis:


metis-core/metis-core-common/src/main/java/eu/europeana/metis/core/workflow/plugins/PluginStatus.java
```diff
@@ -7,7 +7,7 @@
  * @since 2017-05-26
  */
 public enum PluginStatus {
-  INQUEUE, RUNNING, FINISHED, CANCELLED, FAILED;
+  INQUEUE, CLEANING, RUNNING, FINISHED, CANCELLED, FAILED;
```
jochen-vermeulen: There may be some checks (if statements) that need to be changed given the new CLEANING state. I found the following candidates (maybe changing is not strictly necessary in all of them, but it is better to be safe than sorry):
1. WorkflowExecution.setAllRunningAnInqueuePluginsToCancelled
2. WorkflowExecution.checkAndSetAllRunningAndInqueuePluginsToCancelledIfOnePluginHasFailed
3. WorkflowExecutor.runInqueueOrRunningStateWorkflowExecution

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/execution/WorkflowExecutor.java
```diff
@@ -237,7 +243,30 @@ private void periodicCheckingLoop(long sleepTime, AbstractMetisPlugin abstractMe
...
+    //Or if we have progress
+    if (abstractMetisPlugin.getPluginStatus() == PluginStatus.CLEANING
+        || previousProcessedRecords.get() != processedRecords) {
+      stableRecordCountPeriodCounterInSeconds.set(0L);
+      previousProcessedRecords.set(processedRecords);
+      return false;
+    }
+
+    final boolean isMinuteCapOverWithoutChangeInProcessedRecords =
+        stableRecordCountPeriodCounterInSeconds.addAndGet(monitorCheckIntervalInSecs)
```
jochen-vermeulen: This doesn't work properly. You add the monitor check interval in the assumption that this is the amount of time that passed since the last time you came here. But the amount of time that passed is probably significantly bigger. 

I think you should maintain the reset time instead (i.e. in line 255 set System.currentTimeMillis(), so that here you only need to check how much time passed since then).

PR #221, opened 2018-11-28
Bug/met 1512 fix enrichment proxy deployment error

stzanakis:


metis-enrichment/metis-enrichment-service/src/main/java/eu/europeana/enrichment/service/RedisInternalEnricher.java
```diff
@@ -228,8 +221,10 @@ private void loadEntity(EntityType entityType, MongoTerm term, Jedis jedis) {
         jedis.hset(entityType.cachedEntityPrefix + CACHED_URI, term.getCodeUri(),
             OBJECT_MAPPER.writeValueAsString(entityWrapper));
         List<String> parents = this.findParents(termList.getParent(), entityType.entityClass);
-        jedis.sadd(entityType.cachedEntityPrefix + CACHED_PARENT + term.getCodeUri(),
-            parents.toArray(new String[] {}));
+        if (parents != null && !parents.isEmpty()) {
```
jochen-vermeulen: According to my sonarqube, parents can not be null here (it comes from a private method and it is not null).

metis-enrichment/metis-enrichment-service/src/main/java/eu/europeana/enrichment/service/RedisInternalEnricher.java
```diff
@@ -201,13 +193,14 @@ private void loadEntities(EntityType entityType) {
     for (MongoTerm term : terms) {
       loadEntity(entityType, term, jedis);
       i++;
-      if (i % 100 == 0) {
+      final int countOfItemsCollectedToLog = 100;
```
jochen-vermeulen: Shouldn't this be a constant outside this method?

metis-enrichment/metis-enrichment-service/src/main/java/eu/europeana/enrichment/service/RedisInternalEnricher.java
stzanakis: Correct! Removed null check.

metis-enrichment/metis-enrichment-service/src/main/java/eu/europeana/enrichment/service/RedisInternalEnricher.java
stzanakis: Because it's just about logging I find this not, so I find both options of the same value. What do you think?

jochen-vermeulen: I don't know... I'm just used to putting it outside. Also for when I have to create another piece of code that does logging in that class I will easily see that this setting is already there. But maybe I'm thinking this way just because I'm used to it.

Anyway, just do what you think is right. And feel free to merge the code when you're done.

PR #225, opened 2019-02-06
Feat/met 1712 support more information when a cancellation occurs

stzanakis:


metis-authentication/metis-authentication-common/src/main/java/eu/europeana/metis/authentication/user/MetisUser.java
```diff
@@ -71,12 +74,22 @@ public MetisUser() {
...
+   * that should be followed
    */
   public MetisUser(JsonNode jsonNode) throws ParseException, BadContentException {
     parseJsonNodeZohoUserToMetisUser(jsonNode);
   }
 
+  /**
+   * Nullifies any sensitive information of this user.
+   */
+  public void removePrivateDataForPublicUse() {
```
jochen-vermeulen: I typically don't like if there is a method call required to prepare a database object for being shared. Especially if this is sensitive information: the next person to work on this may forget to add this call.

My solution is usually to make a new object with only the fields you need. But in this case, the MetisUser class already has a JsonIgnore annotation on password... wouldn't that be enough?

metis-authentication/metis-authentication-rest/src/main/java/eu/europeana/metis/authentication/rest/AuthenticationController.java
```diff
@@ -253,6 +254,34 @@ public void updateUserToMakeAdmin(@RequestHeader("Authorization") String authori
...
+   * <ul>
+   * <li>{@link UserUnauthorizedException} if the authorization header is un-parsable or the user
+   * cannot be authenticated.</li>
+   * </ul>
+   */
+  @RequestMapping(value = RestEndpoints.AUTHENTICATION_USER_BY_USER_ID, method = RequestMethod.POST, produces = {
+      MediaType.APPLICATION_JSON_VALUE, MediaType.APPLICATION_XML_VALUE})
+  @ResponseBody
+  public MetisUser getUserByUserId(@RequestHeader("Authorization") String authorization,
+      @RequestBody UserIdParameter userIdParameter) throws GenericMetisException {
```
jochen-vermeulen: Can the user ID not be a url parameter (REST-like)? Is this considered too sensitive to put in a query?

metis-core/metis-core-common/src/main/java/eu/europeana/metis/core/workflow/CancelledSystemId.java
```diff
@@ -0,0 +1,11 @@
+package eu.europeana.metis.core.workflow;
+
+/**
+ * Enum that contains values used as identifiers for system specific cancellation operations.
+ *
+ * @author Simon Tzanakis (Simon.Tzanakis@europeana.eu)
+ * @since 2019-02-06
+ */
+public enum CancelledSystemId {
+  SYSTEM_MINUTE_CAP_EXPIRE
```
jochen-vermeulen: Why put this in an enum? I don't think this adds any value. Why not create a string constant somewhere?

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/dao/WorkflowExecutionDao.java
```diff
@@ -132,13 +136,28 @@ public void updateMonitorInformation(WorkflowExecution workflowExecution) {
...
+   */
+  public void setCancellingState(WorkflowExecution workflowExecution, MetisUser metisUser) {
     UpdateOperations<WorkflowExecution> workflowExecutionUpdateOperations = morphiaDatastoreProvider
         .getDatastore().createUpdateOperations(WorkflowExecution.class);
     Query<WorkflowExecution> query = morphiaDatastoreProvider.getDatastore()
         .find(WorkflowExecution.class)
         .filter("_id", workflowExecution.getId());
     workflowExecutionUpdateOperations.set("cancelling", Boolean.TRUE);
+    if (metisUser != null && metisUser.getUserId() != null) {
+      workflowExecutionUpdateOperations.set("cancelledBy", metisUser.getUserId());
```
jochen-vermeulen: This is a tiny thing: maybe use the if statement only to determine the value of the user ID? Then you don't have the 'duplicate' part of setting the 'cancelledBy' parameter in the operations object (and you don't have to use the String constant twice).

metis-core/metis-core-common/src/main/java/eu/europeana/metis/core/workflow/CancelledSystemId.java
stzanakis: I originally had it as a constant field in the class but then I needed in the unit tests so I found enum to be more clear for the time being.

metis-authentication/metis-authentication-rest/src/main/java/eu/europeana/metis/authentication/rest/AuthenticationController.java
stzanakis: Well I followed this practice since we did the same for another call that had as a query parameter the user email.

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/dao/WorkflowExecutionDao.java
stzanakis: Fixed.

metis-authentication/metis-authentication-common/src/main/java/eu/europeana/metis/authentication/user/MetisUser.java
stzanakis: Removed it for the time being.

PR #226, opened 2019-02-06
MET-1675: Allow for some video/audio fields to be in the format object.

jochen-vermeulen:


metis-media-service/src/main/java/eu/europeana/metis/mediaprocessing/extraction/AudioVideoProcessor.java
```diff
@@ -160,6 +165,25 @@ public ResourceExtractionResult process(Resource resource) throws MediaExtractio
     return new ResourceExtractionResult(metadata, null);
   }
 
+  private static int findInt(String key, JSONObject[] candidates) {
+    final int defaultValue = Integer.MIN_VALUE;
+    return Stream.of(candidates).map(candidate -> candidate.optInt(key, defaultValue))
```
stzanakis: Just use the Integer.MIN_VALUE in both usages of defaultValue instead.

metis-media-service/src/main/java/eu/europeana/metis/mediaprocessing/extraction/AudioVideoProcessor.java
```diff
@@ -160,6 +165,25 @@ public ResourceExtractionResult process(Resource resource) throws MediaExtractio
...
+  }
+
+  private static double findDouble(String key, JSONObject[] candidates) {
+    return Stream.of(candidates).map(candidate -> candidate.optDouble(key, Double.NaN))
+        .filter(value -> !Double.isNaN(value)).findFirst()
+        .orElseThrow(() -> new JSONException("Could not find double for field: " + key));
+  }
+
+  private static String findString(String key, JSONObject[] candidates) {
+    return Stream.of(candidates).map(candidate -> candidate.optString(key, ""))
```
stzanakis: Unrelated but I found that StringUtils.EMPTY. Should we start using it with empty strings, what do you think?

jochen-vermeulen: There is a whole conversation: 
[https://stackoverflow.com/questions/4095501/is-stringutils-empty-recommended]

I think essentially, it doesn't matter much. Maybe use it only if you really chose the value because it is empty? Whatever makes the code more understandable in that specific circumstance.

I decided to change it here just because we then use StringUtils again to check for empty.

metis-media-service/src/main/java/eu/europeana/metis/mediaprocessing/extraction/AudioVideoProcessor.java
jochen-vermeulen: Fixed.

PR #228, opened 2019-02-11
MET-1678: Save error message when marking execution as failed.

jochen-vermeulen:


metis-core/metis-core-common/src/main/java/eu/europeana/metis/core/workflow/plugins/AbstractMetisPlugin.java
```diff
@@ -176,10 +178,20 @@ public PluginStatus getPluginStatus() {
...
 
   /**
+   * This method also clears the fail message if the status is set to anything other than {@link
+   * PluginStatus#FAILED}.
+   *
    * @param pluginStatus {@link PluginStatus}
    */
   public void setPluginStatus(PluginStatus pluginStatus) {
     this.pluginStatus = pluginStatus;
+    if (this.pluginStatus != PluginStatus.FAILED) {
```
stzanakis: I think the failMessage should always be reset not only for non-FAILED statuses.
Also probably better to have it in a different call e.g. setPluginStatusAndResetFailMessage so that it's clear what the end result is, and not using the default setter.

metis-core/metis-core-common/src/main/java/eu/europeana/metis/core/workflow/plugins/AbstractMetisPlugin.java
```diff
@@ -57,6 +58,7 @@
   private String id;
 
   private PluginStatus pluginStatus = PluginStatus.INQUEUE;
+  private String failMessage;
```
stzanakis: The getter of this field is missing, which is required for jaxb.

jochen-vermeulen: Good point. I fixed it.

metis-core/metis-core-common/src/main/java/eu/europeana/metis/core/workflow/plugins/AbstractMetisPlugin.java
jochen-vermeulen: Agreed. i changed it.

PR #229, opened 2019-02-13
Feat/ea 1468 ea 1471 organizations foaf depiction and address has geo

gsergiu:
renamed branch v1.0.2-zoho-import-updates to feat/EA-1468_EA-1471_organizations_foaf_depiction_and_address_has_geo


gsergiu:
Just a note that this branch also includes the code for https://europeana.atlassian.net/browse/EA-1490

metis-enrichment/metis-enrichment-service/src/main/java/eu/europeana/enrichment/utils/EnrichmentEntityDao.java
```diff
@@ -88,9 +91,18 @@
 
   private final MongoClient mongo;
 
+  /**
+   * @deprecated see
```
stzanakis: Why is this marked deprecated? It is just another option of connecting..

metis-enrichment/metis-enrichment-service/src/main/java/eu/europeana/enrichment/service/dao/WikidataAccessDao.java
```diff
@@ -117,14 +127,32 @@ public WikidataOrganization parse(File xmlFile) throws JAXBException, IOExceptio
     return parse(xml);
   }
 
+  /**
+   * This method converts XML string to Wikidata organization object.
```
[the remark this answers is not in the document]
gsergiu: I update the javadocs for all 3 parse methods

metis-enrichment/metis-enrichment-service/src/main/java/eu/europeana/enrichment/service/dao/WikidataAccessDao.java
```diff
@@ -138,47 +166,84 @@ public WikidataOrganization parse(String xml) throws JAXBException {
...
-      return endpoint.execDescribe(m);
+      return endpoint.execDescribe(m).getResource(uri);
     } catch (RiotException e) {
       LOGGER.error("Interrupted while querying Wikidata from the WikidataAccessDao", e);
     } finally {
       endpoint.close();
     }
 
-    return m;
+//    return m.getResource(uri);
```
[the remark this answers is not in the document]
gsergiu: I improved the exception handling and eliminated completely the last return statement

metis-enrichment/metis-enrichment-service/src/main/java/eu/europeana/enrichment/service/dao/WikidataAccessDao.java
```diff
@@ -138,47 +166,84 @@ public WikidataOrganization parse(String xml) throws JAXBException {
    * @param uri The Wikidata URI in string format
    * @return RDF model
    */
-  private Model getModelFromSPARQL(String uri) {
+  private Resource getModelFromSPARQL(String uri) {
+    Resource resource = fetchFromSPARQL(uri);
+    if (!isDuplicate(resource)) {
```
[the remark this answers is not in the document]
gsergiu: I improved exception handling, fetchFromSparql is not returning null anymore, but it throws exception. isDupplicate(resource) was updated to avoid NPE

metis-enrichment/metis-enrichment-service/src/main/java/eu/europeana/enrichment/service/dao/WikidataAccessDao.java
```diff
@@ -187,10 +252,11 @@ private synchronized void transform(Model m, Transformer t, StreamResult res)
    * 
    * @param inputFile The file containing the wikidata
    * @return WikidataOrganization object
+   * @throws IOException
    * @throws JAXBException
    */
   public WikidataOrganization parseWikidataOrganization(File inputFile)
-      throws JAXBException {
+      throws IOException, JAXBException {
```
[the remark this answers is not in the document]
gsergiu: in this method it is thrown, but in parse(String xml) is not thrown indeed. Updated on the right position

metis-enrichment/metis-enrichment-service/src/main/java/eu/europeana/enrichment/utils/EnrichmentEntityDao.java
gsergiu: Mongo connection without authentication is not allowed on production environments. The contructor with the connectionUrl  can be used both for production and development environments (the connection URL mongodb://localhost:27017 doing the same thing).

metis-enrichment/metis-enrichment-service/src/main/java/eu/europeana/enrichment/service/dao/WikidataAccessDao.java
```diff
@@ -117,14 +127,32 @@ public WikidataOrganization parse(File xmlFile) throws JAXBException, IOExceptio
...
   /**
    * This method converts XML string to Wikidata organization object.
    * 
    * @param xml The Wikidata organization object in string XML format
    * @return Wikidata organization object
    * @throws JAXBException
+   * @throws IOException
    */
-  public WikidataOrganization parse(String xml) throws JAXBException {
+  public WikidataOrganization parse(String xml) throws JAXBException, IOException {
```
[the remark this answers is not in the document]
gsergiu: this is true, removed IOException from throws list

metis-enrichment/metis-enrichment-service/src/main/java/eu/europeana/enrichment/service/dao/WikidataAccessDao.java
```diff
@@ -138,47 +166,84 @@ public WikidataOrganization parse(String xml) throws JAXBException {
...
+    }
+
+    //if duplication fetch main resource
+    StmtIterator iter = resource.listProperties(OWL.sameAs);
+    try {
+      while (iter.hasNext()) {
+        String sameAs = iter.next().getResource().getURI();
+        Resource r2 = fetchFromSPARQL(sameAs);
+        //check if main resource
+        if (!isDuplicate(r2)) {
```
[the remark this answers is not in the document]
gsergiu: isDupplicate(resource) was updated to avoid NPE

metis-enrichment/metis-enrichment-service/src/main/java/eu/europeana/enrichment/service/dao/WikidataAccessDao.java
```diff
@@ -95,8 +104,9 @@ public final void init(InputStream xslTemplate) throws WikidataAccessException {
    * @param uri The Wikidata URI in string format
    * @return String The Result of Wikidata query in XML format
    * @throws WikidataAccessException
+   * @throws IOException
    */
-  public StringBuilder getEntity(String uri) throws WikidataAccessException {
+  public StringBuilder getEntity(String uri) throws WikidataAccessException, IOException {
```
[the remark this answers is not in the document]
gsergiu: removed IOException from throws list

metis-enrichment/metis-enrichment-service/src/main/java/eu/europeana/enrichment/utils/EnrichmentEntityDao.java
```diff
@@ -92,7 +90,7 @@
   private final MongoClient mongo;
 
   /**
-   * @deprecated see 
+   * @deprecated use {@link #EnrichmentEntityDao(String)}
```
stzanakis: My comment was referring to remove the @deprecated since we it's not a deprecation, we can use both and in fact we use similar methods in other connections as well and not a String URI.

metis-enrichment/metis-enrichment-service/src/main/java/eu/europeana/enrichment/service/dao/WikidataAccessDao.java
```diff
@@ -197,25 +190,24 @@ private Resource getModelFromSPARQL(String uri) {
    * @return true if the retrieved resource is duplicated
    */
   private boolean isDuplicate(Resource resource) {
-    return (resource.hasProperty(OWL.sameAs) && !resource.hasProperty(RDFS.label));
+    return (resource!= null && resource.hasProperty(OWL.sameAs) && !resource.hasProperty(RDFS.label));
```
stzanakis: Wait are you sure about this change?
The isDuplicate is called with a negation, so !isDuplicate inside the getModelFromSPARQL method.
Which means that if resource is null this line here evaluates to false and the !isDuplicate evaluates to true, therefore returning a null value. That null value is passed in the transform method which with through NPE in that scenario..

metis-enrichment/metis-enrichment-service/src/main/java/eu/europeana/enrichment/service/dao/WikidataAccessDao.java
```diff
@@ -197,25 +190,24 @@ private Resource getModelFromSPARQL(String uri) {
...
     String sDescribe = "DESCRIBE <" + uri + ">";
 
     Model m = ModelFactory.createDefaultModel();
     QueryEngineHTTP endpoint = new QueryEngineHTTP(SPARQL, sDescribe);
     try {
       return endpoint.execDescribe(m).getResource(uri);
-    } catch (RiotException e) {
-      LOGGER.error("Interrupted while querying Wikidata from the WikidataAccessDao", e);
+    } catch (Exception e) {
+      //LOGGER.error("Interrupted while querying Wikidata from the WikidataAccessDao", e);
```
stzanakis: No need to log anymore?

metis-enrichment/metis-enrichment-service/src/main/java/eu/europeana/enrichment/service/dao/WikidataAccessDao.java
gsergiu: yes ... resource is dupplicated is exists at all (not null), has same as and has no label.
A resource that doesn't exists (is null) cannot be considered to be dupplicated.
This method is used to check if the resource URI is a kind of redirect (dupplicated resource) in wikidata.

metis-enrichment/metis-enrichment-service/src/main/java/eu/europeana/enrichment/service/dao/WikidataAccessDao.java
gsergiu: the exception is now converted and retrown, logging is needed only when the exception is really handled. I remove the commented code now

metis-enrichment/metis-enrichment-service/src/main/java/eu/europeana/enrichment/utils/EnrichmentEntityDao.java
gsergiu: yes, I understood the meaning. But I still think that this constructor shouldn't be used anymore and should be removed, as any code invoking this constructor will not work in production environments.

metis-enrichment/metis-enrichment-service/src/main/java/eu/europeana/enrichment/service/dao/WikidataAccessDao.java
stzanakis: I get that but the if it is null the transform method will throw an NPE on line 222

PR #239, opened 2019-03-05
MET-1741 Check if there is a current CLEANING state indexing plugin

stzanakis:


metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/service/OrchestratorService.java
```diff
@@ -539,12 +540,23 @@ public DatasetExecutionInformation getDatasetExecutionInformation(MetisUser meti
...
-            .of(PluginType.PUBLISH));
     AbstractMetisPlugin lastPreviewPlugin = workflowExecutionDao
         .getLastFinishedWorkflowExecutionPluginByDatasetIdAndPluginType(datasetId, EnumSet
             .of(PluginType.PREVIEW));
+    AbstractMetisPlugin lastPublishPlugin = workflowExecutionDao
+        .getLastFinishedWorkflowExecutionPluginByDatasetIdAndPluginType(datasetId, EnumSet
+            .of(PluginType.PUBLISH));
+
+    final WorkflowExecution runningOrInQueueExecution = workflowExecutionDao
+        .getRunningOrInQueueExecution(datasetId);
```
jochen-vermeulen: What if runningOrInQueueExecution is null? I don't think we can guarantee that we currently have an execution?

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/service/OrchestratorService.java
```diff
@@ -539,12 +540,23 @@ public DatasetExecutionInformation getDatasetExecutionInformation(MetisUser meti
...
+    final boolean isPreviewCleaning = runningOrInQueueExecution.getMetisPlugins().stream()
+        .filter(metisPlugin -> metisPlugin.getPluginType() == PluginType.PREVIEW)
+        .map(AbstractMetisPlugin::getPluginStatus)
+        .anyMatch(pluginStatus -> pluginStatus == PluginStatus.CLEANING);
```
jochen-vermeulen: I think there is something wrong with the logic... shouldn't we check that the status is cleaning OR running?

In general, the logic should be: [preview/publish]IsReadyForViewing = no preview/publish task is currently cleaning or running AND the last completed preview/publish task finished more than getSolrCommitPeriodInMins() ago.

jochen-vermeulen: In fact, because there is a lot of duplicate code here (the code for preview and publish is practically the same) maybe you can abstract that out?

PR #245, opened 2019-03-12
MET-1781 restrict timespan enrichments for very broad terms

stzanakis:


metis-enrichment/metis-enrichment-service/src/main/java/eu/europeana/enrichment/service/RedisInternalEnricher.java
```diff
@@ -58,6 +58,7 @@
 
   private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
   private static final List<EntityType> ENTITY_TYPES = createEntityTypeList();
+  private static final String REGEX_MATCHING_VERY_BROAD_TIMESPANS = ".*/ChronologicalPeriod$|.*/Time$|.*/(AD|BC)[1-9]x{3}$";
```
jochen-vermeulen: As discussed:
1. Create static final Pattern so that you don't compile the regex every time within the two loops.
2. Add prefix so that we are completely sure that no other records match the regex by accident.

PR #256, opened 2019-04-05
MET-1811: Allow requesting multiple pages in one overview request.

jochen-vermeulen:


metis-core/metis-core-rest/src/main/java/eu/europeana/metis/core/rest/OrchestratorController.java
```diff
@@ -446,16 +447,20 @@ public DatasetExecutionInformation getDatasetExecutionInformation(
...
   public ResponseListWrapper<ExecutionAndDatasetView> getWorkflowExecutionsOverview(
       @RequestHeader("Authorization") String authorization,
-      @RequestParam(value = "nextPage", required = false, defaultValue = "0") int nextPage)
+      @RequestParam(value = "nextPage", required = false, defaultValue = "0") int nextPage,
+      @RequestParam(value = "pageCount", required = false, defaultValue = "1") int pageCount)
       throws GenericMetisException {
     if (nextPage < 0) {
       throw new BadContentException(CommonStringValues.NEXT_PAGE_CANNOT_BE_NEGATIVE);
     }
+    if (pageCount < 1) {
```
stzanakis: We should also enforce a practical upper limit of the pageCount, because this call can be a dangerous operation.
So either enforce the pageCount or the pageCount based on the total returned results relative to the requestLimit value.

jochen-vermeulen: As discussed, this issue will be handled in a separate ticket: MET-1881.

PR #262, opened 2019-05-06
Met 1889 register reindexed and invalidated plugins

jochen-vermeulen:


metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/execution/WorkflowExecutor.java
```diff
@@ -180,71 +181,82 @@ private Date runInqueueOrRunningStateWorkflowExecution() {
...
+        plugin.execute(dpsClient, ecloudBasePluginParameters);
       }
-    }
 
-    // Start periodical check and wait for plugin to be done
-    long sleepTime = TimeUnit.SECONDS.toMillis(monitorCheckIntervalInSecs);
-    periodicCheckingLoop(sleepTime, abstractMetisPlugin);
+      // Start periodical check and wait for plugin to be done
+      long sleepTime = TimeUnit.SECONDS.toMillis(monitorCheckIntervalInSecs);
+      periodicCheckingLoop(sleepTime, plugin);
```
stzanakis: Please double check if the location of this method here is fine.

PR #267, opened 2019-06-13
Feat/met 2008 add quality annotation support in indexing

stzanakis:


metis-indexing/src/main/java/eu/europeana/indexing/fullbean/QualityAnnotationFieldInput.java
```diff
@@ -0,0 +1,37 @@
...
+ * @since 2019-06-13
+ */
+public class QualityAnnotationFieldInput implements
+    Function<QualityAnnotation, QualityAnnotationImpl> {
+
+  @Override
+  public QualityAnnotationImpl apply(QualityAnnotation qualityAnnotation) {
+    QualityAnnotationImpl qualityAnnotationImpl = new QualityAnnotationImpl();
+    qualityAnnotationImpl.setAbout(qualityAnnotation.getAbout());
+    if (qualityAnnotation.getCreatedList() != null) {
```
jochen-vermeulen: You don't have to check for null, here. The methods in FieldInputUtils check for null. Same for the two calls below.

metis-indexing/src/main/java/eu/europeana/indexing/solr/property/QualityAnnotationSolrCreator.java
```diff
@@ -0,0 +1,24 @@
...
+ * @since 2019-06-13
+ */
+public class QualityAnnotationSolrCreator implements PropertySolrCreator<QualityAnnotation> {
+
+  @Override
+  public void addToDocument(SolrInputDocument doc, QualityAnnotation qualityAnnotation) {
+    final EdmLabel tierTypeLabel;
+    if (qualityAnnotation.getAbout().contains("metadataTier")) {
+      tierTypeLabel = EdmLabel.METADATA_TIER;
+    } else {
```
jochen-vermeulen: A couple of small things:
1. Maybe we should check for '#metadataTier' (with hash) to reduce the (already remote) chance that a tier will be added with a name that also contains metadataTier.
2. I think we should also add a condition to the else block that checks the about. Maybe a different value appears for some reason, in which case we should probably skip this value.

metis-indexing/src/main/java/eu/europeana/indexing/solr/property/QualityAnnotationSolrCreator.java
```diff
@@ -0,0 +1,24 @@
...
+/**
+ * @author Simon Tzanakis (Simon.Tzanakis@europeana.eu)
+ * @since 2019-06-13
+ */
+public class QualityAnnotationSolrCreator implements PropertySolrCreator<QualityAnnotation> {
+
+  @Override
+  public void addToDocument(SolrInputDocument doc, QualityAnnotation qualityAnnotation) {
+    final EdmLabel tierTypeLabel;
+    if (qualityAnnotation.getAbout().contains("metadataTier")) {
```
jochen-vermeulen: Check for '#metadataTier' instead? And not contains but endsWith?

This would prevent (a very unlikely) scenario in which there is a quality annotation with an about value that contains but is not equal to 'metadataTier'.

metis-indexing/src/main/java/eu/europeana/indexing/solr/property/QualityAnnotationSolrCreator.java
jochen-vermeulen: I think we should also check the about value here. Just to be safe, and prevent an unknown about value to be characterized as a content tier value.

metis-indexing/src/main/java/eu/europeana/indexing/fullbean/QualityAnnotationFieldInput.java
```diff
@@ -18,12 +18,9 @@
...
     qualityAnnotationImpl.setAbout(qualityAnnotation.getAbout());
-    qualityAnnotationImpl.setDcTermsCreated(
-        FieldInputUtils.resourceOrLiteralListToArray(qualityAnnotation.getCreatedList()));
-    qualityAnnotationImpl.setOaHasTarget(
-        FieldInputUtils.resourceListToArray(qualityAnnotation.getHasTargetList()));
-    qualityAnnotationImpl.setOaHasBody(
-        FieldInputUtils.resourceListToArray(qualityAnnotation.getHasBodyList()));
+    qualityAnnotationImpl.setDcTermsCreated(qualityAnnotation.getCreated().getString());
+    qualityAnnotationImpl.setOaHasTarget(qualityAnnotation.getHasTarget().getResource());
+    qualityAnnotationImpl.setOaHasBody(qualityAnnotation.getHasBody().getResource());
```
jochen-vermeulen: Please check for null in these three lines.

stzanakis: Inderdaad!

PR #268, opened 2019-06-20
Feat/met 1988 perform tier calculation

jochen-vermeulen:


metis-indexing/src/main/java/eu/europeana/indexing/utils/RdfTier.java
```diff
@@ -0,0 +1,52 @@
...
+import eu.europeana.indexing.tiers.model.MediaTier;
+import eu.europeana.indexing.tiers.model.MetadataTier;
+import eu.europeana.indexing.tiers.model.Tier;
+
+/**
+ * This enum lists the content and metadata tiers that exist.
+ */
+public enum RdfTier {
+
+  METADATA_TIER_0("http://www.europeana.eu/schemas/epf/metadataTier0", MetadataTier.T0),
```
stzanakis: I believe the last part of the url should start with "#" and not a "/"?

metis-indexing/src/main/java/eu/europeana/indexing/tiers/model/MetadataTier.java
```diff
@@ -1,4 +1,4 @@
-package eu.europeana.indexing.solr.tiers.model;
+package eu.europeana.indexing.tiers.model;
 
 /**
  * This enum contains the different values of the metadata tier.
```
stzanakis: Why is the label field required?

metis-indexing/src/main/java/eu/europeana/indexing/tiers/model/MediaTier.java
```diff
@@ -1,4 +1,4 @@
-package eu.europeana.indexing.solr.tiers.model;
+package eu.europeana.indexing.tiers.model;
 
 /**
  * This enum contains the different values of the media tier.
```
stzanakis: Why is the label field required?

metis-indexing/src/main/java/eu/europeana/indexing/utils/RdfTier.java
jochen-vermeulen: Good catch! However, I don't know. This document has them with slash:
https://docs.google.com/document/d/1BXdrIbyh2ufiC4WvDvisX8v1M89pAEVSne_tWXpHRdo/edit
And so does the description of this ticket:
https://europeana.atlassian.net/browse/MET-2008
However, the following ticket has it with hash (#):
https://europeana.atlassian.net/browse/MET-1987
Finally, this ticket has both versions (the table vs. the code snippet): 
https://europeana.atlassian.net/browse/MET-1988
So what is the right answer? I'm asking Valentine.

metis-indexing/src/main/java/eu/europeana/indexing/tiers/model/MetadataTier.java
jochen-vermeulen: I anticipated their use, but I saw I now have them also in QualityAnnotationSolrCreator, where they belong. I will remove them.

metis-indexing/src/main/java/eu/europeana/indexing/tiers/model/MediaTier.java
jochen-vermeulen: I anticipated their use, but I saw I now have them also in QualityAnnotationSolrCreator, where they belong. I will remove them.

metis-indexing/src/main/java/eu/europeana/indexing/utils/RdfTier.java
jochen-vermeulen: As you can see in the ticket, the way I had it is correct.

stzanakis:
It has been merged in develop but because of conflict it doesn't show up as merged

PR #287, opened 2019-09-18
Debt/met 2193 fix sonarqube issues

stzanakis:


metis-enrichment/metis-enrichment-service/src/main/java/eu/europeana/enrichment/service/RedisInternalEnricher.java
```diff
@@ -222,16 +232,20 @@ private void loadEntity(EntityType entityType, MongoTerm term, Jedis jedis) {
               parents.toArray(new String[]{}));
         }
         if (termList.getOwlSameAs() != null) {
-          for (String sameAs : termList.getOwlSameAs()) {
-            jedis.hset(entityType.cachedEntityPrefix + CACHED_SAMEAS, sameAs, term.getCodeUri());
-          }
+          loadSameAs(entityType, term, jedis, termList.getOwlSameAs());
```
jochen-vermeulen: It's a bit inconsistent to have a small private method for this while similar operations are done inline. I assume the issue here is the indentation level (too deep nesting)? May I suggest undoing this change and instead move the try-catch block to the calling method (loadEntities())?

metis-media-service/src/main/java/eu/europeana/metis/mediaprocessing/extraction/ThumbnailGenerator.java
```diff
@@ -371,12 +372,18 @@ ImageMetadata parseCommandResponse(List<String> response) throws MediaExtraction
 
   private static List<String> extractDominantColors(List<String> results, int skipLines) {
     final Pattern pattern = Pattern.compile("#([0-9A-F]{6})");
+
+    final boolean aResultDoesNotMatch = results.stream().skip(skipLines)
+        .sorted(Collections.reverseOrder()).limit(COMMAND_RESULT_MAX_COLORS).map(pattern::matcher)
+        .anyMatch(m -> !m.find());
```
jochen-vermeulen: This seems to create duplicate code. Maybe better is to define a stream variable and then reuse it? And also, you can make the anyMatch check nicer, I think? 

How about this?

```
final Stream<Matcher> matchers =  results.stream().skip(skipLines)
        .sorted(Collections.reverseOrder()).limit(COMMAND_RESULT_MAX_COLORS).map(pattern::matcher);
if (!matchers.allMatch(matcher::find) {
  throw new IllegalStateException("Invalid color line found.");
}
return matchers.map(matcher -> matcher.group(1)).collect(Collectors.toList());
```

metis-validation/metis-validation-service/src/main/java/eu/europeana/validation/service/Validator.java
```diff
@@ -120,15 +120,9 @@ private ValidationResult validate() {
...
-          Node nresult = nresults.item(i);
-          if ("failed-assert".equals(nresult.getLocalName())) {
-            String nodeId = nresult.getAttributes().getNamedItem(NODE_ID_ATTR) == null ? null
-                : nresult.getAttributes().getNamedItem(NODE_ID_ATTR).getTextContent();
-            return constructValidationError(document,
-                "Schematron error: " + nresult.getTextContent().trim(), nodeId);
-          }
-        }
+        final ValidationResult validationResult = checkNodeListForErrors(nresults);
+        if (validationResult != null)
```
jochen-vermeulen: Should use curly braces { } for this if statement.

metis-validation/metis-validation-service/src/main/java/eu/europeana/validation/service/Validator.java
```diff
@@ -120,15 +120,9 @@ private ValidationResult validate() {
...
-        for (int i = 0; i < nresults.getLength(); i++) {
-          Node nresult = nresults.item(i);
-          if ("failed-assert".equals(nresult.getLocalName())) {
-            String nodeId = nresult.getAttributes().getNamedItem(NODE_ID_ATTR) == null ? null
-                : nresult.getAttributes().getNamedItem(NODE_ID_ATTR).getTextContent();
-            return constructValidationError(document,
-                "Schematron error: " + nresult.getTextContent().trim(), nodeId);
-          }
-        }
+        final ValidationResult validationResult = checkNodeListForErrors(nresults);
```
jochen-vermeulen: Renaming variable to 'errorResult' would make this more understandable I think?

metis-validation/metis-validation-service/src/main/java/eu/europeana/validation/service/ClasspathResourceResolver.java
```diff
@@ -26,13 +25,17 @@
   public LSInput resolveResource(String type, String namespaceURI, String publicId, String systemId,
       String baseURI) {
     try {
-      LSInput input = new ClasspathLSInput();
+      //Read file from system
       String fullPath = new File(prefix, systemId).getAbsolutePath();
-      InputStream stream = Files.newInputStream(Paths.get(fullPath));
+      final byte[] bytes = Files.readAllBytes(Paths.get(fullPath));
```
jochen-vermeulen: Is there a reason to first read it completely into memory completely? Why not use the stream approach that was there originally?

metis-validation/metis-validation-service/src/main/java/eu/europeana/validation/service/Validator.java
stzanakis: Sure

metis-validation/metis-validation-service/src/main/java/eu/europeana/validation/service/ClasspathResourceResolver.java
stzanakis: Sonarqube complains about not closing the stream, I found this solution only that would make sonarqube complain to go away.

metis-enrichment/metis-enrichment-service/src/main/java/eu/europeana/enrichment/service/RedisInternalEnricher.java
stzanakis: Yeah that makes more sense

metis-core/metis-core-common/src/main/java/eu/europeana/metis/core/workflow/plugins/IndexToPreviewPluginMetadata.java
```diff
@@ -21,7 +21,7 @@ public ExecutablePluginType getExecutablePluginType() {
     return pluginType;
   }
 
-  public boolean getUseAlternativeIndexingEnvironment() {
+  public boolean isUseAlternativeIndexingEnvironment() {
```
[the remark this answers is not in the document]
stzanakis: That is the convention that sonarqube rule wants.
Either we disable the rule do what we want or we follow the rule.

metis-core/metis-core-rest/src/main/java/eu/europeana/metis/core/rest/config/ConfigurationPropertiesHolder.java
```diff
@@ -238,7 +238,7 @@ public String getMetisCoreBaseUrl() {
     return metisCoreBaseUrl;
   }
 
-  public boolean getMetisUseAlternativeIndexingEnvironment() {
+  public boolean isMetisUseAlternativeIndexingEnvironment() {
```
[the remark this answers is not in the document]
stzanakis: That is the convention that sonarqube rule wants.
Either we disable the rule do what we want or we follow the rule.

metis-validation/metis-validation-service/src/main/java/eu/europeana/validation/service/Validator.java
stzanakis: done

PR #295, opened 2019-10-31
MET-2281 Guard last executable plugins from null pointer

stzanakis:


metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/service/OrchestratorService.java
```diff
@@ -560,10 +564,12 @@ public DatasetExecutionInformation getDatasetExecutionInformation(MetisUser meti
         .getLatestSuccessfulExecutablePlugin(datasetId, HARVEST_TYPES, false).getPlugin();
```
jochen-vermeulen: You should probably test this occurrence of the same method call as well!

stzanakis: good point!

PR #298, opened 2019-11-12
debt/MET-2302_TechnicalDebt

SrishtiSingh-eu:
debt/MET-2302_TechnicalDebt

### DIFF
--- a/metis-normalization/src/main/java/eu/europeana/normalization/languages/Languages.java
+++ b/metis-normalization/src/main/java/eu/europeana/normalization/languages/Languages.java
@@ -19,7 +19,7 @@ import org.w3c.dom.Element;
  *
  * @author Nuno Freire (nfreire@gmail.com)
  */
-public class Languages {
+public final class Languages {
 
   private static Languages instance;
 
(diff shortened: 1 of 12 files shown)


metis-validation/metis-validation-rest/src/main/java/eu/europeana/validation/rest/config/Application.java
```diff
@@ -100,7 +100,7 @@ public void addResourceHandlers(ResourceHandlerRegistry registry) {
 
     /**
      * Creates {@link org.springframework.web.multipart.MultipartResolver} for application context
-     * @return
+     * @return commonsMultipartResolver
```
stzanakis: Add description here not a camel case name of the class

metis-normalization/src/main/java/eu/europeana/normalization/languages/Languages.java
```diff
@@ -19,7 +19,7 @@
  *
  * @author Nuno Freire (nfreire@gmail.com)
  */
-public class Languages {
+public final class Languages {
```
stzanakis: Build fails now probably because of this addition

PR #306, opened 2019-12-20
MET-2378 Safeguard pagination variable from overflow

stzanakis:


metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/dao/WorkflowExecutionDao.java
```diff
@@ -871,7 +872,9 @@ private Pagination createPagination(int firstPage, int pageCount,
     final int pageSize = getWorkflowExecutionsPerRequest();
     final int maxResultCount =
         ignoreMaxServedExecutionsLimit ? Integer.MAX_VALUE : getMaxServedExecutionListLength();
-    final int total = Math.min((firstPage + pageCount) * pageSize, maxResultCount);
+    final int minimum = Math.min((firstPage + pageCount) * pageSize, maxResultCount);
+    //Safeguard possible overflow
+    final int total = minimum < 0 ? maxResultCount : minimum;
```
jochen-vermeulen: This is not exactly a clean solution though... perhaps by accident the overflow is equal to 0? Or 1?
How about making the pageCount parameter to be nullable (i.e. Integer instead of int) and interpret null as a no-pagecount-limit?

stzanakis: In our case that is practically not possible since integers on a overflow get wrap around.
Perhaps though a nullable parameter is more visible

PR #308, opened 2019-12-23
Feat/met 1420 improve saving xslt functionality

stzanakis:


metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/dao/WorkflowExecutionDao.java
```diff
@@ -788,6 +790,19 @@ public WorkflowExecution getByTaskExecution(Date startedDate, PluginType pluginT
     return ExternalRequestUtil.retryableExternalRequestConnectionReset(query::first);
   }
 
+  public WorkflowExecution getByXsltId(String xsltId) {
```
jochen-vermeulen: Maybe rename to 'getAnyWorkflowExecutionForXsltId'? Or 'getAnyByXsltId'? There is not just one execution for a xslt in general.

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/service/DatasetService.java
```diff
@@ -202,6 +204,21 @@ public void updateDataset(MetisUser metisUser, Dataset dataset, String xsltStrin
     datasetDao.update(dataset);
   }
 
+  private void cleanDatasetXslt(ObjectId xsltId) {
```
jochen-vermeulen: I'm a bit worried about backwards compatibility. Is it your intention to do a database migration? If not, I have a concern. 

What if an xslt is used, but it is not referred to by ID from a transformation plugin, but by URL (because it is old). Then this method will conclude that the xslt is not referenced at all, and clean it up. With the loss of history as a consequence.

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/service/DatasetService.java
```diff
@@ -193,6 +194,7 @@ public void updateDataset(MetisUser metisUser, Dataset dataset, String xsltStrin
     if (xsltString == null) {
       dataset.setXsltId(storedDataset.getXsltId());
     } else {
+      cleanDatasetXslt(storedDataset.getXsltId());
```
jochen-vermeulen: Is it your intention to do a separate full database cleanup? If not, I am worried that a lot of historic records are never cleaned up because it only looks at the current xslt reference (that it is replacing) instead of all the ones that came before as well. What is your plan for those?

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/service/DatasetService.java
```diff
@@ -341,6 +358,10 @@ public DatasetXslt createDefaultXslt(MetisUser metisUser, String xsltString)
     authorizer.authorizeWriteDefaultXslt(metisUser);
     DatasetXslt datasetXslt = null;
     if (xsltString != null) {
+      final DatasetXslt latestDefaultXslt = datasetXsltDao.getLatestDefaultXslt();
+      if (latestDefaultXslt != null) {
+        cleanDatasetXslt(latestDefaultXslt.getId());
```
jochen-vermeulen: Same as above.

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/service/DatasetService.java
stzanakis: Yes the idea is that we should update those scripts to include the cleanup of xslts as well. The data for those are generally not that big or crucial, so we don't really need a daemon running every day. Even if we forget cleaning one time the data will be very minimal. Nonetheless we could include a cleanup during a reindexing operation or even an "extra" daemon in metis-core that could be performed monthly or so..

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/dao/WorkflowExecutionDao.java
stzanakis: ok

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/service/DatasetService.java
stzanakis: Yes the idea is that we should update those scripts to include the cleanup of xslts as well. The data for those are generally not that big or crucial, so we don't really need a daemon running every day. Even if we forget cleaning one time the data will be very minimal. Nonetheless we could include a cleanup during a reindexing operation or even an "extra" daemon in metis-core that could be performed monthly or so..

PR #310, opened 2020-01-15
Feat/met 2404 include redirection during indexing

stzanakis:


metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/service/DatasetService.java
```diff
@@ -126,6 +127,17 @@ public Dataset createDataset(MetisUser metisUser, Dataset dataset)
...
+
+      // Verify references to old datasetIds
+      if (dataset.getOldDatasetIds() != null) {
+        for (String id : dataset.getOldDatasetIds()) {
+          if (datasetDao.getDatasetByDatasetId(id) == null) {
+            throw new BadContentException(
+                String.format("Old datasetId %s doesn't exist", id));
+          }
+        }
+      }
```
jochen-vermeulen: I don't think we can assume that the IDs necessarily exist? What if the old dataset has been removed (manually as part of cleanup)? But we still want the redirects to exist? Is this a use case we should consider? Or should we move to a system where we only invalidate datasets (and not remove them) so that IDs are not reused?

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/service/DatasetService.java
```diff
@@ -178,7 +190,17 @@ public void updateDataset(MetisUser metisUser, Dataset dataset, String xsltStrin
...
+
+    // Verify references to old datasetIds
+    if (dataset.getOldDatasetIds() != null) {
+      for (String id : dataset.getOldDatasetIds()) {
+        if (datasetDao.getDatasetByDatasetId(id) == null) {
+          throw new BadContentException(
+              String.format("Old datasetId %s doesn't exist", id));
+        }
+      }
     }
```
jochen-vermeulen: See above.

metis-data-checker/metis-data-checker-service/src/main/java/eu/europeana/metis/data/checker/service/persistence/RecordDao.java
```diff
@@ -41,13 +41,13 @@
   /**
    * Constructor with required fields.
    *
-   * @param connectionProvider
    * @throws IndexingException In case of problems setting up the indexer.
    */
   @Autowired
   public RecordDao(AbstractConnectionProvider connectionProvider) throws IndexingException {
```
jochen-vermeulen: Sonarlint: "document the parameter(s): connectionProvider"

metis-data-checker/metis-data-checker-service/src/main/java/eu/europeana/metis/data/checker/service/persistence/RecordDao.java
```diff
@@ -61,16 +61,15 @@ public RecordDao(AbstractConnectionProvider connectionProvider) throws IndexingE
    * Persist a record in mongo and solr
    *
    * @param rdf The record
-   * @param recordDate
    * @throws IndexingException In case indexing failed.
    */
   public void createRecord(RDF rdf, Date recordDate) throws IndexingException {
```
jochen-vermeulen: Sonarlint: "document the parameter(s): recordDate"

metis-indexing/src/main/java/eu/europeana/indexing/ClientsConnectionProvider.java
```diff
@@ -1,37 +1,42 @@
...
-   * @param mongoClient The Mongo client to be used. Cannot be null.
+   *
+   * @param edmMongoClient The Mongo client to be used. Cannot be null.
+   * @param recordRedirectDao The record redirect dao.
    * @param solrClient The Solr client to be used. Cannot be null.
    * @throws SetupRelatedIndexingException In case either of the two clients are null.
    */
-  ClientsConnectionProvider(EdmMongoServer mongoClient, SolrClient solrClient)
+  ClientsConnectionProvider(EdmMongoServer edmMongoClient, RecordRedirectDao recordRedirectDao,
+      SolrClient solrClient)
```
jochen-vermeulen: Should you not do the null checks for the record redirect dao, just like for the other connections?

metis-indexing/src/main/java/eu/europeana/indexing/SettingsConnectionProvider.java
```diff
@@ -70,7 +72,8 @@ public SettingsConnectionProvider(IndexingSettings settings)
     // Create mongo connection.
     try {
       this.mongoClient = createMongoClient(settings);
-      this.mongoServer = setUpMongoConnection(settings, this.mongoClient);
+      this.edmMongoClient = setUpEdmMongoConnection(settings, this.mongoClient);
+      this.recordRedirectDao = setUpRecordRedirectDaoConnection(settings, this.mongoClient);
```
jochen-vermeulen: This assumes that the redirect and edm mongo databases are always on the same server. Is this a valid assumption??

metis-indexing/src/main/java/eu/europeana/indexing/IndexingSettings.java
```diff
@@ -23,6 +23,7 @@
   // Mongo settings
   private final List<ServerAddress> mongoHosts = new ArrayList<>();
   private String mongoDatabaseName;
+  private String recordRedirectDatabaseName;
```
jochen-vermeulen: There is no setter for this field!

metis-indexing/src/main/java/eu/europeana/indexing/FullBeanPublisher.java
```diff
@@ -159,4 +189,68 @@ private void publishToSolr(RdfWrapper rdf, FullBeanImpl fullBean) throws Indexin
...
+  }
+
+  private void createRedirection(String newIdentifier, String oldIdentifier,
+      Date recordRedirectDate) {
+    final RecordRedirect recordRedirect = new RecordRedirect(newIdentifier, oldIdentifier,
+        recordRedirectDate);
+    recordRedirectDao.createUpdate(recordRedirect);
+    //Update the previous redirect item in db that has newId == oldIdentifier
+    final RecordRedirect recordRedirectByNewId = recordRedirectDao
+        .getRecordRedirectByNewId(oldIdentifier);
```
jochen-vermeulen: There could be multiple of these, I think. We should update all of them.

metis-indexing/src/main/java/eu/europeana/indexing/FullBeanPublisher.java
```diff
@@ -159,4 +189,68 @@ private void publishToSolr(RdfWrapper rdf, FullBeanImpl fullBean) throws Indexin
...
+    SolrDocumentList documents = getSolrDocuments(queryParamMap);
+    if (!CollectionUtils.isEmpty(documents)) {
+      return null;
+    }
+
+    //Check matches with older dataset identifiers
+    if (!CollectionUtils.isEmpty(datasetIdsForRedirection)) {
+      //The incoming structure of the identifier is /datasetId/recordId
+      final String[] splitRecordIdentifier = recordIdIncludingDatasetId.split("/");
+      String datasetId = splitRecordIdentifier[1];
```
jochen-vermeulen: Apparently we don't need this variable. Is that right? The code does not use it.

metis-indexing/src/main/java/eu/europeana/indexing/FullBeanPublisher.java
```diff
@@ -159,4 +189,68 @@ private void publishToSolr(RdfWrapper rdf, FullBeanImpl fullBean) throws Indexin
       throw new RecordRelatedIndexingException(SOLR_SERVER_PUBLISH_ERROR, e);
     }
   }
+
+  private String searchMatchingRecordForRedirection(String recordIdIncludingDatasetId,
+      List<String> datasetIdsForRedirection)
+      throws IndexerRelatedIndexingException, RecordRelatedIndexingException {
+    //Check if record with same record identifier exists
+    final Map<String, String> queryParamMap = new HashMap<>();
+    queryParamMap.put("q",
```
jochen-vermeulen: Use 'fq' instead of 'q'? And set q="*.*". We don't want the server to be concerned with relevance scores.

metis-indexing/src/main/java/eu/europeana/indexing/FullBeanPublisher.java
```diff
@@ -159,4 +189,68 @@ private void publishToSolr(RdfWrapper rdf, FullBeanImpl fullBean) throws Indexin
       throw new RecordRelatedIndexingException(SOLR_SERVER_PUBLISH_ERROR, e);
     }
   }
+
+  private String searchMatchingRecordForRedirection(String recordIdIncludingDatasetId,
+      List<String> datasetIdsForRedirection)
+      throws IndexerRelatedIndexingException, RecordRelatedIndexingException {
+    //Check if record with same record identifier exists
```
jochen-vermeulen: I think this code skips redirects if the record already exists? Because we assume that the redirects would have already been added during the previous indexing? Is this an assumption we can make? What if the user made a mistake and changed the previous dataset list and wishes to index again?
I think we should either add comments to highlight this assumption, or, even better, skip this step and recompute the redirects.

metis-indexing/src/main/java/eu/europeana/indexing/FullBeanPublisher.java
```diff
@@ -159,4 +189,68 @@ private void publishToSolr(RdfWrapper rdf, FullBeanImpl fullBean) throws Indexin
...
+      //The incoming structure of the identifier is /datasetId/recordId
+      final String[] splitRecordIdentifier = recordIdIncludingDatasetId.split("/");
+      String datasetId = splitRecordIdentifier[1];
+      String recordId = splitRecordIdentifier[2];
+
+      final String combinedQueryForRedirectedDatasetIds = datasetIdsForRedirection.stream()
+          .map(datasetIdForRedirection -> String
+              .format("%s:/%s/%s", EdmLabel.EUROPEANA_ID.toString(), datasetIdForRedirection,
+                  recordId)).collect(Collectors.joining(" OR "));
+      queryParamMap.put("q", combinedQueryForRedirectedDatasetIds);
```
jochen-vermeulen: Also here we should probably use fq instead?

metis-indexing/src/main/java/eu/europeana/indexing/FullBeanPublisher.java
```diff
@@ -159,4 +189,68 @@ private void publishToSolr(RdfWrapper rdf, FullBeanImpl fullBean) throws Indexin
...
+    }
+
+    documents = getSolrDocuments(queryParamMap);
+    if (!CollectionUtils.isEmpty(documents)) {
+      //Get identifier of the first document found
+      return (String) documents.get(0).getFieldValue(EdmLabel.EUROPEANA_ID.toString());
```
jochen-vermeulen: I don't think we can assume that there is just one result! We should return all of them and then create/update mappings for all.

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/service/DatasetService.java
stzanakis: I'm actually not sure on how to answer that.
I assumed that if we did a cleanup of a dataset we should remove it's reference from other datasets.
The validation is here to verify that a user will not add a reference to a dataset that doesn't exist.
Maybe we should analyze this a bit

metis-core/metis-core-common/src/main/java/eu/europeana/metis/core/workflow/plugins/IndexToPreviewPluginMetadata.java
```diff
@@ -11,6 +14,7 @@
   private static final ExecutablePluginType pluginType = ExecutablePluginType.PREVIEW;
   private boolean useAlternativeIndexingEnvironment;
   private boolean preserveTimestamps;
+  private List<String> datasetIdsForRedirection = new ArrayList<>();
```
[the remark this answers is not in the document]
stzanakis: Yes I wasn't sure about the naming, this seems a bit more clear.

metis-data-checker/metis-data-checker-service/src/main/java/eu/europeana/metis/data/checker/service/persistence/RecordDao.java
stzanakis: Done

metis-data-checker/metis-data-checker-service/src/main/java/eu/europeana/metis/data/checker/service/persistence/RecordDao.java
stzanakis: Done

metis-indexing/src/main/java/eu/europeana/indexing/ClientsConnectionProvider.java
stzanakis: Indeed.

metis-indexing/src/main/java/eu/europeana/indexing/SettingsConnectionProvider.java
stzanakis: Yes this is a valid assumption. At least so far.

metis-indexing/src/main/java/eu/europeana/indexing/IndexingSettings.java
stzanakis: Oops

metis-indexing/src/main/java/eu/europeana/indexing/FullBeanPublisher.java
stzanakis: No that shouldn't be possible

metis-indexing/src/main/java/eu/europeana/indexing/FullBeanPublisher.java
stzanakis: Correct, I was holding it for the next ticket, will remove it for now.

metis-indexing/src/main/java/eu/europeana/indexing/FullBeanPublisher.java
stzanakis: Normally it should be just one, but could be a provider mistake I assume, so may we should expect that as well.

metis-indexing/src/main/java/eu/europeana/indexing/FullBeanPublisher.java
stzanakis: Yes that is the assumption. There reasoning here is improvement in performance

stzanakis: Created a ticket for this to have a bit more time to think about this operation

metis-indexing/src/main/java/eu/europeana/indexing/FullBeanPublisher.java
stzanakis: I'm leaving this out since I'm not sure how to effectively use the fq/q combination, and plus I read that fq is caching the result which might not be what we want. Open for discussion later on.

metis-indexing/src/main/java/eu/europeana/indexing/FullBeanPublisher.java
stzanakis: I'm leaving this out since I'm not sure how to effectively use the fq/q combination, and plus I read that fq is caching the result which might not be what we want. Open for discussion later on.

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/service/DatasetService.java
stzanakis: For this I'm removing the references when a dataset gets deleted.

PR #313, opened 2020-01-29
Feat/met 1486 redirects implementation

stzanakis:


metis-indexing/src/main/java/eu/europeana/indexing/FullBeanPublisher.java
```diff
@@ -159,4 +256,226 @@ private void publishToSolr(RdfWrapper rdf, FullBeanImpl fullBean) throws Indexin
...
+          .put(EdmLabel.PROVIDER_AGGREGATION_EDM_IS_SHOWN_AT.toString(), isShownAtList);
+    }
+    return listsToCombineMapsIsShownAtAndBy;
+  }
+
+  private String generateQueryForFields(Map<String, List<String>> listsToCombine) {
+    final String combinationResult = listsToCombine.entrySet().stream()
+        .map(entry -> generateOrOperationFromList(entry.getKey(), entry.getValue()))
+        .collect(Collectors.joining(" AND "));
+    return String.format("(%s)", combinationResult);
```
jochen-vermeulen: There is a Collectors.joining method that accepts a prefix and suffix. So no need to do the subsequent String.format.

metis-indexing/src/main/java/eu/europeana/indexing/FullBeanPublisher.java
```diff
@@ -159,4 +256,226 @@ private void publishToSolr(RdfWrapper rdf, FullBeanImpl fullBean) throws Indexin
...
+  }
+
+  private String generateOrOperationFromList(String queryFieldName, List<String> items) {
+    final String orCombinedItems = items.stream()
+        .map(item -> String.format("%s:%s", queryFieldName, item))
+        .collect(Collectors.joining(" OR "));
+    return String.format("(%s)", orCombinedItems);
```
jochen-vermeulen: See above.

jochen-vermeulen: Also: there is the following option to search in a list of options:
[https://stackoverflow.com/questions/2613592/how-can-i-search-on-a-list-of-values-using-solr-lucene]
Perhaps this makes it better/shorter/more readable.

metis-indexing/src/main/java/eu/europeana/indexing/FullBeanPublisher.java
```diff
@@ -159,4 +256,226 @@ private void publishToSolr(RdfWrapper rdf, FullBeanImpl fullBean) throws Indexin
...
+    //Combine different list of fields in groups
+    final String firstQueryGroup = generateQueryForFields(firstMapOfLists);
+    final String secondQueryGroup = generateQueryForFields(secondMapOfLists);
+    final String thirdQueryGroup = generateQueryForFields(thirdMapOfLists);
+    final String fourthQueryGroup = generateQueryForFields(fourthMapOfLists);
+
+    //Join all groups
+    final String combinedQuery = Stream
+        .of(firstQueryGroup, secondQueryGroup, thirdQueryGroup, fourthQueryGroup)
+        .filter(StringUtils::isNotBlank).collect(Collectors.joining(" OR "));
```
jochen-vermeulen: I don't think this works, as the group string is never blank. At minimum, it is equal to '()'.

metis-indexing/src/main/java/eu/europeana/indexing/FullBeanPublisher.java
```diff
@@ -159,4 +256,226 @@ private void publishToSolr(RdfWrapper rdf, FullBeanImpl fullBean) throws Indexin
...
+    final List<String> descriptions = rdfWrapper.getProviderProxyDescriptions().stream()
+        .map(description -> {
+          if (StringUtils.isNotBlank(description.getString())) {
+            return description.getString();
+          } else if (description.getResource() != null && StringUtils
+              .isNotBlank(description.getResource().getResource())) {
+            return description.getResource().getResource();
+          }
+          return null;
+        }).collect(Collectors.toList());
```
jochen-vermeulen: Should you not filter for null? This is apparently a possible outcome, and it would screw up the following logic.

metis-indexing/src/main/java/eu/europeana/indexing/FullBeanPublisher.java
```diff
@@ -159,4 +256,226 @@ private void publishToSolr(RdfWrapper rdf, FullBeanImpl fullBean) throws Indexin
...
+    Predicate<SolrDocument> predicate = document -> {
+      boolean doesDocumentMatchWithQuery = false;
+      for (Entry<String, List<String>> entry : queryMatchingFieldsAndFirstQueryGroup.getRight()
+          .entrySet()) {
+        doesDocumentMatchWithQuery = document.getFieldValues(entry.getKey()).stream()
+            .map(String.class::cast)
+            .anyMatch(entry.getValue()::contains);
+      }
+      return doesDocumentMatchWithQuery;
+    };
```
jochen-vermeulen: I just realized something that might need to be changed in the document (as well as here in the code). I think there is another result that should be prioritized, possibly even before this one, and that is anything that comes out of `queryForDatasetIds`. If you find an exact match of the record id in one of the datasets to redirect from, perhaps this should be the most trusted result?

metis-indexing/src/main/java/eu/europeana/indexing/FullBeanPublisher.java
```diff
@@ -159,4 +256,226 @@ private void publishToSolr(RdfWrapper rdf, FullBeanImpl fullBean) throws Indexin
...
+    //Preprocess sub-query and replace documents based on the result
+    SolrDocumentList solrDocuments = getSolrDocuments(queryParamMap);
+    final SolrDocumentList subQueryResults = solrDocuments.stream().filter(predicate)
+        .collect(Collectors.toCollection(SolrDocumentList::new));
+    if (!CollectionUtils.isEmpty(subQueryResults)) {
```
jochen-vermeulen: subQueryResults cannot be null at this point, so you can just use subQueryResults.isEmpty()

metis-indexing/src/main/java/eu/europeana/indexing/FullBeanPublisher.java
```diff
@@ -159,4 +256,226 @@ private void publishToSolr(RdfWrapper rdf, FullBeanImpl fullBean) throws Indexin
...
+      solrDocuments = subQueryResults;
+    }
+    //Return all identifiers found
+    if (!CollectionUtils.isEmpty(solrDocuments)) {
+      return solrDocuments.stream()
+          .map(document -> (String) document.getFieldValue(EdmLabel.EUROPEANA_ID.toString()))
+          .collect(Collectors.toList());
+    }
+    return new ArrayList<>();
```
jochen-vermeulen: solrDocuments cannot be null at this point, so you can remove the if statement and return the result of the stream operation.

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/service/WorkflowExecutionFactory.java
```diff
@@ -91,6 +118,30 @@ private AbstractExecutablePlugin createWorkflowPlugin(
...
+    } catch (PluginExecutionNotAllowed pluginExecutionNotAllowed) {
+      //It means we need to do redirects because a new harvest was performed
+      performRedirects = true;
+      LOGGER.info("Predecessor plugin computation failed, so we want to create redirects",
+          pluginExecutionNotAllowed);
+    }
+
+    final PluginWithExecutionId<ExecutablePlugin> latestSuccessfulExecutablePlugin = workflowExecutionDao
+        .getLatestSuccessfulExecutablePlugin(dataset.getDatasetId(), EnumSet
+            .of(executablePluginType), true);
```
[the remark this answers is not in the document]
stzanakis: If you mean reindexing, then no. A reindexing should not be our base to apply redirects or not.

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/service/WorkflowExecutionFactory.java
```diff
@@ -91,6 +118,30 @@ private AbstractExecutablePlugin createWorkflowPlugin(
...
+    boolean performRedirects = false;
+    try {
+      workflowUtils
+          .computePredecessorPlugin(executablePluginType, null, dataset.getDatasetId());
+    } catch (PluginExecutionNotAllowed pluginExecutionNotAllowed) {
+      //It means we need to do redirects because a new harvest was performed
+      performRedirects = true;
+      LOGGER.info("Predecessor plugin computation failed, so we want to create redirects",
+          pluginExecutionNotAllowed);
+    }
```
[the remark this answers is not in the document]
stzanakis: Alright, perhaps I would just get the latests successful harvest and check the date with the latest index and compare.

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/service/WorkflowExecutionFactory.java
```diff
@@ -91,6 +118,30 @@ private AbstractExecutablePlugin createWorkflowPlugin(
...
+
+    final PluginWithExecutionId<ExecutablePlugin> latestSuccessfulExecutablePlugin = workflowExecutionDao
+        .getLatestSuccessfulExecutablePlugin(dataset.getDatasetId(), EnumSet
+            .of(executablePluginType), true);
+    if (dataset.getUpdatedDate()
+        .compareTo(latestSuccessfulExecutablePlugin.getPlugin().getFinishedDate()) >= 0
+        && !CollectionUtils.isEmpty(dataset.getDatasetIdsToRedirectFrom())) {
```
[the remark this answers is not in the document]
stzanakis: Well yes but the code block above should turn reflect that and turn true.

metis-indexing/src/main/java/eu/europeana/indexing/FullBeanPublisher.java
```diff
@@ -105,7 +137,49 @@ private static void setUpdateAndCreateTime(IdBean current, FullBean updated, Dat
...
+   * <ul>
+   * <li>{@link IndexerRelatedIndexingException} In case an error occurred during publication.</li>
+   * <li>{@link SetupRelatedIndexingException} in case an error occurred during indexing setup</li>
+   * <li>{@link RecordRelatedIndexingException} in case an error occurred related to record
+   * contents</li>
+   * </ul>
+   */
+  private void publish(RdfWrapper rdf, Date recordDate, List<String> datasetIdsToRedirectFrom,
+      boolean performRedirects)
+      throws IndexingException {
```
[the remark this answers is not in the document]
stzanakis: I'm following code convention here.

metis-indexing/src/main/java/eu/europeana/indexing/FullBeanPublisher.java
```diff
@@ -116,10 +190,16 @@ public void publish(RdfWrapper rdf, Date recordDate) throws IndexingException {
...
+    List<String> recordsForRedirection = checkAndApplyRedirects(rdf, recordDate,
+        datasetIdsToRedirectFrom, performRedirects);
+
     // Publish to Mongo
     final FullBeanImpl savedFullBean;
     try {
-      savedFullBean = new FullBeanUpdater(fullBeanPreprocessor).update(fullBean, recordDate, mongoClient);
+      savedFullBean = new FullBeanUpdater(fullBeanPreprocessor)
+          .update(fullBean, recordDate, recordsForRedirection.stream().findFirst().orElse(null),
+              edmMongoClient);
```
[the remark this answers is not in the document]
stzanakis: Good point, I'll have to ask.

metis-indexing/src/main/java/eu/europeana/indexing/FullBeanPublisher.java
stzanakis: Updated

metis-indexing/src/main/java/eu/europeana/indexing/FullBeanPublisher.java
stzanakis: I made a check inside the method generateQueryForFields

metis-indexing/src/main/java/eu/europeana/indexing/FullBeanPublisher.java
stzanakis: Fixed

metis-indexing/src/main/java/eu/europeana/indexing/FullBeanPublisher.java
stzanakis: Don't think is such cases there would be a difference but we can add this

metis-indexing/src/main/java/eu/europeana/indexing/FullBeanPublisher.java
stzanakis: Sure

metis-indexing/src/main/java/eu/europeana/indexing/FullBeanPublisher.java
stzanakis: Sure

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/service/WorkflowExecutionFactory.java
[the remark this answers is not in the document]
stzanakis: This should work, it would check if the latest harvest is the same with the linked to the requested plugin from db. So we are comparing if there is a new harvest executed therefore a redirect for an indexing operation would be applicable..

metis-indexing/src/main/java/eu/europeana/indexing/IndexerImpl.java
```diff
@@ -53,14 +53,26 @@
...
+        preprocessRecord(record);
+        publisher
+            .publishWithRedirects(new RdfWrapper(record), recordDate, datasetIdsForRedirection);
+      }
+    } else {
+      for (RDF record : records) {
+        preprocessRecord(record);
+        publisher.publish(new RdfWrapper(record), recordDate, datasetIdsForRedirection);
+      }
     }
```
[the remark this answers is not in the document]
stzanakis: Fixed

metis-indexing/src/main/java/eu/europeana/indexing/FullBeanPublisher.java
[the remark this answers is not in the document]
stzanakis: As discussed I'm leaving this as is for now

metis-indexing/src/main/java/eu/europeana/indexing/mongo/FullBeanUpdater.java
```diff
@@ -38,18 +38,21 @@ public FullBeanUpdater(TriConsumer<FullBeanImpl, FullBeanImpl, Date> fullBeanPre
...
    * @param newEntity The new entity (to take the new values from).
    * @param recordDate The date that would represent the created/updated date of a record
+   * @param recordIdToRedirectFrom The record that we are redirecting from, this is used so that the created and updated timestamp can be maintained
    * @param mongoServer The mongo server.
    * @return The updated entity.
    */
-  public final FullBeanImpl update(FullBeanImpl newEntity, Date recordDate, MongoServer mongoServer) {
-    return update(newEntity,null, recordDate, mongoServer);
+  public final FullBeanImpl update(FullBeanImpl newEntity, Date recordDate, String recordIdToRedirectFrom,
+      MongoServer mongoServer) {
```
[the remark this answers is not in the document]
stzanakis: Not a bad idea, I'll change to that.

metis-indexing/src/main/java/eu/europeana/indexing/mongo/property/MongoPropertyUpdaterFactory.java
```diff
@@ -98,9 +100,11 @@ private MongoPropertyUpdaterFactory() {
       throw new IllegalArgumentException("Object does not have an 'about' value.");
     }
 
-    // Create object.
+    // Find object with the same about, it will be either the redirected one or the previous inserted.
     final Supplier<Query<T>> queryCreator = () -> mongoServer.getDatastore().find(objectClass)
-        .field(ABOUT_FIELD).equal(aboutGetter.apply(updated));
+        .field(ABOUT_FIELD).equal(
+            StringUtils.isNotBlank(recordIdToRedirectFrom) ? recordIdToRedirectFrom
+                : aboutGetter.apply(updated));
```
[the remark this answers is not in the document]
stzanakis: See above.

metis-indexing/src/main/java/eu/europeana/indexing/FullBeanPublisher.java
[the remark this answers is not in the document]
stzanakis: Decided as the lowest value, so the oldest date.

metis-indexing/src/main/java/eu/europeana/indexing/FullBeanPublisher.java
stzanakis: I gave it a try

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/service/WorkflowExecutionFactory.java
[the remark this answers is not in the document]
jochen-vermeulen: You're right. There is another issue though. The method workflowExecutionDao.getLatestSuccessfulExecutablePlugin can return null (e.g. if there is no valid data). You should allow for that.

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/service/WorkflowExecutionFactory.java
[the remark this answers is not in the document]
jochen-vermeulen: After giving it some thought, I think I understand what you mean. But I do think that this is not the best way to test what you want to test. This method tests other things as well, and may have other criteria and is not meant for this purpose. Perhaps it is better to define exactly what it is we want to test and then test it. 

I guess that we want to test whether the previous harvest occurred after the previous index. You're getting the previous index below. So just get the previous harvest here and compare times?

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/service/WorkflowExecutionFactory.java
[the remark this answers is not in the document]
jochen-vermeulen: OK. It's indeed safer that way.

metis-indexing/src/main/java/eu/europeana/indexing/FullBeanPublisher.java
```diff
@@ -186,20 +177,22 @@ private void publish(RdfWrapper rdf, Date recordDate, List<String> datasetIdsToR
...
             : (FullBeanPublisher::setUpdateAndCreateTime);
 
-    //The list can be empty if redirects are not applied if no redirect were found
-    List<String> recordsForRedirection = checkAndApplyRedirects(rdf, recordDate,
-        datasetIdsToRedirectFrom, performRedirects);
+    //The list can be empty if redirects are not applied or if no redirect were found
+    List<Pair<String, Date>> recordsForRedirection = RecordRedirectsUtil
+        .checkAndApplyRedirects(recordRedirectDao::createUpdate,
+            recordRedirectDao::getRecordRedirectByNewId, rdf, recordDate, datasetIdsToRedirectFrom,
+            performRedirects, this::getSolrDocuments);
```
jochen-vermeulen: Why not pass recordRedirectDao as a parameter either to the method or on construction of a RecordRedirectsUtil instance?

metis-indexing/src/main/java/eu/europeana/indexing/FullBeanPublisher.java
jochen-vermeulen: I don't think this problem is solved yet.

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/service/WorkflowExecutionFactory.java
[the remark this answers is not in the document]
stzanakis: Okay I'm changing to the original suggestion.

metis-indexing/src/main/java/eu/europeana/indexing/FullBeanPublisher.java
stzanakis: Functionally I wanted transparency on the utils class, but I have changed it to provide a recordRedirectDao instead.

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/service/WorkflowExecutionFactory.java
```diff
@@ -120,23 +119,26 @@ private AbstractExecutablePlugin createWorkflowPlugin(
...
             .of(executablePluginType), true);
-    if (dataset.getUpdatedDate()
-        .compareTo(latestSuccessfulExecutablePlugin.getPlugin().getFinishedDate()) >= 0
-        && !CollectionUtils.isEmpty(dataset.getDatasetIdsToRedirectFrom())) {
+
+    //Since previous plugin execution, a new harvest occurred after the latest plugin supplied
+    final boolean doesNewHarvestAfterPluginExist =
+        latestSuccessfulHarvest != null && latestSuccessfulExecutablePlugin != null
+            && latestSuccessfulHarvest.getPlugin().getFinishedDate()
+            .compareTo(latestSuccessfulExecutablePlugin.getPlugin().getFinishedDate()) >= 0;
```
jochen-vermeulen: I think this should also be true if there is NO latest successful executable plugin? For instance if this is the first harvest to occur for this dataset?

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/service/WorkflowExecutionFactory.java
```diff
@@ -120,23 +119,26 @@ private AbstractExecutablePlugin createWorkflowPlugin(
...
+    final boolean isDatasetInformationUpdated =
+        latestSuccessfulExecutablePlugin != null && dataset.getUpdatedDate()
+            .compareTo(latestSuccessfulExecutablePlugin.getPlugin().getFinishedDate()) >= 0
+            && !CollectionUtils.isEmpty(dataset.getDatasetIdsToRedirectFrom());
+
+    boolean performRedirects = false;
+    if (doesNewHarvestAfterPluginExist || isDatasetInformationUpdated) {
       performRedirects = true;
     }
     return performRedirects;
```
jochen-vermeulen: These 5 lines can be replaced by
`return doesNewHarvestAfterPluginExist || isDatasetInformationUpdated;`

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/service/WorkflowExecutionFactory.java
stzanakis: Indeed as discussed I've added it

PR #315, opened 2020-02-04
MET-2427 Update morphia deprecated methods

stzanakis:


metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/dao/ScheduledWorkflowDao.java
```diff
@@ -206,13 +210,21 @@ public boolean deleteAllByDatasetId(String datasetId) {
...
+    return ExternalRequestUtil
+        .retryableExternalRequestConnectionReset(() -> getListOfQuery(query, findOptions));
 
   }
 
+  private <T> List<T> getListOfQuery(Query<T> query, FindOptions findOptions) {
+    try (MorphiaCursor<T> cursor = query.find(findOptions)) {
+      return cursor.toList();
+    }
+  }
```
jochen-vermeulen: Since you're only calling this method from within a retryable connection, perhaps you can move the 'ExternalRequestUtil.retryableExternalRequestConnectionReset' to this private method as well? Same for the other DAO above.

stzanakis: sure

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/dao/DatasetDao.java
```diff
@@ -410,11 +421,18 @@ public String checkAndCreateDatasetInEcloud(Dataset dataset) throws ExternalTask
...
+        .limit(getDatasetsPerRequest());
+    return ExternalRequestUtil
+        .retryableExternalRequestConnectionReset(() -> getListOfQuery(query, findOptions));
+  }
+
+  private <T> List<T> getListOfQuery(Query<T> query, FindOptions findOptions) {
+    try (MorphiaCursor<T> cursor = query.find(findOptions)) {
+      return cursor.toList();
+    }
   }
```
jochen-vermeulen: Sorry if my comment was not clear. I think you should do the same for this one as well: move the 'ExternalRequestUtil.retryableExternalRequestConnectionReset' to this private method.

PR #316, opened 2020-02-10
Bug/met 2419 change namespace in dereferencing service

jochen-vermeulen:


metis-enrichment/metis-enrichment-client/src/main/java/eu/europeana/enrichment/utils/EntityMergeEngine.java
```diff
@@ -146,6 +148,14 @@ private static AgentType convertAndAddAgent(Agent agent, List<AgentType> destina
     // dateList
     agentType.setDateList(ItemExtractorUtils.extractLabelResources(agent.getDate(), Date::new));
 
+    // placeOfBirth
+    agentType.setPlaceOfBirthList(
```
acerdas84: do you have a format standard (in a file that I can import) for the code ?

jochen-vermeulen: We're using the google java format style. 
https://github.com/google/styleguide/blob/gh-pages/intellij-java-google-style.xml

acerdas84:
Is the placeOfDeath and placeOfBirth specifically related to this bug ?

metis-enrichment/metis-enrichment-client/src/main/java/eu/europeana/enrichment/utils/EntityMergeEngine.java
acerdas84: Thanks

PR #317, opened 2020-02-25
Feat/met 2413 more greedy dereferencing

jochen-vermeulen:


metis-enrichment/metis-enrichment-service/src/main/java/eu/europeana/enrichment/service/RedisInternalEnricher.java
```diff
@@ -355,44 +356,58 @@ private void loadEntities(EntityType entityType, MongoTerm term, Jedis jedis,
...
+            (uri, jedis) -> getEntityWrapperFromSameAs(uri, jedis, CACHED_TIMESPAN),
+            (uri, jedis) -> getEntityWrapperFromSameAs(uri, jedis, CACHED_PLACE)
+    );
+
+    // Get the first non-null result.
+    try (final Jedis jedis = redisProvider.getJedis()) {
+      for (GetterByUri getter : getters) {
+        final EntityWrapper result = getter.getByUri(requestedUri, jedis);
+        if (result != null) {
+          return result;
```
stzanakis: This actually changes behavior of the retrieved result.
Before we were getting the last found and now we are getting the first found.
Are we aware of and fine with this change?

metis-dereference/metis-dereference-service/src/main/java/eu/europeana/metis/dereference/service/MongoDereferenceService.java
```diff
@@ -129,16 +134,43 @@ public EnrichmentResultList dereference(String resourceId)
...
+    final Map<String, EnrichmentBase> result;
     if (iterations > 0) {
       result = GraphUtils.breadthFirstSearch(resourceId, resource.getLeft(),
           resource.getRight().getIterations(), valueResolver, this::extractBroaderResources);
     } else {
-      result = Collections.singleton(resource.getLeft());
+      result = Collections.singletonMap(resourceId, resource.getLeft());
     }
 
+    // Get links from dereferenced objects and try to enrich them. Get: 1) entities that have any of
```
stzanakis: Does this comment block need update? I read it a bit differently on the ticket and in code below seems much simpler..?

metis-enrichment/metis-enrichment-service/src/main/java/eu/europeana/enrichment/service/RedisInternalEnricher.java
jochen-vermeulen: I was aware of it. At that point, you expect only one of these to yield any result at all anyway, so it doesn't matter. And even if it does, we would like to return the record with the matching rdf:about value, instead of the record with the matching sameAs value.

metis-dereference/metis-dereference-service/src/main/java/eu/europeana/metis/dereference/service/MongoDereferenceService.java
jochen-vermeulen: Well, let's say that I added it like this precisely because the ticket is not 100% clear. I'm going by the latest comment of Valentine in the ticket, which seems the most clear to me. And it is this that should be implemented by the code below the comment. I would suggest leaving the wordy comment in place, just so that we know what the deceptively simple-looking code is actually doing behind the scenes.

PR #322, opened 2020-03-10
MET-2570 Fix redirect issues

stzanakis:


metis-common/src/main/java/eu/europeana/metis/mongo/RecordRedirectDao.java
```diff
@@ -108,6 +108,10 @@ public RecordRedirect getById(String id) {
                     .toList());
   }
 
+  public Datastore getDatastore() {
+    return datastore;
+  }
+
```
jochen-vermeulen: I think this method is not used  anywhere? Nor should it be. Perhaps we can remove it?

stzanakis: It's used in the tools script, it's just a convenience here..

PR #324, opened 2020-03-13
MET-1527 Clean up database connection classes

stzanakis:


metis-common/src/main/java/eu/europeana/metis/mongo/RecordRedirectDao.java
```diff
@@ -21,28 +21,40 @@
...
   public RecordRedirectDao(MongoClient mongoClient, String databaseName, boolean createIndexes) {
-    this.mongoClient = mongoClient;
+    createDatastore(mongoClient, databaseName);
+    if (createIndexes) {
+      LOGGER.info("Initializing database indices");
+      datastore.ensureIndexes();
+    }
+  }
 
+  private void createDatastore(MongoClient mongoClient, String databaseName) {
```
jochen-vermeulen: Is it necessary to have this in a separate method? You can integrate this into the constructor and then make the class variable datastore final (or make this method static).
You can do the same in the MorhpiaDatastoreProviderImpl.

metis-common/src/main/java/eu/europeana/metis/mongo/RecordRedirectDao.java
```diff
@@ -21,28 +21,40 @@
...
-   * <p>Initializes the datastore using the provided mongo client</p>
+   * Constructor to initialize the mongo mappings/collections and the {@link Datastore} connection.
+   * This constructor is meant to be used mostly for when the creation of the database is required.
    *
    * @param mongoClient the mongo client connection
    * @param databaseName the database name of the record redirect database
-   * @param createIndexes flag that enables/disables the creation of the indexes in the database
+   * @param createIndexes flag that initiates the database/indices
    */
   public RecordRedirectDao(MongoClient mongoClient, String databaseName, boolean createIndexes) {
```
jochen-vermeulen: I can see that the constructor with the createIndexes parameter is still called from the metis-redirects-migration project. Shouldn't that be changed?

stzanakis: It should be changed.

metis-common/src/main/java/eu/europeana/metis/mongo/RecordRedirectDao.java
stzanakis: I though in this way of encapsulating that part, makes the constructor more clear.
Do you feel it's not clear or what you are suggesting is better? I would then change it..

jochen-vermeulen: It's not important. I think a good way forward would be to make the private method static and return the datastore, so that it can be assigned by the constructor to a final class variable. But we can do that some other time.

PR #328, opened 2020-03-24
Debt/met 2592 technical debt metis

jochen-vermeulen:


metis-authentication/metis-authentication-common/src/main/java/eu/europeana/metis/authentication/user/MetisUserRecord.java
```diff
@@ -0,0 +1,178 @@
...
+/**
+ * The Metis user containing all parameters.
+ * <p>This class is used as a JPA class to the postgresql database</p>
+ *
+ * @author Simon Tzanakis (Simon.Tzanakis@europeana.eu)
+ * @since 2017-10-27
+ */
+@Entity
+@Table(name = "metis_users")
+public class MetisUserRecord {
```
stzanakis: If we are going for this approach, to splitting model classes and representation classes. We should agree on naming suffices.. I wouldn't vote for the current suffix of "Record"..

metis-dereference/metis-dereference-service/src/main/java/eu/europeana/metis/dereference/service/utils/RdfRetriever.java
```diff
@@ -96,6 +96,8 @@ private static String retrieveFromSource(String resourceId, String url, int redi
       throws IOException {
 
     // Make the connection and retrieve the result.
+    // Note: we have no choice but to follow the provided URL.
+    @SuppressWarnings("findsecbugs:URLCONNECTION_SSRF_FD")
```
stzanakis: Normally we were marking the issue in sonarqube so that it's not considered in the analysis.
I don't know if applying that annotation would work, but we can give it a try otherwise if it doesn't work we might want to remove it. 
Same for the other annotations

jochen-vermeulen: It does work, I tested it. I used it for Clio as well (and it also worked there).
This way, there is less chance of it being lost when the code is changed.
And it even generates a (much lower priority) warning in SonarQube that ensures that we don't completely forget about it.

metis-authentication/metis-authentication-common/src/main/java/eu/europeana/metis/authentication/user/MetisUserRecord.java
jochen-vermeulen: After discussion, I changed the name of this class to MetisUserModel.

stzanakis: Great

metis-dereference/metis-dereference-service/src/main/java/eu/europeana/metis/dereference/service/utils/RdfRetriever.java
stzanakis: Cool okay let's see how this pans out.

stzanakis:
Ready to be merge. DO NOT MERGE TILL AFTER NEXT RELEASE

PR #330, opened 2020-03-27
Bug/met 2564 separate enrichment and dereference caches

jochen-vermeulen:


metis-enrichment/metis-enrichment-proxy-rest/src/main/java/eu/europeana/enrichment/cache/proxy/config/Application.java
```diff
@@ -68,7 +68,10 @@ EnrichmentEntityDao getEntityDao() {
 
   @Bean(name = "redisInternalEnricher")
   RedisInternalEnricher getRedisInternalEnricher() {
-    return new RedisInternalEnricher(getEntityDao(), getRedisProvider(), true);
+    final RedisInternalEnricher enricher = new RedisInternalEnricher(getEntityDao(),
+            getRedisProvider());
+    new Thread(enricher::recreateCache).start();
```
stzanakis: I think this shouldn't be executed on every redeployment/restart.
We should have something like a rest call that enables a flag and enforces the recreation on the next restart.

jochen-vermeulen: OK, I fixed it.

stzanakis:
Ready to be merge. DO NOT MERGE TILL AFTER NEXT RELEASE

PR #334, opened 2020-05-04
Debt/met 2609 update libraries

stzanakis:


metis-dereference/metis-dereference-rest/pom.xml
```diff
@@ -13,10 +13,6 @@
       <groupId>javax.xml.bind</groupId>
       <artifactId>jaxb-api</artifactId>
     </dependency>
-    <dependency>
-      <groupId>org.glassfish.jaxb</groupId>
-      <artifactId>jaxb-runtime</artifactId>
-    </dependency>
```
jochen-vermeulen, line 19: This I've added for java 11 support. Perhaps we should keep these? This is the implementation of xml transformations in java. As such, it doesn't seem to be directly referenced from the code (it references the dependency above it), but it may be needed nonetheless.
This goes for all occurrences of this library (also in other pom files). Did you check this?

metis-dereference/metis-dereference-rest/pom.xml
```diff
@@ -67,7 +63,6 @@
       <groupId>javax.servlet</groupId>
       <artifactId>javax.servlet-api</artifactId>
       <version>${version.servlet.api}</version>
-      <scope>provided</scope>
```
jochen-vermeulen, line 70: Why can't we assume that this is provided by the application container? This also goes for all cases where you remove this scope.

stzanakis: Well the new version is not part of tomcat or cloud foundry build pack.
So I figured why risk not having it and not just always have it available.

metis-dereference/metis-dereference-rest/pom.xml
stzanakis: At the moment it seems that it's not required and until we go to java 11 it will take some time from what we've seen based on the progress..
But yes you are right, when we do move to java 11 we will need to some how provide the xml logic from a lib, I think though that we need to provide more libs that just this one.

metis-dereference/metis-dereference-rest/pom.xml
jochen-vermeulen: OK, sounds fair.

metis-dereference/metis-dereference-rest/pom.xml
jochen-vermeulen: I've been running our code on java 11 for a while now and needed nothing else than just these two libraries (this one and the one above). I'm not saying that our code is completely ready for java 11, but all the services that I usually work on and run seem to be behaving as expected now. I would like to keep this library if possible for that reason.

stzanakis: Ok I've added it back.

PR #338, opened 2020-05-14
MET-2666: Create new builder class for EnrichmentClient functionality.

jochen-vermeulen:


metis-enrichment/metis-enrichment-client/src/main/java/eu/europeana/enrichment/rest/client/EnrichmentWorkerBuilder.java
```diff
@@ -0,0 +1,124 @@
...
+
+    // Make sure that the worker can do something.
+    if (StringUtils.isBlank(dereferenceUrl) && StringUtils.isBlank(enrichmentUrl)) {
+      throw new IllegalStateException(
+              "Either dereferencing or enrichment (or both) must be enabled.");
+    }
+
+    // Create the request factory
+    final HttpComponentsClientHttpRequestFactory requestFactory = new HttpComponentsClientHttpRequestFactory();
+    requestFactory.setConnectTimeout(Math.max(connectTimeout, 0));
```
stzanakis, line 102: You need to set the default values here if there is none specified.. Seems that the defaults are not currently used

jochen-vermeulen: Oops.. yes, i forgot to assign those to the right variables.

PR #339, opened 2020-05-15
MET-2678 Update list of embeddable media urls and provide pattern mat…

stzanakis:
…ching

### DIFF
--- a/metis-indexing/src/main/java/eu/europeana/indexing/tiers/media/EmbeddableMedia.java
+++ b/metis-indexing/src/main/java/eu/europeana/indexing/tiers/media/EmbeddableMedia.java
@@ -53,20 +53,21 @@ final class EmbeddableMedia {
       "http://www.euscreen.eu/item.html"
   );
 
-  // Create pattern that will contain all special regex characters and with that
-  // we'll find those characters on other Strings to properly escape them
-  static final String ESCAPED_REGEX_SPECIAL_CHARACTERS = "<([{^\\-=$!|]})?*+.>".chars()
-      .mapToObj(c -> "\\" + (char) c).collect(Collectors.joining());
-  static final Pattern escapedCharactersPattern = Pattern
-      .compile("[" + ESCAPED_REGEX_SPECIAL_CHARACTERS + "]");
-
-  // Create patterns from the urls, escape all regex special characters,
-  // we want to use wildcards(*) so we unescape them and add a wildcard at the end of each url
+  // Create patterns from the urls, quote url, wildcards are allowed in the pattern so we do not quote those
+  // and we also add a wildcard at the end of each url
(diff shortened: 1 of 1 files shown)


metis-indexing/src/main/java/eu/europeana/indexing/tiers/media/EmbeddableMedia.java
```diff
@@ -5,30 +5,54 @@
...
+      "http://sounds.bl.uk/embed/",
+      //EUScreen
+      "http://www.euscreen.eu/item.html"
   );
 
+  // Create patterns, update wildcards to proper format and add a wildcard at the end of each url
+  private static final Collection<Pattern> PATTERNS = URL_MATCHING_LIST.stream()
+      .map(prefix -> prefix.replace("*", ".*"))
+      .map(prefix -> prefix.concat(".*")).map(Pattern::compile)
+      .collect(Collectors.toList());
```
jochen-vermeulen, line 54: This doesn't always work as a regex because `?` is  a reserved character, and it occurs in some of the URLS. So it should be escaped. And it is too easy to later add URLs with other escaped characters (e.g. `-` or `&`) without realizing it.

Perhaps a better sequence (using the `Pattern.quote()` method twice) is:
```
private static final Collection<Pattern> PATTERNS = URL_MATCHING_LIST.stream()
      .map(Pattern::quote)
      .map(prefix -> prefix.replace(Pattern.quote("*"), ".*"))
      .map(prefix -> prefix.concat(".*")).map(Pattern::compile)
      .collect(Collectors.toList());
```
And I think we could also use some comments above somewhere for future developers explaining how to maintain and add to this list.

stzanakis: The quote example you wrote doesn't work, because the quote wraps a pattern with a prefix and suffix so that it's translated as literal but it doesn't add escape characters..
I see you point though, will try to find another way to solve this

jochen-vermeulen: I can see that the solution works. I have another one to suggest that doesn't depend on having a constant string with all special characters in it:
1. split the url on occurrences of '*'
2. escape (i.e. quote) the parts
3. concatenate them with .* in between and after
Do you think this would work? Do you think this would be better?

Separately: I think we should we put a '\A' before to make sure that the string starts with the URL, instead of also matching some weird URL where it occurs somewhere in the middle... what do you think?

stzanakis: It's a good alternative, I will try it out

jochen-vermeulen: It seems to work well.

PR #340, opened 2020-05-27
Feat/met 2583 load dereference vocabularies from GitHub

jochen-vermeulen:


metis-dereference/metis-dereference-common/src/main/java/eu/europeana/metis/dereference/Vocabulary.java
```diff
@@ -62,15 +83,26 @@
...
+    return Collections.unmodifiableSet(this.uris);
   }
 
-  public void setUri(String uri) {
-    this.uri = uri;
+  public void setUris(Collection<String> uris) {
+    this.uris = new HashSet<>(uris);
+
+    // Also take care of backwards compatibility
+    final String sampleUrl = uris.iterator().next();
```
stzanakis, line 95: Isn't it better to use the this.uris instead of the method parameter?

metis-dereference/metis-dereference-common/src/main/java/eu/europeana/metis/dereference/Vocabulary.java
```diff
@@ -62,15 +83,26 @@
...
+  public Set<String> getUris() {
+    return Collections.unmodifiableSet(this.uris);
   }
 
-  public void setUri(String uri) {
-    this.uri = uri;
+  public void setUris(Collection<String> uris) {
+    this.uris = new HashSet<>(uris);
+
+    // Also take care of backwards compatibility
```
stzanakis, line 94: How about we encapsulate this in a private method and mark it as deprecated so that we know we need to remove it?

jochen-vermeulen: Sure.

metis-dereference/metis-dereference-common/src/main/java/eu/europeana/metis/dereference/Vocabulary.java
jochen-vermeulen: I don't think it matters in this case... but no problem.

PR #341, opened 2020-05-27
MET-2526 Update spring version and deprecated methods

stzanakis:


pom.xml
```diff
@@ -95,11 +94,12 @@
...
-    <version.spring>5.1.2.RELEASE</version.spring>
-    <version.spring.context>5.1.2.RELEASE</version.spring.context>
-    <version.spring.beans>5.1.2.RELEASE</version.spring.beans>
-    <version.spring.test>5.1.2.RELEASE</version.spring.test>
+    <version.spring.security>5.3.2.RELEASE</version.spring.security>
+    <version.spring>5.2.6.RELEASE</version.spring>
+    <version.spring.test>${version.spring}</version.spring.test>
+    <version.spring.context>${version.spring}</version.spring.context>
+    <version.spring.beans>${version.spring}</version.spring.beans>
+    <version.spring.test>${version.spring}</version.spring.test>
```
jochen-vermeulen, line 102: I think you have the same property twice?

metis-core/metis-core-rest/src/test/java/eu/europeana/metis/core/rest/TestDatasetController.java
```diff
@@ -105,11 +105,11 @@ void createDataset() throws Exception {
 
     datasetControllerMock.perform(post("/datasets")
         .header("Authorization", TestObjectFactory.AUTHORIZATION_HEADER)
-        .contentType(MediaType.APPLICATION_JSON_UTF8)
-        .accept(MediaType.APPLICATION_JSON_UTF8)
+        .contentType(MediaType.APPLICATION_JSON)
```
jochen-vermeulen, line 108: Why drop the restriction that the result is UTF8? Isn't that allowing a change in the API? Have you tested the frontend with that?

pom.xml
stzanakis: Fixed!

metis-core/metis-core-rest/src/test/java/eu/europeana/metis/core/rest/TestDatasetController.java
stzanakis: Well other than the fact that they are deprecated, they were also causing some of the tests to fail..
And yes I tested the endpoint and it works..

Not related to this specific question but in general we still need to deploy the applications in test to be sure the dependencies are working

PR #343, opened 2020-05-29
MET-2721: Three way choice for dataset fitness for publication.

jochen-vermeulen:


metis-core/metis-core-common/src/main/java/eu/europeana/metis/core/dataset/PublicationFitness.java
```diff
@@ -0,0 +1,11 @@
+package eu.europeana.metis.core.dataset;
+
+/**
+ * Whether a dataset is fit for publication. {@link #PARTIALLY_FIT} means that some records may be
+ * unfit for publication.
+ */
+public enum PublicationFitness {
```
stzanakis, line 7: Maybe we should make this a static nested enum inside the dataset class, what do you think?

jochen-vermeulen: Sure. Why not.

stzanakis:
This is ready to be merge after next release.

PR #352, opened 2020-06-23
Feat/met 2795 handle post processing depublish

stzanakis:


metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/dao/DepublishedRecordDao.java
```diff
@@ -147,59 +150,107 @@ long countDepublishedRecordsForDataset(String datasetId) {
...
+   * @param sortField The sorting field. Cannot be null.
+   * @param sortDirection The sorting direction. Cannot be null.
+   * @param depublicationState The depublication state of the records. Can be null.
+   * @param searchQuery Search query for the record ID. Can be null.
+   * @return A (possibly empty) list of depublished records.
    */
-  public void markAllAsNotDepublished(String datasetId){
+  public Set<String> getAllDepublishedRecords(String datasetId,
+      DepublishedRecordSortField sortField, SortDirection sortDirection,
+      DepublicationStatus depublicationState, String searchQuery) {
```
jochen-vermeulen, line 189: And this method is never called with a non-null searchQuery.

metis-core/metis-core-common/src/main/java/eu/europeana/metis/core/dataset/DepublishedRecord.java
```diff
@@ -79,13 +82,14 @@ public void setRecordId(String recordId) {
...
+    return depublicationStatus;
   }
 
-  public void setDepublicationState(
-          DepublicationState depublicationState) {
-    this.depublicationState = depublicationState;
+  public void setDepublicationStatus(DepublicationStatus depublicationStatus) {
+    this.depublicationStatus =
+        depublicationStatus == null ? DepublicationStatus.PENDING_DEPUBLICATION
+            : depublicationStatus;
```
jochen-vermeulen, line 92: This seems like overkill? This method is called just once, and not with a null value.

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/dao/DepublishedRecordDao.java
```diff
@@ -147,59 +150,107 @@ long countDepublishedRecordsForDataset(String datasetId) {
...
+      DepublicationStatus depublicationStatus, @Nullable Date depublicationDate) {
+
+    // Create query.
+    final Query<DepublishedRecord> query = morphiaDatastoreProvider.getDatastore()
+        .createQuery(DepublishedRecord.class);
+    query.field(DepublishedRecord.DATASET_ID_FIELD).equal(datasetId);
+    //Convert status if provided status is null
+    final DepublicationStatus depublicationStatusChecked =
+        depublicationStatus == null ? DepublicationStatus.PENDING_DEPUBLICATION
+            : depublicationStatus;
```
jochen-vermeulen, line 243: This seems overkill? Why not require that this parameter is non-null? It is never  called with null anyway.

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/dao/DepublishedRecordDao.java
```diff
@@ -147,59 +150,107 @@ long countDepublishedRecordsForDataset(String datasetId) {
...
 
     // Convert result to right object.
     return result.stream().map(DepublishedRecordView::new).collect(Collectors.toList());
   }
 
   /**
-   * This method marks all depublished records of a given dataset as not depublished (and removes
-   * the depublished date).
+   * Get all depublished records for a given dataset.
+   * <p>This method is to be used with caution sicne it doesn't have a limit on the returned items.
```
jochen-vermeulen, line 177: Spelling error!

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/execution/WorkflowPostProcessor.java
```diff
@@ -32,11 +47,45 @@ public WorkflowPostProcessor(DepublishedRecordDao depublishedRecordDao) {
...
+        final Optional<AbstractMetisPlugin> metisPluginWithType = workflowExecutionToUpdate
+            .getMetisPluginWithType(latestSuccessfulPlugin.getPlugin().getPluginType());
+        if (metisPluginWithType.isPresent()) {
+          metisPluginWithType.get().setDataStatus(DataStatus.DELETED);
+          workflowExecutionDao.updateWorkflowPlugins(workflowExecutionToUpdate);
+        }
+      }
+    } else { //Set depublish status if depublishing records
+      depublishedRecordDao.markRecordIdsWithDepublicationStatus(datasetId,
+          DepublicationStatus.DEPUBLISHED, plugin.getStartedDate());
```
jochen-vermeulen, line 88: This is not right. It is also not required (yet), but still. This should check the actual depublished records from the topology and only mark those. Perhaps it is best to remove this code here for now and revisit this properly during the ticket that deals with depublication of record ids?

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/dao/DepublishedRecordDao.java
```diff
@@ -147,59 +150,107 @@ long countDepublishedRecordsForDataset(String datasetId) {
...
+   * <p>A {@link DepublicationStatus#PENDING_DEPUBLICATION} unsets the depublication date</p>
+   * <p>A {@link DepublicationStatus#DEPUBLISHED} sets the depublication date with the one
+   * provided</p>
+   *
+   * @param datasetId the dataset for which to do this
+   * @param depublicationStatus the depublication status. Null status is translated to {@link
+   * DepublicationStatus#PENDING_DEPUBLICATION}
+   * @param depublicationDate the depublication date. Can be null
+   */
+  public void markRecordIdsWithDepublicationStatus(String datasetId,
```
jochen-vermeulen, line 233: Should you bring this method back in the original state? I think this will only ever be called with DepublicationStatus.PENDING_DEPUBLICATION (see my remark in the WorkflowPostProcessor. Or do you plan to make this method generic enough to cover the other use case as well?

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/service/DepublishedRecordService.java
```diff
@@ -27,6 +27,7 @@
 @Service
 public class DepublishedRecordService {
 
+  private static final Pattern PATTERN_PER_LINE = Pattern.compile("\\R");
```
jochen-vermeulen, line 30: Maybe give clearer name? LINE_SEPARATION_PATTERN for instance?

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/service/WorkflowExecutionFactory.java
```diff
@@ -247,6 +258,17 @@ private void setupXsltIdForPluginMetadata(Dataset dataset,
     pluginMetadata.setLanguage(dataset.getLanguage().name().toLowerCase(Locale.US));
   }
 
+  private void setupDepublishPluginMetadata(Dataset dataset,
```
jochen-vermeulen, line 261: I guess at this point we have no certainty that the depublish plugin is the only plugin of any kind in the workflow. Should the WorkflowUtils, when validating the workflow, not check that a workflow that has a depublication cannot have anything else?

metis-core/metis-core-common/src/main/java/eu/europeana/metis/core/workflow/plugins/AbstractMetisPlugin.java
```diff
@@ -38,6 +38,7 @@
   private String id;
 
   private PluginStatus pluginStatus = PluginStatus.INQUEUE;
+  private DataStatus dataStatus;
```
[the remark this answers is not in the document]
stzanakis: Indeed, we want to mark reindex operations too if there is depublication right? I think in this way we have more control for those as well, but yes indeed we might need to update the reindexing script once we work on it again. But in principle we just need the DELETE status and the others are not really required atm..

metis-core/metis-core-common/src/main/java/eu/europeana/metis/core/workflow/plugins/DepublishPlugin.java
```diff
@@ -1,6 +1,8 @@
 package eu.europeana.metis.core.workflow.plugins;
 
 import eu.europeana.cloud.service.dps.DpsTask;
+import java.util.HashMap;
+import java.util.Map;
 
 /**
  * Depublish Plugin.
```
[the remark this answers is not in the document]
stzanakis: No we said that we keep it as is for now and we'll change it if needed later or just use DePreview if we decide not to change it

metis-core/metis-core-common/src/main/java/eu/europeana/metis/core/rest/DepublishedRecordView.java
```diff
@@ -12,28 +13,17 @@
 
   @JsonSerialize(using = ObjectIdSerializer.class)
   private ObjectId id;
-  private String recordId;
-  private DepublicationStatus depublicationStatus;
+  private final String recordId;
+  private final DepublicationStatus depublicationStatus;
```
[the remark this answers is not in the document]
stzanakis: I thought about it, but then I wasn't sure. What was there before made total sense since it was mapping different values but now that is exactly the same values and it's an enum in this case shouldn't this suffice? 
Or you think we should have separate ones anyway? That could save us from having to change the type in a future change.. I'm in between those, so what's your preference?

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/dao/DepublishedRecordDao.java
```diff
@@ -108,20 +111,20 @@ public int createRecordsToBeDepublished(String datasetId, Set<String> candidateR
...
       record.setDatasetId(datasetId);
       record.setRecordId(recordId);
-      record.setDepublicationState(DepublicationState.NOT_DEPUBLISHED);
+      record.setDepublicationStatus(DepublicationStatus.PENDING_DEPUBLICATION);
       return record;
     }).collect(Collectors.toList());
     ExternalRequestUtil.retryableExternalRequestConnectionReset(() -> {
       morphiaDatastoreProvider.getDatastore().save(objectsToAdd);
-      return null;
+      return Optional.empty();
```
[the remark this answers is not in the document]
stzanakis: Either is perfectly correct since the return value is not used.
I changed it to start following the concept of not returning null values anymore, where possible..

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/dao/DepublishedRecordDao.java
stzanakis: Fixed

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/execution/WorkflowPostProcessor.java
stzanakis: Okay I will remove this. Wasn't sure about it, I thought ecloud will return it as a report and not as part of the monitor call?

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/dao/DepublishedRecordDao.java
stzanakis: I guess we could keep it and extended with the record ids to match for other states?
Shouldn't harm the execution of the query if we send it as a batch which we usually do

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/service/DepublishedRecordService.java
stzanakis: Sure

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/service/WorkflowExecutionFactory.java
stzanakis: Actually there is a check that validates what you just mentioned.

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/dao/DepublishedRecordDao.java
```diff
@@ -147,59 +150,107 @@ long countDepublishedRecordsForDataset(String datasetId) {
...
-            .createQuery(DepublishedRecord.class);
-    query.field(DepublishedRecord.DATASET_ID_FIELD).equal(datasetId);
-    if (StringUtils.isNotBlank(searchQuery)) {
-      query.field(DepublishedRecord.RECORD_ID_FIELD).contains(searchQuery);
-    }
-
-    // Set ordering
-    query.order(sortDirection.createSort(sortField.getDatabaseField()));
+      DepublishedRecordSortField sortField, SortDirection sortDirection,
+      DepublicationStatus depublicationState, String searchQuery) {
```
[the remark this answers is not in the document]
stzanakis: Indeed, I simplified it

metis-core/metis-core-common/src/main/java/eu/europeana/metis/core/rest/DepublishedRecordView.java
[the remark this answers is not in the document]
jochen-vermeulen: I usually favor having a separate one (a view vs a model) and I feel that should apply here too. But I will leave it up to you.

metis-core/metis-core-common/src/main/java/eu/europeana/metis/core/workflow/plugins/AbstractMetisPlugin.java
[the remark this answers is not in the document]
jochen-vermeulen: Perhaps it is indeed best. We can have it, and I think we have a default handling for absent states right?

metis-core/metis-core-common/src/main/java/eu/europeana/metis/core/workflow/plugins/DepublishPlugin.java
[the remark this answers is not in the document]
jochen-vermeulen: OK!

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/dao/DepublishedRecordDao.java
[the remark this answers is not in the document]
jochen-vermeulen: Ah, I see. SonarQube...

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/execution/WorkflowPostProcessor.java
jochen-vermeulen: We need to take care to only mark those that were in fact deleted. I think they were going to implement a service where we can obtain this information. But if we mark one as depublished and it's not, this can cause some issues.

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/dao/DepublishedRecordDao.java
jochen-vermeulen: OK, doesn't matter too much.

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/service/WorkflowExecutionFactory.java
jochen-vermeulen: Oh, I missed it. Ok then.

metis-core/metis-core-common/src/main/java/eu/europeana/metis/core/rest/DepublishedRecordView.java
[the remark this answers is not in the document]
stzanakis: Ok let's do that then.

metis-core/metis-core-common/src/main/java/eu/europeana/metis/core/workflow/plugins/AbstractMetisPlugin.java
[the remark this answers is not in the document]
stzanakis: I believe we do, for historical data.

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/dao/DepublishedRecordDao.java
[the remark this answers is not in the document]
stzanakis: Wasn't actually SonarQube, rather a common issue that I feel I've been facing for long with null returns, so thought perhaps to try this and see how it pans out with always returning optionals in cause of a return value being absent

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/dao/DepublishedRecordDao.java
stzanakis: I thought I'd be generic in both cases, but fair enough I'll remove the parameter

metis-core/metis-core-common/src/main/java/eu/europeana/metis/core/dataset/DepublishedRecord.java
stzanakis: I guess it goes the same for the other fields too so I changed it.

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/dao/DepublishedRecordDao.java
stzanakis: Changing but give it another thought.. Do we accept null values to go in the db?
Same goes for the depublicationDate which is not safe guarded either.

jochen-vermeulen: Then in my view we should have a sanity check for null in the method. This way we force the calling method to do it's homework too (and not just pass null, which is not clear to the reader). 
But I agree with you that it's subjective and that there are arguments for your approach as well.
I leave it up to you.

PR #355, opened 2020-06-26
Feat/met 2702 avoid needless thumbnail generation

jochen-vermeulen:


metis-media-service/src/main/java/eu/europeana/metis/mediaprocessing/RdfDeserializerImpl.java
```diff
@@ -6,50 +6,134 @@
...
- * This object implements RDF deserialization functionality.
+ * This implements RDF deserialization functionality. The code that obtains the individual resources
+ * does not assume that we can convert the record to an EDM internal format. Link checking must also
+ * run on EDM external. We therefore use XPath expressions to obtain the required data.
  */
 class RdfDeserializerImpl implements RdfDeserializer {
 
   private final IUnmarshallingContext context;
 
+  private final XPathExpression getObjectExpression;
```
stzanakis, line 50: We should clearly mention that this class is not thread safe on the documentation because of the XPathExpression..

metis-media-service/src/main/java/eu/europeana/metis/mediaprocessing/RdfDeserializerImpl.java
```diff
@@ -61,8 +145,78 @@
...
+        urls.computeIfAbsent(url, k -> new HashSet<>()).add(type);
+      }
+    }
+    return urls;
+  }
+
+  private Set<String> getUrls(Document document, UrlType type) throws RdfDeserializationException {
+
+    // Determine the right expression to apply.
+    final XPathExpression expression;
```
stzanakis, line 177: I understand that the XPathExpression field of the class are usually all used right?
But in the case of thumbnail only one of those is used..
I think we can gain a lot of performance here if we only create one of those instead of 4.

metis-media-service/src/main/java/eu/europeana/metis/mediaprocessing/RdfDeserializerImpl.java
```diff
@@ -6,50 +6,134 @@
...
+
+    // Create the unmarshalling context
     try {
       context = RdfBindingFactoryProvider.getBindingFactory().createUnmarshallingContext();
     } catch (JiBXException e) {
       throw new RdfConverterException("Problem creating deserializer.", e);
     }
+
+    // Create xpath engine and set namespace context.
+    final XPath xPath = XPathFactory.newInstance().newXPath();
```
stzanakis, line 70: I think we should be using the more robust Saxon library rather than the default..

metis-media-service/src/main/java/eu/europeana/metis/mediaprocessing/RdfDeserializerImpl.java
stzanakis: On the other hand(Please check the other comments on this class as well, because we need to probably choose one or the other option):
This class is instantiated per record right?
If so maybe it's better to make the XPath, XPathExpressions static and synchronize the evaluate method in synchronized blocks where we would probably gain a lot of speed...
What do you think?

metis-media-service/src/main/java/eu/europeana/metis/mediaprocessing/extraction/TextProcessor.java
```diff
@@ -63,9 +64,13 @@ public ResourceExtractionResultImpl copyMetadata(Resource resource, String detec
         resource.getResourceUrl(), resource.getProvidedFileSize()));
   }
 
+  boolean generateThumbnailForPdf(Resource resource, boolean mainThumbnailAvailable) {
+    return !(mainThumbnailAvailable && resource.getUrlTypes().contains(UrlType.IS_SHOWN_BY));
```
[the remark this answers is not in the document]
jochen-vermeulen: You mean the check whether the resource is an isShownBy? Yes, we need that. It can be any kind of link (like hasView) and we only skip thumbnail generation in the case of isShownBy links. 
Or am I missing something?

metis-media-service/src/main/java/eu/europeana/metis/mediaprocessing/RdfDeserializerImpl.java
```diff
@@ -6,50 +6,134 @@
...
+  private RdfResourceEntry getMainThumbnailResourceForMediaExtraction(Document record)
+          throws RdfDeserializationException {
+
+    // Get the entries of the required types.
+    final Map<String, Set<UrlType>> resourceEntries = getResourceEntries(record,
+            Collections.singleton(UrlType.URL_TYPE_FOR_MAIN_THUMBNAIL_RESOURCE));
+
+    // If there is not exactly one, we return null.
+    if (resourceEntries.size() != 1) {
+      return null;
```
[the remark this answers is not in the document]
jochen-vermeulen: OK!

metis-media-service/src/main/java/eu/europeana/metis/mediaprocessing/RdfDeserializerImpl.java
jochen-vermeulen: Perhaps. But that is not necessarily a straight-forward change (by the looks of it). This is code I just copied over, so it's not like I chose this way of processing here. It would be a new change and one that also need to happen in other parts of the code. Perhaps we can create a maintenance ticket for this to separate this from the work here?

metis-media-service/src/main/java/eu/europeana/metis/mediaprocessing/RdfDeserializerImpl.java
jochen-vermeulen: The interface that this class implements (called `RdfDeserializer` mentions this fact in the documentation.

jochen-vermeulen: This class may be instantiated per record, but it doesn't have to be. It was designed to be instantiated per worker thread, which would provide the same speedup. I think that perhaps the best way to make this better is to create a maintenance ticket to do what you suggest (not just here but also in other parts of the code).

jochen-vermeulen: Actually, I could do on-demand evaluation of the xpath expressions that would probably solve some of the issues that you mentioned.

metis-media-service/src/main/java/eu/europeana/metis/mediaprocessing/RdfDeserializerImpl.java
jochen-vermeulen: I'll do some on-demand evaluation that should solve this issue.

metis-media-service/src/main/java/eu/europeana/metis/mediaprocessing/extraction/TextProcessor.java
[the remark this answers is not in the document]
stzanakis: Isn't the resource coming in with the mainThumbnailAvailable flag pre-computed?
I mean, we don't need to check if the resource is an isShownBy here because we would already know if we need to process the resource because it comes with the mainThumbnailAvailable.
So if it's mainThumbnailAvailable is true, we don't really need to find the urlTypes here since we already checked that, right? Or not?

metis-media-service/src/main/java/eu/europeana/metis/mediaprocessing/RdfDeserializerImpl.java
stzanakis: Ok let's create a separate ticket to possibly improve that.
But if you are saying that it's design to be instantiated per worker then we should definitely use Saxon because the worker threads can be on the same JVM and the default implementation has some static fields which can cause an issue..
And/Or synchronize those..
We found an issue like that recently(I forgot the component, I think transformation..)

metis-media-service/src/main/java/eu/europeana/metis/mediaprocessing/RdfDeserializerImpl.java
stzanakis: ok

metis-media-service/src/main/java/eu/europeana/metis/mediaprocessing/RdfDeserializerImpl.java
stzanakis: Same here as above:

"Ok let's create a separate ticket to possibly improve that.
But if you are saying that it's design to be instantiated per worker then we should definitely use Saxon because the worker threads can be on the same JVM and the default implementation has some static fields which can cause an issue..(Saxon implementation does not contain those static fields)
And/Or synchronize those..
We found an issue like that recently(I forgot the component, I think transformation..)"

metis-media-service/src/main/java/eu/europeana/metis/mediaprocessing/extraction/TextProcessor.java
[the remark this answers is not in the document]
jochen-vermeulen: No, the flag tells us whether a main thumbnail is available. But that doesn't mean we don't need to perform any thumbnail extraction anymore at all. For instance, for hasView resources we still want to extract the thumbnails because their thumbnail is not going to be the 'same' as the main thumbnail. We specifically exclude the isShownBy because that thumbnail is expected to be the 'same' as the main thumbnail.

metis-media-service/src/main/java/eu/europeana/metis/mediaprocessing/RdfDeserializerImpl.java
jochen-vermeulen: I made some changes to make things load lazily and in a thread-safe way. Could you create the ticket for what's left to do?

metis-media-service/src/main/java/eu/europeana/metis/mediaprocessing/RdfDeserializerImpl.java
jochen-vermeulen: Same as above:
I made some changes to make things load lazily and in a thread-safe way. Could you create the ticket for what's left to do?

metis-media-service/src/main/java/eu/europeana/metis/mediaprocessing/extraction/TextProcessor.java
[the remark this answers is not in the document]
stzanakis: ok

metis-media-service/src/main/java/eu/europeana/metis/mediaprocessing/RdfDeserializerImpl.java
```diff
@@ -45,41 +44,65 @@
...
+  private final XPathExpressionWrapper getIsShownByExpression = new XPathExpressionWrapper(
+          xPath -> xPath.compile("/rdf:RDF/ore:Aggregation/edm:isShownBy/@rdf:resource"));
+
+  private static class XPathExpressionWrapper extends
+          ThreadSafeWrapper<XPathExpression, RdfDeserializationException> {
+
+    public XPathExpressionWrapper(
+            ThrowingFunction<XPath, XPathExpression, XPathExpressionException> expressionCreator) {
+      super(() -> {
+        final XPath xPath = XPathFactory.newInstance().newXPath();
```
stzanakis, line 64: Ok so this is much better.
I checked the issues we had with transformation and they don't exactly apply here, so we don't really have to use Saxon I suppose.
But in this code there is still a synchronization issue.
The call to XPathFactory.newInstance() is not thread safe and in this case it can be accessed from two threads at the same if those two threads are requesting an expression of two different types. So it works for the same type, but not for two different types.
Do you understand what I mean?

metis-media-service/src/main/java/eu/europeana/metis/mediaprocessing/ThreadSafeWrapper.java
```diff
@@ -0,0 +1,53 @@
...
+  /**
+   * Provides access to the object.
+   *
+   * @param processor The operation that needs to be executed on the object.
+   * @param <O> The output/result type of the operation.
+   * @return The output/result of the operation.
+   * @throws E In case there was a problem.
+   */
+  protected <O> O process(ThrowingFunction<T, O, E> processor) throws E {
+    synchronized (this) {
```
stzanakis, line 34: To solve the issue I mentioned above, we should somehow statically synchronize here perhaps and not on the object.

metis-media-service/src/main/java/eu/europeana/metis/mediaprocessing/RdfDeserializerImpl.java
jochen-vermeulen: I understand. I think the documentation is a bit vague, I assumed that what was meant was that the instance thus created was not thread safe (but the static method was). But it is an easy fix.

metis-media-service/src/main/java/eu/europeana/metis/mediaprocessing/ThreadSafeWrapper.java
jochen-vermeulen: No, I did it somewhere else. :)

metis-media-service/src/main/java/eu/europeana/metis/mediaprocessing/RdfDeserializerImpl.java
stzanakis: Yes it is quite vague..
I hope we are fine thread wise here though now.

PR #357, opened 2020-06-30
Feat/met 2701 upgrade zoho version

stzanakis:


metis-authentication/metis-authentication-rest/src/main/resources/create_tables.sql
```diff
@@ -0,0 +1,28 @@
+CREATE TABLE IF NOT EXISTS metis_users (
+ user_id VARCHAR(100),
+ email VARCHAR(40) PRIMARY KEY,
+ last_name VARCHAR(40),
+ first_name VARCHAR(40),
+ password VARCHAR(255),
+ organization_id VARCHAR(100),
+ organization_name VARCHAR(100),
+ account_role VARCHAR(40) CHECK (account_role IN ('METIS_ADMIN', 'EUROPEANA_DATA_OFFICER', 'PROVIDER_VIEWER') ),
```
jochen-vermeulen, line 9: You can't make this an enum value?

metis-authentication/metis-authentication-service/src/main/java/eu/europeana/metis/authentication/service/AuthenticationService.java
```diff
@@ -45,6 +48,10 @@
   private static final int CREDENTIAL_FIELDS_NUMBER = 2;
   private static final String ACCESS_TOKEN_CHARACTER_BASKET = "0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz";
   private static final int ACCESS_TOKEN_LENGTH = 32;
+  private static final Pattern TOKEN_MATCHING_PATTERN = Pattern
+      .compile("^[" + ACCESS_TOKEN_CHARACTER_BASKET + "]*$");
```
jochen-vermeulen, line 52: Is this not the same as `^[a-zA-Z0-9]*$`?

metis-authentication/metis-authentication-service/src/main/java/eu/europeana/metis/authentication/service/AuthenticationService.java
```diff
@@ -119,21 +126,24 @@ public MetisUser updateUserFromZoho(String email)
     }
 
     psqlMetisUserDao.updateMetisUser(metisUser);
-    return convert(metisUser);
+    return convert(storedMetisUser)
+        .orElseThrow(COULD_NOT_CONVERT_EXCEPTION_SUPPLIER);
```
jochen-vermeulen, line 130: Why not put the `.orElseThrow(...)` in the `convert` method? It is applied every time when that method is called.

metis-authentication/metis-authentication-service/src/main/java/eu/europeana/metis/authentication/service/AuthenticationService.java
```diff
@@ -192,7 +202,7 @@ private boolean isPasswordValid(MetisUserModel metisUser, String passwordToTry)
    */
   public Credentials validateAuthorizationHeaderWithCredentials(String authorization)
       throws GenericMetisException {
-    if (StringUtils.isEmpty(authorization)) {
+    if (StringUtils.isBlank(authorization)) {
```
jochen-vermeulen, line 205: The `String` class in java 11 has a `.isBlank()` method. This way we don't have to use an external library. There are many occurrences in this class. Or perhaps you're allowing for the string to be null?

metis-authentication/metis-authentication-service/src/test/java/eu/europeana/metis/authentication/service/AuthenticationServiceTest.java
```diff
@@ -1,499 +1,504 @@
...
-
-  private static ZCRMRecord getZCRMRecordAccountNoOrganizationName() {
-    final ZCRMRecord zcrmRecordAccount = getZCRMRecordAccount();
-    zcrmRecordAccount.setLookupLabel(null);
-
-    return zcrmRecordAccount;
-  }
-}
\ No newline at end of file
+//package eu.europeana.metis.authentication.service;
```
jochen-vermeulen, line 1: I guess this file should be removed? Or do you intend to fix it at some point? Perhaps it is better to uncomment it and put the `@Ignore` annotation?

metis-common/src/main/java/eu/europeana/metis/zoho/ZohoAccessClient.java
```diff
@@ -53,27 +47,31 @@ public ZohoAccessClient() {
    * tokens from the persistence storage.
    * </p>
    *
-   * @param grantToken the grant token key to be used
+   * @param grantToken the grant token key to be used. Can be null
+   * @param zcrmConfigurations the key value map of zoho configuration
    * @throws ZohoOAuthException if initialization of {@link ZCRMRestClient} failed.
    */
-  public ZohoAccessClient(String grantToken) throws ZohoOAuthException {
+  public ZohoAccessClient(@Nullable String grantToken, HashMap<String, String> zcrmConfigurations)
```
jochen-vermeulen, line 54: SonarLint warning: Use Map<String, String> instead of HashMap to define input type.

metis-common/src/main/java/eu/europeana/metis/zoho/ZohoAccessClient.java
```diff
@@ -53,27 +47,31 @@ public ZohoAccessClient() {
...
-    try {
-      cli.generateAccessToken(grantToken);
-    } catch (ZohoOAuthException e) {
-      LOGGER.warn(
-          "THIS EXCEPTION IS PROBABLY NORMAL. THAT CAN HAPPEN IF THE GRANT TOKEN HAS BEEN SUCCESSFULLY USED ONCE ALREADY!"
-              + "Exception when generating access token. Grant tokens can be used only once, "
-              + "so when the access and refresh tokens are generated, the grant token becomes obsolete on subsequent deployments",
-          e);
+    //Try the grand token if available
+    if (StringUtils.isNotBlank(grantToken)) {
```
jochen-vermeulen, line 63: Also here: can use `String.isBlank()`. I think there are other occurrences in this class. Or are you allowing for it to be null as well?

metis-common/src/main/java/eu/europeana/metis/zoho/ZohoAccessClient.java
```diff
@@ -263,4 +263,10 @@ private String createZohoCriteriaString(Map<String, String> searchCriteria,
                     finalCriteriaOperator, value.trim())).collect(
             Collectors.joining(ZohoConstants.OR))).collect(Collectors.joining(ZohoConstants.OR));
   }
+
+  public static <T extends ZCRMEntity> List<T> castItemsToType(List<? extends ZCRMEntity> zcrmEntities,
```
jochen-vermeulen, line 267: Can't this be private?

metis-common/src/main/java/eu/europeana/metis/zoho/ZohoAccessClient.java
```diff
@@ -85,26 +83,26 @@ public ZohoAccessClient(String grantToken) throws ZohoOAuthException {
...
-
   /**
    * Get a contact by using an organization email.
    *
    * @param organizationName the organization name to search for
    * @return the organization corresponding to the organization name
    * @throws BadContentException if an exception occurred during searching
    */
-  public ZCRMRecord getZcrmRecordOrganizationByName(String organizationName)
+  public Optional<ZCRMRecord> getZcrmRecordOrganizationByName(String organizationName)
```
jochen-vermeulen, line 105: According to my IDE this method, nor the following two, are in use. Is this true?

metis-authentication/metis-authentication-service/src/main/java/eu/europeana/metis/authentication/service/AuthenticationService.java
stzanakis: Indeed, I guess I did an automatic conversion from the string comparison.

metis-authentication/metis-authentication-service/src/main/java/eu/europeana/metis/authentication/service/AuthenticationService.java
stzanakis: Sure

metis-common/src/main/java/eu/europeana/metis/zoho/ZohoAccessClient.java
stzanakis: Yes and no.
True for our code base.
False, it's used in some script that the api team is using..

metis-common/src/main/java/eu/europeana/metis/zoho/ZohoAccessClient.java
stzanakis: Yes it can

metis-authentication/metis-authentication-service/src/main/java/eu/europeana/metis/authentication/service/AuthenticationService.java
stzanakis: I'm safe guarding from null too.
I could do the same with Objects.isNull() && String.isBlank().
What do you prefeer?

metis-authentication/metis-authentication-service/src/test/java/eu/europeana/metis/authentication/service/AuthenticationServiceTest.java
stzanakis: Darn I knew I will forget about it :D
No totally not, need to fix that..

metis-common/src/main/java/eu/europeana/metis/zoho/ZohoAccessClient.java
stzanakis: Yes I had the same. The issue comes from 
ZCRMRestClient.initialize() which requires a HashMap apparently.
I could change the parameter to Map, but then I have to cast it to HashMap which can be unsafe.
What do you think?

metis-common/src/main/java/eu/europeana/metis/zoho/ZohoAccessClient.java
stzanakis: Same as the other remark above

metis-authentication/metis-authentication-rest/src/main/resources/create_tables.sql
stzanakis: I could. There is an issue with ENUM types though..
I cannot say IF NOT EXISTS. I can only DROP it first, and if I do that it complains that it's used in other tables..
Do you know perhaps a better way to do this?

stzanakis: Or maybe I'll remove the IF NOT EXISTS on everything?
And it will always fails if it is executed on a db that already has any of the tables?

metis-common/src/main/java/eu/europeana/metis/zoho/ZohoAccessClient.java
stzanakis: Yes I am aware of this. The reason is that the ZCRMRestClient.initialize(zcrmConfigurations); accepts a HashMap, so I thought the class here should also be a HashMap.
The alternative is to provide a Map here and cast it before sending it to the inner class..
Not sure what is better, somehow the user of this class should know that the Map has to be of type HashMap. Perhaps though documentation and we check the type of the Map inside here with an exception if it's not the type we want?
What do you think?

metis-common/src/main/java/eu/europeana/metis/zoho/ZohoAccessClient.java
jochen-vermeulen: Ah, ok. I didn't know that.

metis-common/src/main/java/eu/europeana/metis/zoho/ZohoAccessClient.java
jochen-vermeulen: OK, so we specifically need it to be a HashMap here? Then I see two options that I don't dislike: 
1. Make it a generic map and then use `new HashMap<>(zcrmConfigurations)` that we then send to zoho. That way we accept generic input and still satisfy Zoho.
2. Put an exception for the sonar lint rule in the code with a piece of comments explaining this.
I think I slightly favor the first one. What do you think?

metis-authentication/metis-authentication-service/src/main/java/eu/europeana/metis/authentication/service/AuthenticationService.java
jochen-vermeulen: This is a good question. I don't know what I prefer. Is it the only reason to include the library in this module? If it isn't, perhaps for the time being we should just keep using `StringUtils`.

metis-authentication/metis-authentication-service/src/main/java/eu/europeana/metis/authentication/service/AuthenticationService.java
jochen-vermeulen: Ah, i guess I didn't see that the BASKET is used outside a regex setting. THen perhaps it is better to leave it as it is?

metis-authentication/metis-authentication-rest/src/main/resources/create_tables.sql
jochen-vermeulen: Maybe not overcomplicate things... In fact, perhaps it is best to just remove the check for values? It is a text field, who cares what values are in it? The java code will handle it.

stzanakis: Ok let's remove the check completely.. I've added a comment also to give a bit of info.

metis-authentication/metis-authentication-service/src/main/java/eu/europeana/metis/authentication/service/AuthenticationService.java
stzanakis: I couldn't find another way to use a regex to generate the tokens, so leaving it as is.

metis-authentication/metis-authentication-service/src/main/java/eu/europeana/metis/authentication/service/AuthenticationService.java
stzanakis: I understand the point but I'm leaving it as is, I think the commons.lang3 is very common library and we use it everywhere. Even if we use one function of it, we use it so often and we know exactly what it does.. Creating/duplicating the functionality ,takes us away from that security, besides the library is 500kb.

I think this can be of broader discussion.. We could do what you are suggesting, if we do this everywhere, which could make sense. Leaving it out of this pr

metis-common/src/main/java/eu/europeana/metis/zoho/ZohoAccessClient.java
stzanakis: Yeah number 1 seems better

PR #359, opened 2020-07-02
MET-2828 Support for removal of depublish record ids

stzanakis:


metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/service/DepublishRecordIdService.java
```diff
@@ -119,13 +120,15 @@ public DepublishRecordIdService(Authorizer authorizer, OrchestratorService orche
...
+   * Deletes a list of record ids from the database. Only record ids that are in a {@link
+   * eu.europeana.metis.core.dataset.DepublishRecordId.DepublicationStatus#PENDING_DEPUBLICATION}
+   * state will be removed.
    *
    * @param metisUser The user performing this operation.
    * @param datasetId The ID of the dataset to which the depublished records belong.
    * @param recordIdsInSeparateLines The string containing the record IDs in separate lines.
-   * @return How many of the passed records were in fact added. This counter is not thread-safe: if
-   * multiple threads try to add the same records, their combined counters may overrepresent the
+   * @return How many of the passed records were in fact removed. This counter is not thread-safe:
```
jochen-vermeulen, line 130: I think something went wrong with the documentation here. The method that this documentation is for is called `add...`

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/dao/DepublishRecordIdDao.java
```diff
@@ -131,6 +131,37 @@ public int createRecordIdsToBeDepublished(String datasetId, Set<String> candidat
...
+    final Query<DepublishRecordId> query = morphiaDatastoreProvider.getDatastore()
+        .createQuery(DepublishRecordId.class);
+    query.field(DepublishRecordId.DATASET_ID_FIELD).equal(datasetId);
+    query.field(DepublishRecordId.RECORD_ID_FIELD).in(recordIds);
+    query.field(DepublishRecordId.DEPUBLICATION_STATUS_FIELD)
+        .equal(DepublicationStatus.PENDING_DEPUBLICATION);
+
+    return ExternalRequestUtil.retryableExternalRequestConnectionReset(() -> {
+      return morphiaDatastoreProvider.getDatastore().delete(query).getN();
+    });
```
jochen-vermeulen, line 162: This lambda doesn't need to have curly braces and a return statement.

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/dao/DepublishRecordIdDao.java
```diff
@@ -131,6 +131,37 @@ public int createRecordIdsToBeDepublished(String datasetId, Set<String> candidat
...
+      throw new BadContentException(
+          "Can't remove these records: this would violate the maximum number of records per dataset.");
+    }
+
+    final Query<DepublishRecordId> query = morphiaDatastoreProvider.getDatastore()
+        .createQuery(DepublishRecordId.class);
+    query.field(DepublishRecordId.DATASET_ID_FIELD).equal(datasetId);
+    query.field(DepublishRecordId.RECORD_ID_FIELD).in(recordIds);
+    query.field(DepublishRecordId.DEPUBLICATION_STATUS_FIELD)
+        .equal(DepublicationStatus.PENDING_DEPUBLICATION);
```
jochen-vermeulen, line 158: Should we instead not throw an error if any record is found to be undeletable because it has the wrong status? What do you think? Should we silently skip it?

stzanakis: I think we should silently skip it yes.
The default behavior of a delete operation is to be silent in general.
And also we expect the incoming list to be the correct one, since we are in control of it.

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/dao/DepublishRecordIdDao.java
stzanakis: True!

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/service/DepublishRecordIdService.java
stzanakis: Fixed

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/dao/DepublishRecordIdDao.java
```diff
@@ -157,9 +157,8 @@ public Integer deletePendingRecordIds(String datasetId, Set<String> recordIds)
     query.field(DepublishRecordId.DEPUBLICATION_STATUS_FIELD)
         .equal(DepublicationStatus.PENDING_DEPUBLICATION);
 
-    return ExternalRequestUtil.retryableExternalRequestConnectionReset(() -> {
-      return morphiaDatastoreProvider.getDatastore().delete(query).getN();
-    });
+    return ExternalRequestUtil.retryableExternalRequestConnectionReset(
+        morphiaDatastoreProvider.getDatastore().delete(query)::getN);
```
jochen-vermeulen, line 161: Ah, I don't think this is right! This evaluates the delete query immediately, upon creation of the lambda! Not when it is called. I think you should use the ()-> notation here.

stzanakis: Ooh yes! Good catch!

PR #360, opened 2020-07-10
Bug/met 2724 fix depublish computation of predecessor

stzanakis:


metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/dao/WorkflowUtils.java
```diff
@@ -323,42 +326,48 @@ public static AbstractExecutablePlugin computePredecessorPlugin(ExecutablePlugin
...
 
-    // If the plugin type does not need a predecessor (even the enforced one) we are done.
+    final PluginWithExecutionId<ExecutablePlugin> predecessorPlugin;
     final Set<ExecutablePluginType> defaultPredecessorTypes = getPredecessorTypes(pluginType);
-    if (defaultPredecessorTypes.isEmpty()) {
-      return null;
+    // If the plugin type does not need a predecessor (even the enforced one) we are done.
+    // We also return if a DEPUBLISH Operation is requested
+    if (defaultPredecessorTypes.isEmpty() || (defaultPredecessorTypes
+        .contains(ExecutablePluginType.PUBLISH) && pluginType == ExecutablePluginType.DEPUBLISH)) {
```
jochen-vermeulen, line 334: Shouldn't we check that a valid publish exists? We could have someone triggering a depublish without there being a publish, right? We could execute the check in line 350-354 for that case as well? What do you think?

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/dao/WorkflowUtils.java
```diff
@@ -262,7 +262,9 @@ public static AbstractExecutablePlugin computePredecessorPlugin(ExecutablePlugin
 
     // If the plugin type does not need a predecessor we are done.
     final Set<ExecutablePluginType> predecessorTypes = getPredecessorTypes(pluginType);
-    if (predecessorTypes.isEmpty()) {
+    if (predecessorTypes.isEmpty()
+        || pluginType == ExecutablePluginType.DEPUBLISH && predecessorTypes.size() == 1
+        && predecessorTypes.contains(ExecutablePluginType.PUBLISH)) {
```
jochen-vermeulen, line 267: Is the precedence of operations clear here? Is it && before || ? Or is it perhaps better to add parentheses?

stzanakis: The result wouldn't change, but I added them for clarity

PR #365, opened 2020-07-17
Feat/met 278 eliminate redis in enrichment

stzanakis:


metis-enrichment/metis-enrichment-common/src/main/java/eu/europeana/enrichment/utils/EntityTypeUtils.java
```diff
@@ -0,0 +1,83 @@
...
+    entityInfos.add(new EntityInfo<>(EntityType.PLACE, PlaceTermList.class, PlaceImpl.class));
+    entityInfos
+        .add(new EntityInfo<>(EntityType.TIMESPAN, TimespanTermList.class, TimespanImpl.class));
+    entityInfos.add(new EntityInfo<>(EntityType.ORGANIZATION, OrganizationTermList.class,
+        OrganizationImpl.class));
+    return entityInfos;
+  }
+
+  public static EntityInfo<?, ?> getEntityMongoTermListClass(EntityType entityType) {
+    return ENTITY_INFOS.stream().filter((entityInfo) -> entityType == entityInfo.entityType)
```
jochen-vermeulen, line 39: IDE: Remove the parentheses around the entityInfo lambda parameter.

metis-enrichment/metis-enrichment-common/src/main/java/eu/europeana/enrichment/utils/EntityTypeUtils.java
```diff
@@ -0,0 +1,83 @@
...
+import eu.europeana.enrichment.api.internal.PlaceTermList;
+import eu.europeana.enrichment.api.internal.TimespanTermList;
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * @author Simon Tzanakis
+ * @since 2020-07-17
+ */
+public class EntityTypeUtils {
```
jochen-vermeulen, line 22: IDE: Add a private constructor to hide the implicit public one

metis-enrichment/metis-enrichment-common/src/main/java/eu/europeana/enrichment/utils/EntityTypeUtils.java
```diff
@@ -0,0 +1,83 @@
...
+
+  }
+
+  public static EntityType getEntityTypeFromClassImpl(String entityClassImpl) {
+    return ENTITY_INFOS.stream()
+        .filter(entityInfo -> entityInfo.getEntityClassImpl().getSimpleName()
+            .equalsIgnoreCase(entityClassImpl)).findFirst().map(EntityInfo::getEntityType)
+        .orElse(null);
+  }
+  public static EntityInfo<?, ?> getEntityTypeFromClassImplCLASS(String entityClassImpl) {
```
jochen-vermeulen, line 50: According to my IDE, this method is never used. Also, the name is not very good.

metis-enrichment/metis-enrichment-rest/src/main/java/eu/europeana/enrichment/rest/EnrichmentController.java
```diff
@@ -33,115 +30,98 @@
...
+   * @param uri The URI to check for match
    * @return the structured result of the enrichment
-   * @throws EnrichmentException if an exception occurred during enrichment
    */
-  @GetMapping(value = RestEndpoints.ENRICHMENT_BYURI, produces = {MediaType.APPLICATION_JSON_VALUE,
+  @GetMapping(value = RestEndpoints.ENRICH_CODEURI_OR_OWLSAMEAS, produces = {
+      MediaType.APPLICATION_JSON_VALUE,
       MediaType.APPLICATION_XML_VALUE})
-  @ApiOperation(value = "Retrieve an entity by rdf:about or owl:sameAs/skos:exactMatch", response = EnrichmentBase.class)
+  @ApiOperation(value = "Get an enrichment by URI, might match owl:sameAs", response = EnrichmentBase.class)
```
jochen-vermeulen, line 54: Why change these comments? I thought the original was clearer. It's still true, right? Also below.

metis-enrichment/metis-enrichment-rest/src/main/java/eu/europeana/enrichment/rest/config/Application.java
```diff
@@ -99,24 +84,18 @@ public void addResourceHandlers(ResourceHandlerRegistry registry) {
...
+  EnrichmentDao getEnrichmentDao() {
     final MongoProperties<IllegalArgumentException> mongoProperties = new MongoProperties<>(
         IllegalArgumentException::new);
-    mongoProperties.setMongoHosts(new String[]{enrichmentMongoHost}, new int[]{enrichmentMongoPort});
-    final MongoClient mongoClient = new MongoClientProvider<>(mongoProperties).createMongoClient();
-    final EntityDao entityDao = new EntityDao(mongoClient, enrichmentMongoDatabase);
-    return new RedisInternalEnricher(entityDao, redisProvider);
+    mongoProperties
+        .setMongoHosts(new String[]{enrichmentMongoHost}, new int[]{enrichmentMongoPort});
+    mongoClient = new MongoClientProvider<>(mongoProperties).createMongoClient();
```
jochen-vermeulen, line 97: I guess we know that this is only called once? Maybe it is more clear to give the mongo client its own bean or a if not null check?

metis-enrichment/metis-enrichment-service/src/main/java/eu/europeana/enrichment/service/Converter.java
```diff
@@ -1,74 +1,62 @@
...
-   * @param wrapperList the {@link List<EntityWrapper>} to convert
-   * @return the result of the convertion
-   * @throws IOException if conversion fails
-   */
-  public EnrichmentResultList convert(List<EntityWrapper> wrapperList) throws IOException {
-    EnrichmentResultList list = new EnrichmentResultList();
-
-    for (EntityWrapper wrapper : wrapperList) {
-      list.getEnrichmentBaseWrapperList().add(new EnrichmentBaseWrapper(wrapper.getOriginalField(), convert(wrapper)));
+public abstract class Converter {
```
jochen-vermeulen, line 29: My IDE is giving me some warnings here. Perhaps:
- Make final instead of abstract (because sonarlint complains that an abstract class does not have instances).
- Add private no-args constructor

metis-enrichment/metis-enrichment-common/src/main/java/eu/europeana/enrichment/utils/EntityTypeUtils.java
stzanakis: Done

metis-enrichment/metis-enrichment-common/src/main/java/eu/europeana/enrichment/utils/EntityTypeUtils.java
stzanakis: Done

metis-enrichment/metis-enrichment-common/src/main/java/eu/europeana/enrichment/utils/EntityTypeUtils.java
stzanakis: That IDE... :D 
Weird leftovers, removed!

metis-enrichment/metis-enrichment-rest/src/main/java/eu/europeana/enrichment/rest/EnrichmentController.java
stzanakis: What was there makes sense generally but if you think about it maybe not exactly..
The db has the values:
- codeUri
- owlSameAs

I do not know how exactly those fields were filled in exactly.
And whoever is searching the db is using a uri(we don't know from what field that comes from), with the scope of getting a match from the codeUri or owlSameAs field in the db..
What do you think the comments should be in this case?

metis-enrichment/metis-enrichment-rest/src/main/java/eu/europeana/enrichment/rest/config/Application.java
stzanakis: We could do that, I think spring will close it automatically too.

metis-enrichment/metis-enrichment-service/src/main/java/eu/europeana/enrichment/service/Converter.java
stzanakis: Adding IDE suggestion

metis-enrichment/metis-enrichment-rest/src/main/java/eu/europeana/enrichment/rest/EnrichmentController.java
jochen-vermeulen: What threw me off is that you specifically name owl:sameAs, but no longer rdf:about. Seems oddly matched this way.

Admittedly, this is not the most vital comment in the world, so I'm fine to leave it as it is (i.e. your new phrasing). But if you feel like changing it, perhaps we can have the best of both worlds, and just put the specific field names it matches in parentheses by way of explanation?

stzanakis: Updated the documentation

PR #370, opened 2020-08-19
Feat/met 2725 process depublication result

jochen-vermeulen:


metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/dao/DepublishRecordIdDao.java
```diff
@@ -327,10 +336,34 @@ public void markRecordIdsWithDepublicationStatus(String datasetId,
           DepublicationStatus.DEPUBLISHED.name()));
     }
 
+    // If we have a specific record list, make sure that missing records are added.
+    final Set<String> recordIdsToUpdate; // null if and only if we need to update all records
+    if (recordIds == null || recordIds.isEmpty()) {
```
stzanakis, line 341: CollectionUtils.isEmpty()?

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/execution/WorkflowPostProcessor.java
```diff
@@ -50,49 +66,91 @@ public WorkflowPostProcessor(DepublishRecordIdDao depublishRecordIdDao,
...
+
+    // Retrieve the successfully depublished records.
+    final long externalTaskId = Long.parseLong(depublishPlugin.getExternalTaskId());
+    final List<SubTaskInfo> subTasks = new ArrayList<>();
+    while (true) {
+      final List<SubTaskInfo> subTasksBatch = dpsClient.getDetailedTaskReportBetweenChunks(
+              depublishPlugin.getTopologyName(), externalTaskId, subTasks.size(),
+              subTasks.size() + ECLOUD_REQUEST_BATCH_SIZE);
+      subTasks.addAll(subTasksBatch);
+      if (subTasksBatch.size() < ECLOUD_REQUEST_BATCH_SIZE) {
```
stzanakis, line 134: Integrate this into the loop I would say

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/execution/WorkflowPostProcessor.java
```diff
@@ -50,49 +66,91 @@ public WorkflowPostProcessor(DepublishRecordIdDao depublishRecordIdDao,
...
+  }
+
+  private void depublishDatasetPostProcess(String datasetId){
+
+    // Set all depublished records back to PENDING.
+    depublishRecordIdDao.markRecordIdsWithDepublicationStatus(datasetId, null,
+            DepublicationStatus.PENDING_DEPUBLICATION, null);
+
+    // Find latest PUBLISH Type Plugin and set dataStatus to DELETED.
+    final PluginWithExecutionId<MetisPlugin> latestSuccessfulPlugin = workflowExecutionDao
```
stzanakis, line 103: IDE complaints about generic type

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/execution/WorkflowPostProcessor.java
```diff
@@ -50,49 +66,91 @@ public WorkflowPostProcessor(DepublishRecordIdDao depublishRecordIdDao,
...
+            .getLatestSuccessfulPlugin(datasetId, OrchestratorService.PUBLISH_TYPES);
+    if (Objects.nonNull(latestSuccessfulPlugin) && Objects
+            .nonNull(latestSuccessfulPlugin.getPlugin())) {
+      final WorkflowExecution workflowExecutionToUpdate = workflowExecutionDao
+              .getById(latestSuccessfulPlugin.getExecutionId());
+      final Optional<AbstractMetisPlugin> metisPluginWithType = workflowExecutionToUpdate
```
stzanakis, line 109: IDE complaints about generic type

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/dao/DepublishRecordIdDao.java
```diff
@@ -309,14 +315,17 @@ public long countSuccessfullyDepublishedRecordIdsForDataset(String datasetId) {
...
    * @param datasetId the dataset for which to do this. Cannot be null
+   * @param recordIds the records for which to set this. Can be null or empty, in which case the
+   * operation will be performed on all records. If it is not empty, a new record will be created if
+   * a record with the given record ID is not already present.
    * @param depublicationStatus the depublication status. Cannot be null
    * @param depublicationDate the depublication date. Can be null only if depublicationStatus is
    * {@link DepublicationStatus#PENDING_DEPUBLICATION}
    */
-  public void markRecordIdsWithDepublicationStatus(String datasetId,
+  public void markRecordIdsWithDepublicationStatus(String datasetId, Set<String> recordIds,
```
stzanakis, line 325: Why are we adding records here, I feel we shouldn't be doing that..?

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/execution/WorkflowPostProcessor.java
```diff
@@ -50,49 +66,91 @@ public WorkflowPostProcessor(DepublishRecordIdDao depublishRecordIdDao,
...
+    }
+
+    // Mark the records as DEPUBLISHED.
+    final Map<String, Set<String>> successfulRecords = subTasks.stream()
+            .filter(subTask -> subTask.getRecordState() == RecordState.SUCCESS)
+            .map(SubTaskInfo::getResource).map(DepublishRecordIdUtils::decomposeFullRecordId)
+            .collect(Collectors.groupingBy(Pair::getLeft,
+                    Collectors.mapping(Pair::getRight, Collectors.toSet())));
+    successfulRecords.forEach((dataset, records) ->
+            depublishRecordIdDao.markRecordIdsWithDepublicationStatus(dataset, records,
```
stzanakis, line 146: I also commented on merkRecordIdsWithDepublicationStatus, why are we adding those records if missing?
I feel we should just skip in case that happens.. That shouldn't be happening in a normal scenario and if it does it will be confusing on the frontend side

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/dao/DepublishRecordIdDao.java
jochen-vermeulen: Sure.

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/execution/WorkflowPostProcessor.java
jochen-vermeulen: Ok. Like so?

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/execution/WorkflowPostProcessor.java
jochen-vermeulen: I think this is outside scope... it's just what the called method returns. I think we need a separate tech debt ticket to solve those issues.

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/execution/WorkflowPostProcessor.java
jochen-vermeulen: Same as above.

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/dao/DepublishRecordIdDao.java
jochen-vermeulen: Imagine the following situation. 
1. Trigger depublication of records.
2. Remove those records from the list while the depublication is running
3. The depublication finishes, but it no longer has the record to set the DEPUBLISHED status.
I think we should reintroduce the record and set the status so that the user knows what happened.
(also, this is what the specs say - even though I wrote them myself :) )

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/execution/WorkflowPostProcessor.java
jochen-vermeulen: See my other comment.

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/dao/DepublishRecordIdDao.java
stzanakis: Perhaps we should not allow editing the list when an execution is running, instead?
Same thing we do with updating a workflow..

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/execution/WorkflowPostProcessor.java
stzanakis: Okay no problem

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/execution/WorkflowPostProcessor.java
stzanakis: Yeap!

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/dao/DepublishRecordIdDao.java
jochen-vermeulen: Actually, it is possible to edit a workflow while an execution is in progress, isn't it?
But perhaps this is the solution in this case.
However, perhaps it would be best to do this in a separate ticket? I feel reluctant to change the scope of this ticket now, and whether we (I) change this now, or later, doesn't really matter. Also that gives us (me) some time to think about this. What do you think?

stzanakis: As discussed we'll handle this as a separate ticket

PR #372, opened 2020-09-03
MET-2952 Fix saving functionality of caching entities

stzanakis:


metis-dereference/metis-dereference-common/src/main/java/eu/europeana/metis/dereference/ProcessedEntity.java
```diff
@@ -15,7 +18,8 @@
 public class ProcessedEntity {
 
   @Id
-  private String id;
+  @JsonSerialize(using = ObjectIdSerializer.class)
```
JoanaCMS, line 21: The input file is json? I believed it was XML? Or am I confusing with something else?

metis-dereference/metis-dereference-common/src/main/java/eu/europeana/metis/dereference/Vocabulary.java
```diff
@@ -24,7 +27,8 @@
   private static final long serialVersionUID = 2946293185967000824L;
 
   @Id
-  private String id;
+  @JsonSerialize(using = ObjectIdSerializer.class)
```
JoanaCMS, line 30: Same question as before

metis-dereference/metis-dereference-service/src/main/java/eu/europeana/metis/dereference/service/MongoDereferenceService.java
```diff
@@ -180,18 +180,18 @@ private void extractBroaderResources(EnrichmentBase resource, Set<String> destin
...
     if (entityString == null || vocabulary == null) {
+      final VocabularyCandidates candidates =
+          VocabularyCandidates.findVocabulariesForUrl(resourceId, vocabularyDao::getByUriSearch);
       final Pair<String, Vocabulary> transformedEntity =
           retrieveTransformedEntity(resourceId, candidates);
       if (transformedEntity != null) {
         entityString = transformedEntity.getLeft();
         vocabulary = transformedEntity.getRight();
-        final ProcessedEntity entityToCache = new ProcessedEntity();
+        ProcessedEntity entityToCache = (cachedEntity == null) ? new ProcessedEntity() : cachedEntity;
```
JoanaCMS, line 191: Isn't this check done before? On line 176 if not mistaken. Why this if statement?

metis-dereference/metis-dereference-service/src/main/java/eu/europeana/metis/dereference/service/MongoDereferenceService.java
```diff
@@ -180,18 +180,18 @@ private void extractBroaderResources(EnrichmentBase resource, Set<String> destin
...
         entityToCache.setXml(entityString);
         entityToCache.setResourceId(resourceId);
-        entityToCache.setVocabularyId(vocabulary == null ? null : vocabulary.getId());
+        entityToCache.setVocabularyId((vocabulary == null) ? null : vocabulary.getId().toString());
```
JoanaCMS, line 194: I thought the id of Vocabulary was type ObjectId, it accepts String?

metis-dereference/metis-dereference-service/src/main/java/eu/europeana/metis/dereference/service/MongoDereferenceService.java
stzanakis: Do you mean why the ternary operator is applied here?
Well yeah cachedEntity is checked on 176 but it can still be null here in which case we want a new ProcessedEntity object otherwise we want the non null cachedEntity.
That is done in that way to that the processedEntityDao.save(entityToCache); will know if it's an update(we are saving an object that we already have which has an identifier internally that is already known) or an insertion.

metis-dereference/metis-dereference-service/src/main/java/eu/europeana/metis/dereference/service/MongoDereferenceService.java
stzanakis: The id of the vocabulary is ObjectId indeed but we save it as string in the ProcessedEntity..
A better way to do this is to use mongo references see here:
https://docs.mongodb.com/manual/reference/database-references/

But this was out of scope of this ticket, so I've kept the previous functionality here.

metis-dereference/metis-dereference-common/src/main/java/eu/europeana/metis/dereference/ProcessedEntity.java
stzanakis: The Vocabulary class is displayed through swagger, so for that it's definitely required.
For the ProcessedEntity we don't really display it anywhere so it could have been omitted but I added it for consistency..

metis-dereference/metis-dereference-common/src/main/java/eu/europeana/metis/dereference/Vocabulary.java
stzanakis: Same answer

PR #373, opened 2020-09-04
MET-657 Clean Normalize URI

JoanaCMS:


metis-normalization/src/main/java/eu/europeana/normalization/normalizers/CleanIRIViolationsNormalizer.java
```diff
@@ -0,0 +1,94 @@
...
+  }
+
+  @Override
+  public List<NormalizedValueWithConfidence> normalizeValue(String value) {
+    if(value == null || value.isEmpty()){
+      return Collections.emptyList();
+    }
+
+    iri = iriFactory.create(value);
+    String normalizedValue = "";
```
stzanakis, line 75: This can be final and initialized in the try block

metis-normalization/src/main/java/eu/europeana/normalization/normalizers/CleanIRIViolationsNormalizer.java
```diff
@@ -0,0 +1,94 @@
...
+
+  @Override
+  public RecordNormalizeAction getAsRecordNormalizer() {
+    return new ValueNormalizeActionWrapper(this, RESOURCE_IS_SHOWN_BY_QUERY, RESOURCE_HAS_VIEW_QUERY,
+        RESOURCE_IS_SHOWN_AT_QUERY, RESOURCE_OBJECT_QUERY, WEB_RESOURCE_ABOUT_QUERY);
+  }
+
+  @Override
+  public List<NormalizedValueWithConfidence> normalizeValue(String value) {
+    if(value == null || value.isEmpty()){
```
stzanakis, line 70: Better use StringUtils.isBlank().
In general this is what we use, for such operations.

metis-normalization/src/main/java/eu/europeana/normalization/normalizers/CleanIRIViolationsNormalizer.java
```diff
@@ -0,0 +1,94 @@
...
+      normalizedValue = iri.toURI().toString();
+    } catch (URISyntaxException e) {
+      LOG.debug("There was some trouble normalizing the value for IRI Violation");
+      return Collections.emptyList();
+    }
+
+    return Collections.singletonList(new NormalizedValueWithConfidence(normalizedValue, 1));
+  }
+
+  Iterator<Violation> getViolations(){
```
stzanakis, line 87: Try to use one return value instead of two.
Also provide javadoc? Not clear what taht iri.violations(false) returns..

metis-normalization/src/main/java/eu/europeana/normalization/normalizers/CleanIRIViolationsNormalizer.java
```diff
@@ -0,0 +1,94 @@
...
+  @Override
+  public RecordNormalizeAction getAsRecordNormalizer() {
+    return new ValueNormalizeActionWrapper(this, RESOURCE_IS_SHOWN_BY_QUERY, RESOURCE_HAS_VIEW_QUERY,
+        RESOURCE_IS_SHOWN_AT_QUERY, RESOURCE_OBJECT_QUERY, WEB_RESOURCE_ABOUT_QUERY);
+  }
+
+  @Override
+  public List<NormalizedValueWithConfidence> normalizeValue(String value) {
+    if(value == null || value.isEmpty()){
+      return Collections.emptyList();
```
stzanakis, line 71: Try to reduce the return statements in this method

metis-normalization/src/main/java/eu/europeana/normalization/normalizers/CleanIRIViolationsNormalizer.java
```diff
@@ -0,0 +1,94 @@
...
+      "/%s/%s/@%s", XpathQuery.RDF_TAG, EDM_WEB_RESOURCE, RDF_ABOUT);
+
+  private final IRIFactory iriFactory;
+  private IRI iri = null;
+
+  /**
+   * Creates a new instance of this class.
+   */
+  public CleanIRIViolationsNormalizer() {
+    iriFactory = IRIFactory.iriImplementation();
```
stzanakis, line 59: Is it possible to initialize this on the declaration and avoid typing the default constructor here?

metis-normalization/src/test/java/eu/europeana/normalization/normalizers/CleanIRIViolationsNormalizerTest.java
```diff
@@ -0,0 +1,89 @@
...
+import org.apache.jena.iri.Violation;
+import org.junit.jupiter.api.Test;
+
+import java.util.*;
+import java.util.stream.StreamSupport;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+
+public class CleanIRIViolationsNormalizerTest {
```
stzanakis, line 12: All public modifiers can be omitted

metis-normalization/src/test/java/eu/europeana/normalization/normalizers/CleanIRIViolationsNormalizerTest.java
```diff
@@ -0,0 +1,89 @@
...
+  private static final String DOUBLE_WHITESPACE_VIOLATION = "http://example.com/query?q=random  word";
+  private static final String CONTROL_CHARACTER_VIOLATION = "http://example.com/query?q=random\u0085word";
+  private static final String COMPATIBILITY_CHARACTER_VIOLATION = "http://example.com/query?q=randomໜword";
+
+
+  @Test
+  public void whitespaceViolationCleanupTest(){
+    List<NormalizedValueWithConfidence> result = normalizer.normalizeValue(WHITESPACE_VIOLATION);
+    Iterator<Violation> violations = normalizer.getViolations();
+    assertTrue(StreamSupport.stream(Spliterators.spliteratorUnknownSize(violations, Spliterator.SORTED), false)
```
stzanakis, line 28: How do you know that the list is SORTED? Not that it is important in this case but I would probably use another characteristic..
Also all the tests are having one violation, since in the test this is what is expected you can directly access the first violation by violations.next().codeName(), in that way you win some processing time here, what do you think?

stzanakis (changes requested):

metis-normalization/src/main/java/eu/europeana/normalization/normalizers/CleanIRIViolationsNormalizer.java
JoanaCMS: Will see what I can do. But is this some kind of protocol? Or is there a different reason to reduce to one return statement?

metis-normalization/src/main/java/eu/europeana/normalization/normalizers/CleanIRIViolationsNormalizer.java
JoanaCMS: I can do this change. What is the difference between StringUtils.isBlank() and string.isEmpty()?

metis-normalization/src/main/java/eu/europeana/normalization/normalizers/CleanIRIViolationsNormalizer.java
JoanaCMS: Just took care of the javadoc. What do you mean by returning one value instead of 2? Only one return statement?

metis-normalization/src/test/java/eu/europeana/normalization/normalizers/CleanIRIViolationsNormalizerTest.java
JoanaCMS: I actually did initially violations.next().codeName(), but for some of the test cases, there are more than 1 violation besides the one we want to test, so I decided to change to what it shows now.

metis-normalization/src/main/java/eu/europeana/normalization/normalizers/CleanIRIViolationsNormalizer.java
```diff
@@ -0,0 +1,94 @@
...
+  private static final Namespace.Element RDF_RESOURCE = Namespace.RDF.getElement("resource");
+  private static final Namespace.Element RDF_ABOUT = Namespace.RDF.getElement("about");
+  private static final Namespace.Element ORE_AGGREGATION = Namespace.ORE.getElement("Aggregation");
+  private static final Namespace.Element EDM_IS_SHOWN_BY = Namespace.EDM.getElement("isShownBy");
+  private static final Namespace.Element EDM_IS_SHOWN_AT = Namespace.EDM.getElement("isShownAt");
+  private static final Namespace.Element EDM_OBJECT = Namespace.EDM.getElement("object");
+  private static final Namespace.Element EDM_HAS_VIEW = Namespace.EDM.getElement("hasView");
+  private static final Namespace.Element EDM_WEB_RESOURCE = Namespace.EDM.getElement("WebResource");
+
+  private static final XpathQuery RESOURCE_IS_SHOWN_BY_QUERY = new XpathQuery(
```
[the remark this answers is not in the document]
JoanaCMS: Indeed, there is some code repetition for these queries. I will make a private method to help out

metis-normalization/src/main/java/eu/europeana/normalization/normalizers/CleanIRIViolationsNormalizer.java
stzanakis: Yes reducing the return statement to the minimum

metis-normalization/src/main/java/eu/europeana/normalization/normalizers/CleanIRIViolationsNormalizer.java
stzanakis: The isBlank method will return true for case where the string consists of only spaces and string.isEmpty will return false..
Also isBlank is null safe which means it returns true if the string is null, while string.isEmpty with throw an exception if it happens to reach that point and it's null..

metis-normalization/src/main/java/eu/europeana/normalization/normalizers/CleanIRIViolationsNormalizer.java
stzanakis: It's a good practice to have as less return values as possible in general.. Think about it like a GOTO operation which as you know it's bad practice..

metis-normalization/src/test/java/eu/europeana/normalization/normalizers/CleanIRIViolationsNormalizerTest.java
stzanakis: I see, makes sense! I'll leave it up to you to decide here then.
Same goes for the Spliterator.SORTED, as I said not crucial. If we know it's sorted then fine otherwise perhaps a SIZED or ORDER is more relevant?

metis-normalization/src/main/java/eu/europeana/normalization/normalizers/CleanIRIViolationsNormalizer.java
stzanakis: Also you can omit the nullity check here and simply check for StringUtils.isBlank(value)

metis-normalization/src/main/java/eu/europeana/normalization/normalizers/CleanIRIViolationsNormalizer.java
JoanaCMS: Oh, I did not know about that, will keep in mind for the next time

metis-normalization/src/test/java/eu/europeana/normalization/normalizers/CleanIRIViolationsNormalizerTest.java
JoanaCMS: Indeed, either SIZED or ORDER seem more relevant, will change it

metis-normalization/src/main/java/eu/europeana/normalization/normalizers/CleanIRIViolationsNormalizer.java
```diff
@@ -0,0 +1,95 @@
...
+    if (!StringUtils.isBlank(value)) {
+      iri = iriFactory.create(value);
+      final String normalizedValue;
+
+      try {
+        normalizedValue = iri.toURI().toString();
+        result = Collections.singletonList(new NormalizedValueWithConfidence(normalizedValue, 1));
+      } catch (URISyntaxException e) {
+        LOG.debug("There was some trouble normalizing the value for IRI Violation");
+        result =  Collections.emptyList();
```
stzanakis, line 74: You probably don't need to assign this value here do you?

metis-normalization/src/test/java/eu/europeana/normalization/normalizers/CleanIRIViolationsNormalizerTest.java
stzanakis: There still seems to be a public here

metis-normalization/src/main/java/eu/europeana/normalization/normalizers/CleanIRIViolationsNormalizer.java
```diff
@@ -0,0 +1,95 @@
...
+  private static final Namespace.Element RDF_RESOURCE = Namespace.RDF.getElement("resource");
+  private static final Namespace.Element RDF_ABOUT = Namespace.RDF.getElement("about");
+  private static final Namespace.Element ORE_AGGREGATION = Namespace.ORE.getElement("Aggregation");
+  private static final Namespace.Element EDM_IS_SHOWN_BY = Namespace.EDM.getElement("isShownBy");
+  private static final Namespace.Element EDM_IS_SHOWN_AT = Namespace.EDM.getElement("isShownAt");
+  private static final Namespace.Element EDM_OBJECT = Namespace.EDM.getElement("object");
+  private static final Namespace.Element EDM_HAS_VIEW = Namespace.EDM.getElement("hasView");
+  private static final Namespace.Element EDM_WEB_RESOURCE = Namespace.EDM.getElement("WebResource");
+
+  private final XpathQuery RESOURCE_IS_SHOWN_BY_QUERY = createResourceQueries(EDM_IS_SHOWN_BY);
```
stzanakis, line 39: Those field should be static right?

metis-normalization/src/main/java/eu/europeana/normalization/normalizers/CleanIRIViolationsNormalizer.java
JoanaCMS: You're right, let me fix that

metis-normalization/src/main/java/eu/europeana/normalization/normalizers/CleanIRIViolationsNormalizer.java
JoanaCMS: Yup, right, this is a mistake, will fix that

metis-normalization/src/test/java/eu/europeana/normalization/normalizers/CleanIRIViolationsNormalizerTest.java
JoanaCMS: Will do that

PR #374, opened 2020-09-09
Bug/MET-2949: Lack of authentication bearer token results in 400 error

JoanaCMS:


metis-core/metis-core-rest/src/main/java/eu/europeana/metis/core/rest/exception/RestResponseExceptionHandler.java
```diff
@@ -103,4 +104,29 @@ public StructuredExceptionWrapper handleMessageNotReadable(Exception exception,
...
+   * @param exception the exception thrown
+   * @param response the response that should be updated
+   * @return {@link StructuredExceptionWrapper} a json friendly class that contains the error message for the client
+   */
+  @ExceptionHandler(value = MissingRequestHeaderException.class)
+  @ResponseBody
+  public StructuredExceptionWrapper handleMissingRequestHeaderException(Exception exception,
+      HttpServletResponse response) {
+
+    String header = ((MissingRequestHeaderException) exception).getHeaderName();
```
stzanakis, line 119: You can define the exception parameter of type MissingRequestHeaderException and avoid the cast.
Also the method is quite simple and descriptive, so you could directly use exception.getHeaderName() on the if statement.

metis-core/metis-core-rest/src/main/java/eu/europeana/metis/core/rest/exception/RestResponseExceptionHandler.java
```diff
@@ -103,4 +104,29 @@ public StructuredExceptionWrapper handleMessageNotReadable(Exception exception,
...
+
+    StructuredExceptionWrapper output = new StructuredExceptionWrapper(exception.getMessage());
+    response.setStatus(HttpStatus.BAD_REQUEST.value());
+
+    if(header.equalsIgnoreCase("Authorization")) {
```
stzanakis, line 124: I would use an if/else statement here and declare the output final.
That makes it more understandable and you also avoid resetting/recreating the fields..
What do you think?

metis-core/metis-core-rest/src/main/java/eu/europeana/metis/core/rest/exception/RestResponseExceptionHandler.java
```diff
@@ -103,4 +104,29 @@ public StructuredExceptionWrapper handleMessageNotReadable(Exception exception,
...
+      response.setStatus(HttpStatus.UNAUTHORIZED.value());
+      output = new StructuredExceptionWrapper(
+          "Authorization token is missing.");
```
stzanakis, line 127: It's not the token that is missing but the whole header is not provided, perhaps write something along the lines "Authorization header is missing in the request"?

metis-core/metis-core-rest/src/main/java/eu/europeana/metis/core/rest/exception/RestResponseExceptionHandler.java
stzanakis, line 124: Not very important, I think good to know is that when comparing to a constant string it's better to use the constant first, like for example so:
"Authorization.equalsIgnoreCase(header)
That is to avoid nullpointerexception

stzanakis (changes requested):

metis-core/metis-core-rest/src/main/java/eu/europeana/metis/core/rest/exception/RestResponseExceptionHandler.java
JoanaCMS: That is a good point, will do that

metis-core/metis-core-rest/src/main/java/eu/europeana/metis/core/rest/exception/RestResponseExceptionHandler.java
JoanaCMS: That is more specific and better, will write that instead

metis-core/metis-core-rest/src/main/java/eu/europeana/metis/core/rest/exception/RestResponseExceptionHandler.java
JoanaCMS: Good point, will do that

metis-core/metis-core-rest/src/main/java/eu/europeana/metis/core/rest/exception/RestResponseExceptionHandler.java
JoanaCMS: I didn't think on these details, but that indeed sounds better, will do that

PR #375, opened 2020-09-10
MET-2826 Update handling of retryable external connections, to handle…

stzanakis:
… more network exceptions.

Also save on the fail message of the plugin the full stack trace on exceptions on WorkflowExecutor.

### DIFF
--- a/metis-common/src/main/java/eu/europeana/metis/utils/ExternalRequestUtil.java
+++ b/metis-common/src/main/java/eu/europeana/metis/utils/ExternalRequestUtil.java
@@ -1,17 +1,23 @@
 package eu.europeana.metis.utils;
 
-import com.mongodb.MongoSecurityException;
-import com.mongodb.MongoSocketException;
 import java.net.SocketException;
+import java.net.SocketTimeoutException;
+import java.net.UnknownHostException;
 import java.util.Collections;
 import java.util.Locale;
 import java.util.Map;
+import java.util.Map.Entry;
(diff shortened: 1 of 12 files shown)


metis-common/src/main/java/eu/europeana/metis/utils/ExternalRequestUtil.java
```diff
@@ -26,31 +32,46 @@
   private static final int MAX_RETRIES = 10;
   private static final int SLEEP_TIMEOUT = 500;
 
+  public static final Map<Class<?>, String> UNMODIFIABLE_MAP_WITH_NETWORK_EXCEPTIONS;
+
+  static {
```
JoanaCMS, line 37: Why did you write this static block?

stzanakis: So that I can initialize the final UNMODIFIABLE_MAP_WITH_NETWORK_EXCEPTIONS statically.
There are other alternatives for static initialization of course, I just chose to keep the same code that was used previously inside the WorkflowExecutor.
Does that make sense, or have you had something else in mind?

JoanaCMS: Don't have anything else in mind. I don't believe I have seen a static block being used before so I was surprised to see it here. But now that I think of it, this makes sense. Thanks for answering!

PR #376, opened 2020-09-10
feat/MET-2973_avoid_retrying_dereference_result_if_cache_has_null_or_emtpy_value

stzanakis:
Storing both null for a transformed entity and a vocabulary is not possible anymore.
Storing of null of the transformed entity and a non null vocabulary is possible, and indicates that for a given resource a transformed result was null

### DIFF
--- a/metis-dereference/metis-dereference-service/src/main/java/eu/europeana/metis/dereference/service/MongoDereferenceService.java
+++ b/metis-dereference/metis-dereference-service/src/main/java/eu/europeana/metis/dereference/service/MongoDereferenceService.java
@@ -32,8 +32,10 @@ import java.util.stream.Stream;
 import javax.xml.bind.JAXBContext;
 import javax.xml.bind.JAXBException;
 import javax.xml.transform.TransformerException;
+import org.apache.commons.lang.StringUtils;
 import org.apache.commons.lang3.tuple.ImmutablePair;
 import org.apache.commons.lang3.tuple.Pair;
+import org.bson.types.ObjectId;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
@@ -116,7 +118,8 @@ public class MongoDereferenceService implements DereferenceService {
       throws JAXBException, TransformerException, URISyntaxException {
 
     // Get the main object to dereference. If null, we are done.
-    final Pair<EnrichmentBase, Vocabulary> resource = retrieveCachedEntity(resourceId);
+    final Pair<EnrichmentBase, Vocabulary> resource = computeEnrichmentBaseVocabularyPair(
+        resourceId);
     if (resource == null) {
       return Collections.emptyList();
     }
@@ -125,10 +128,11 @@ public class MongoDereferenceService implements DereferenceService {
     final Function<String, EnrichmentBase> valueResolver = key -> {
       Pair<EnrichmentBase, Vocabulary> result;
       try {
-        result = retrieveCachedEntity(key);
+        result = computeEnrichmentBaseVocabularyPair(key);
         return result == null ? null : result.getLeft();
       } catch (JAXBException | TransformerException | URISyntaxException e) {
-        LOGGER.warn("Problem occurred while dereferencing broader resource " + key + ".", e);
+        LOGGER.warn(String.format("Problem occurred while dereferencing broader resource %s.", key),
+            e);
         return null;
       }
     };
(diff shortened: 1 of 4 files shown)


metis-dereference/metis-dereference-service/src/main/java/eu/europeana/metis/dereference/service/MongoDereferenceService.java
```diff
@@ -171,74 +172,125 @@ private void extractBroaderResources(EnrichmentBase resource, Set<String> destin
...
+      //If we have proper results store them
+      if (transformedEntityAndVocabularyPair != null) {
+        //Save entity
+        ProcessedEntity entityToCache =
+            (cachedEntity == null) ? new ProcessedEntity() : cachedEntity;
         entityToCache.setResourceId(resourceId);
-        entityToCache.setVocabularyId((vocabulary == null) ? null : vocabulary.getId().toString());
+        entityToCache.setXml(transformedEntityAndVocabularyPair.getLeft());
+        entityToCache
+            .setVocabularyId(transformedEntityAndVocabularyPair.getRight().getId().toString());
```
jochen-vermeulen, line 240: According to my IDE, getId() is already a string, so no need to call toString() on it.

metis-dereference/metis-dereference-service/src/main/java/eu/europeana/metis/dereference/service/MongoDereferenceService.java
```diff
@@ -171,74 +172,125 @@ private void extractBroaderResources(EnrichmentBase resource, Set<String> destin
...
+    ImmutablePair<String, Vocabulary> transformedEntityAndVocabularyPair = null;
+    if (originalEntity != null) {
+      // Transform the original entity.
+      for (Vocabulary vocabulary : vocabularyCandidates.getVocabularies()) {
+        final String transformedEntity = retrieveEntityAndTransform(vocabulary, originalEntity,
+            resourceId);
+        if (transformedEntity != null) {
+          transformedEntityAndVocabularyPair = new ImmutablePair<>(transformedEntity, vocabulary);
+          break;
+        }
```
jochen-vermeulen, line 282: I see that here (like you announced) if there is no vocabulary for a record, you return null and therefore it is not saved to the cache.
But is it not a good idea to save it to the cache regardless (with a null vocabulary)? If we can't find a vocabulary then we know that the next time we don't need to try. Could save some database access time.

metis-dereference/metis-dereference-service/src/main/java/eu/europeana/metis/dereference/service/MongoDereferenceService.java
```diff
@@ -171,74 +172,125 @@ private void extractBroaderResources(EnrichmentBase resource, Set<String> destin
...
+   * @param transformedEntityXml the previously transformed provider entity
+   * @param entityVocabulary the previously used vocabulary for the transformation
+   * @return a pair with the computed values
+   * @throws URISyntaxException if the resource identifier url is invalid
+   * @throws TransformerException if an exception occurred during transformation of the original
+   * entity
+   */
+  private Pair<String, Vocabulary> computeEntityAndVocabulary(String resourceId,
+      ProcessedEntity cachedEntity, String transformedEntityXml, Vocabulary entityVocabulary)
+      throws URISyntaxException, TransformerException {
```
jochen-vermeulen, line 221: I guess you extracted this code into a separate method due to concerns about complexity? I think you could perhaps split this up differently, though. You're passing too much information (the transformedEntityXml is also in the cachedEntity) and it does more than the name suggests (it also saves to the cache). Perhaps a rearrangement of the code makes more sense here?

jochen-vermeulen (changes requested):

metis-dereference/metis-dereference-service/src/main/java/eu/europeana/metis/dereference/service/MongoDereferenceService.java
stzanakis: But it's not, did your IDE mess up?

metis-dereference/metis-dereference-service/src/main/java/eu/europeana/metis/dereference/service/MongoDereferenceService.java
stzanakis: Well the functionality about the null vocabulary was replicated as before.
So a cache that would have a null value as vocabulary will be "retried", because at later stage that vocabulary would succeed. From what you are saying, we don't want that so we don't re-compute it and we just return null for that resourceId, right?
Also another reason was that a cache item can point to a vocabulary that does not exist anymore therefore it was null, but I can handle that in a different way..

metis-dereference/metis-dereference-service/src/main/java/eu/europeana/metis/dereference/service/MongoDereferenceService.java
jochen-vermeulen: Yes it did! :)

metis-dereference/metis-dereference-service/src/main/java/eu/europeana/metis/dereference/service/MongoDereferenceService.java
```diff
@@ -166,79 +171,158 @@ private void extractBroaderResources(EnrichmentBase resource, Set<String> destin
...
+    Vocabulary cachedVocabulary = null;
+    boolean cachedVocabularyChanged = false;
+    if (cachedEntity != null && StringUtils.isNotBlank(cachedEntity.getVocabularyId())) {
+      cachedVocabulary = vocabularyDao.get(cachedEntity.getVocabularyId());
+      cachedVocabularyChanged = cachedVocabulary == null;
+    }
+
+    // If we do not have any cached entity, we need to compute it
+    if (cachedEntity == null || cachedVocabularyChanged) {
+      transformedEntityVocabularyPair = retrieveAndTransformEntity(resourceId);
```
jochen-vermeulen, line 250: I think (but please verify) that if you just add a line saying 
```
saveEntity(resourceId, cachedEntity, transformedEntityVocabularyPair)
```
after this line, you can remove the whole `didVocabularyChange()` method. You're essentially checking the exact same condition here, right? What do you think?

stzanakis: Yes indeed! But I thought that on your previous comment you wanted the saving functionality outside of the method.
I can add it back in the method if you want

jochen-vermeulen: First of all, I didn't say that :innocent: but secondly: you did, it is now in it's own method... I think it would be good to put it back especially since it doesn't add complexity (it's just one method call now). Does that sound OK?

stzanakis: That's how I interpreted it.. :sweat_smile: 
I moved that call inside the method

PR #383, opened 2020-09-29
MET-3063 Fix indexing issue with lists that contains duplicate items.

stzanakis:


metis-indexing/src/main/java/eu/europeana/indexing/mongo/property/MongoPropertyUpdaterImpl.java
```diff
@@ -199,9 +200,10 @@ private static boolean equals(AbstractEdmEntity entity1, AbstractEdmEntity entit
       Function<T, List<P>> getter, Function<T, A> ancestorInfoGetter,
       MongoObjectUpdater<P, A> objectUpdater) {
     final A ancestorInformation = ancestorInfoGetter.apply(updated);
-    final UnaryOperator<List<P>> preprocessing = entities -> entities.stream()
+    final UnaryOperator<List<P>> preprocessing = entities -> new ArrayList<>(entities.stream()
         .map(entity -> objectUpdater.update(entity, ancestorInformation, null, null, mongoServer))
-        .collect(Collectors.toList());
+        .collect(Collectors.toMap(AbstractEdmEntity::getAbout, Function.identity(),
+            BinaryOperator.maxBy((o1, o2) -> 1))).values());
```
jochen-vermeulen, line 206: Why not replace
```
BinaryOperator.maxBy((o1, o2) -> 1))
```
With
```
(o1, o2) -> o1
```
? I think this is easier to understand and involves not creating a comparator.

stzanakis: Like it! Yes but o2 should be returned instead of o1.
I'll update it!

PR #385, opened 2020-10-05
MET-3011: Conversion to HttpClient Java 11

JoanaCMS:


metis-media-service/src/main/java/eu/europeana/metis/mediaprocessing/http/AbstractHttpClient.java
```diff
@@ -106,71 +79,139 @@ private void cleanConnections(PoolingHttpClientConnectionManager connectionManag
...
+    // Obtain header information.
+    final Optional<String> mimeType = httpResponse.headers().firstValue("Content-Type");
+    final long fileSize = httpResponse.headers().firstValueAsLong("Content-Length").orElse(0);
+
+    // Process the result.
+    final ContentRetriever content = httpResponse.body() == null ?
+        ContentRetriever.forEmptyContent() : httpResponse::body;
+    final R result;
+    try {
+      result = createResult(resourceEntry, actualUri, mimeType.orElse(null),
```
stzanakis, line 143: Same here, what if mimeType is blank, are we ok with that?

metis-media-service/src/main/java/eu/europeana/metis/mediaprocessing/http/AbstractHttpClient.java
```diff
@@ -106,71 +79,139 @@ private void cleanConnections(PoolingHttpClientConnectionManager connectionManag
...
+        httpResponse = makeHttpRequest(location, bodyWrapper);
 
-      // If aborted, provide a nicer message. Otherwise, just rethrow.
-      if (httpGet.isAborted()) {
-        throw new IOException("The request was aborted: it exceeded the time limit.", e);
-      }
-      throw e;
+        if(httpResponse != null){
+          statusCode = httpResponse.statusCode();
+          location = httpResponse.headers().firstValue("Location").isEmpty() ?
```
stzanakis, line 177: Isn't there a constant of "Location" on the http client libs?

metis-media-service/src/main/java/eu/europeana/metis/mediaprocessing/http/AbstractHttpClient.java
```diff
@@ -106,71 +79,139 @@ private void cleanConnections(PoolingHttpClientConnectionManager connectionManag
...
+        .GET()
+        .timeout(Duration.ofMillis(responseTimeout))
+        .uri(uri)
+        .build();
+
+    // Execute the request.
+    try {
+      httpResponse = httpClient.send(httpRequest, bodyWrapper);
+    } catch (InterruptedException | IOException interruptedException) {
+      LOGGER.info("A problem occurred while sending a request");
```
stzanakis, line 211: If the flow was indeed interrupted, we probably want to reset the interrupted flag here..

metis-media-service/src/main/java/eu/europeana/metis/mediaprocessing/http/AbstractHttpClient.java
```diff
@@ -106,71 +79,139 @@ private void cleanConnections(PoolingHttpClientConnectionManager connectionManag
...
 
-      // If aborted, provide a nicer message. Otherwise, just rethrow.
-      if (httpGet.isAborted()) {
-        throw new IOException("The request was aborted: it exceeded the time limit.", e);
-      }
-      throw e;
+        if(httpResponse != null){
+          statusCode = httpResponse.statusCode();
+          location = httpResponse.headers().firstValue("Location").isEmpty() ?
+              location : location.resolve(httpResponse.headers().firstValue("Location").get());
```
stzanakis, line 178: Why not better simplify to:
location = httpResponse.headers().firstValue("Location").map(location::resolve).orElse(location);

metis-media-service/src/main/java/eu/europeana/metis/mediaprocessing/http/AbstractHttpClient.java
```diff
@@ -106,71 +79,139 @@ private void cleanConnections(PoolingHttpClientConnectionManager connectionManag
...
+
+    return httpResponse;
   }
 
-  private static boolean httpCallIsSuccessful(int status) {
-    return status >= HTTP_SUCCESS_MIN_INCLUSIVE && status < HTTP_SUCCESS_MAX_EXCLUSIVE;
+  private HttpResponse<InputStream> makeHttpRequest(URI uri,
+      CancelableBodyWrapper<InputStream> bodyWrapper) {
+
+    HttpResponse<InputStream> httpResponse = null;
```
stzanakis, line 199: My IDE says that this shadows the field that has the exact same name

metis-media-service/src/main/java/eu/europeana/metis/mediaprocessing/http/AbstractHttpClient.java
```diff
@@ -106,71 +79,139 @@ private void cleanConnections(PoolingHttpClientConnectionManager connectionManag
...
+    } else {
+      throw new IOException("A problem occurred when sending the request");
+    }
+
+    final Optional<String> redirectUris = httpResponse.headers().firstValue("Location");
+    final URI actualUri;
+
+    // Do first check redirection and analysis
+    if(Family.familyOf(statusCode) == Family.REDIRECTION){
+      httpResponse = performRedirect(statusCode, resourceUlr.resolve(redirectUris.get()),
```
stzanakis, line 122: httpResponse according to performRedirect can return null, although it's not possible in this case.
Nonetheless nullity should be also checked here, or performRedirect shouldn't have the possibility to return null.

metis-transformation/metis-transformation-service/src/main/java/eu/europeana/metis/transformation/service/XsltTransformer.java
```diff
@@ -134,12 +140,19 @@ private static Templates getTemplates(String xsltUrl) throws CacheValueSupplierE
...
+        .uri(URI.create(xsltUrl))
+        .build();
+
     final TransformerFactory transformerFactory = new TransformerFactoryImpl();
     // We know where the xslt files are coming from, we consider them safe.
-    try (@SuppressWarnings("findsecbugs:URLCONNECTION_SSRF_FD") final InputStream xsltStream = new URL(
-            xsltUrl).openStream()) {
+    try {
+      final InputStream xsltStream = httpClient.send(httpRequest, BodyHandlers.ofInputStream())
+          .body();
```
stzanakis, line 153: The inputStream is not closed, perhaps use try with resources

metis-transformation/metis-transformation-service/src/main/java/eu/europeana/metis/transformation/service/XsltTransformer.java
```diff
@@ -134,12 +140,19 @@ private static Templates getTemplates(String xsltUrl) throws CacheValueSupplierE
...
       return transformerFactory.newTemplates(new StreamSource(xsltStream));
-    } catch (IOException | TransformerConfigurationException e) {
+    } catch (IOException | TransformerConfigurationException | InterruptedException e) {
```
stzanakis, line 155: If InterruptedException occurs and we are catching it, the interrupted flag should be reset

metis-validation/metis-validation-service/src/main/java/eu/europeana/validation/service/SchemaProvider.java
```diff
@@ -149,17 +159,17 @@ private File downloadZipIfNeeded(String zipLocation, String destinationDir)
...
-      return destinationFile;
-    } catch (IOException e) {
-      throw new SchemaProviderException("Unable to store schema file", e);
+    File destinationFile = new File(schemasLocation, ZIP_FILE_NAME);
+    final HttpResponse<Path> httpResponse;
+
+    try {
+      httpResponse = httpClient.send(httpRequest, BodyHandlers.ofFile(destinationFile.toPath()));
+      destinationFile = httpResponse.body().toFile();
+    } catch (IOException | InterruptedException e) {
```
stzanakis, line 168: If InterruptedException occurs and we are catching it, the interrupted flag should be reset

metis-transformation/metis-transformation-service/src/main/java/eu/europeana/metis/transformation/service/XsltTransformer.java
```diff
@@ -31,6 +35,8 @@
   private static final CacheWithExpirationTime<String, Templates> TEMPLATES_CACHE =
       new CacheWithExpirationTime<>();
 
+  private static HttpClient httpClient = HttpClient.newBuilder().build();
```
stzanakis, line 38: can be final

stzanakis (changes requested):

metis-media-service/src/main/java/eu/europeana/metis/mediaprocessing/http/AbstractHttpClient.java
```diff
@@ -106,71 +79,139 @@ private void cleanConnections(PoolingHttpClientConnectionManager connectionManag
...
+          maxNumberOfRedirects, bodyWrapper);
+      actualUri = httpResponse.uri();
+      if(actualUri == null){
+        throw new IOException("There was some trouble retrieving the uri");
+      }
+    } else if (Status.fromStatusCode(statusCode) != Status.OK) {
+      throw new IOException(
+          "Download failed of resource " + resourceUlr + ". Status code " + statusCode);
+    } else {
+      actualUri = httpResponse.uri();
```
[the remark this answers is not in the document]
JoanaCMS: You are right, I can just give this value to `actualUri` and check if it is null before performing the rest of the code. Will change that

metis-media-service/src/main/java/eu/europeana/metis/mediaprocessing/http/AbstractHttpClient.java
JoanaCMS: Unfortunately, there isn't any constants of the sort. That is one of the ways to retrieve the value

metis-media-service/src/main/java/eu/europeana/metis/mediaprocessing/http/AbstractHttpClient.java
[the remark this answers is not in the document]
JoanaCMS: In my point of view, if the `statusCode`  is redirect, then there should be a Location value, so I don't believe it will reach a point where the value is null.
However, this just made me realize I can instantiate the value `redirectUris` inside the `if` statement since the value is not used anywhere else. Also, even though I don't believe the value will ever be null, I can add the null check, as a sanity check

metis-media-service/src/main/java/eu/europeana/metis/mediaprocessing/http/AbstractHttpClient.java
JoanaCMS: Indeed, it is more simple, will do that

metis-media-service/src/main/java/eu/europeana/metis/mediaprocessing/http/AbstractHttpClient.java
JoanaCMS: Good catch, will change the name then

metis-media-service/src/main/java/eu/europeana/metis/mediaprocessing/http/AbstractHttpClient.java
JoanaCMS: Hmm ok, I'll make the method `performRedirct` throw an exception in case the response is still `null`

metis-media-service/src/main/java/eu/europeana/metis/mediaprocessing/http/AbstractHttpClient.java
JoanaCMS: Not quite sure if `mimeType` being blank scenario is handled in `createResult` method. I did check with Jochen if it was ok to pass a `null` `mimeType`, for which the method does handle if the value is `null`. 
I'll add a `isBlank` check

metis-media-service/src/main/java/eu/europeana/metis/mediaprocessing/http/AbstractHttpClient.java
JoanaCMS: Done

metis-transformation/metis-transformation-service/src/main/java/eu/europeana/metis/transformation/service/XsltTransformer.java
JoanaCMS: Done

metis-validation/metis-validation-service/src/main/java/eu/europeana/validation/service/SchemaProvider.java
JoanaCMS: Done

metis-transformation/metis-transformation-service/src/main/java/eu/europeana/metis/transformation/service/XsltTransformer.java
JoanaCMS: Done

metis-media-service/src/main/java/eu/europeana/metis/mediaprocessing/http/AbstractHttpClient.java
```diff
@@ -106,71 +82,153 @@ private void cleanConnections(PoolingHttpClientConnectionManager connectionManag
...
+      statusCode = httpResponse.statusCode();
+    } else {
+      throw new IOException("A problem occurred when sending the request");
+    }
 
-      // Process the result.
-      final ContentRetriever content = responseEntity == null ?
-          ContentRetriever.forEmptyContent() : responseEntity::getContent;
-      return createResult(resourceEntry, actualUri, mimeType, fileSize, content);
+    final URI actualUri = httpResponse.uri();
```
stzanakis, line 120: I think this will not behave the same as before..
Previously if it was a redirect the actualUri is taken after performRedirect method.

JoanaCMS: Aahh you are right, I missed that point, I will put it back to how it was and add a new null check on the `actualUri`

metis-media-service/src/main/java/eu/europeana/metis/mediaprocessing/http/AbstractHttpClient.java
```diff
@@ -106,71 +82,153 @@ private void cleanConnections(PoolingHttpClientConnectionManager connectionManag
...
+    }
 
-      // Shouldn't really happen.
-      throw new IOException("An unexpected exception occurred.", e);
 
-    } catch (IOException e) {
+    // Do first check redirection and analysis
+    if(Family.familyOf(statusCode) == Family.REDIRECTION){
+      final String redirectUris = Optional.ofNullable(httpResponse.headers().firstValue(HttpHeaders.LOCATION))
+          .map(Optional::get)
```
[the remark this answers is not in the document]
JoanaCMS: That is a good point, the `isNotBlank` method already checks if it is not null, no need for `Optional.ofNullable`.
And thanks for mentioning the message on the exception, I will add it

metis-media-service/src/main/java/eu/europeana/metis/mediaprocessing/http/AbstractHttpClient.java
```diff
@@ -106,71 +82,153 @@ private void cleanConnections(PoolingHttpClientConnectionManager connectionManag
...
+        .GET()
+        .timeout(Duration.ofMillis(responseTimeout))
+        .uri(uri)
+        .build();
+
+    // Execute the request.
+    try {
+      httpResponse = httpClient.send(httpRequest, bodyWrapper);
+    } catch (InterruptedException interruptedException) {
+      LOGGER.info("A problem occurred while sending a request");
```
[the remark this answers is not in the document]
JoanaCMS: Oops, forgot to change the message, will do that

metis-validation/metis-validation-service/src/main/java/eu/europeana/validation/service/SchemaProvider.java
```diff
@@ -149,17 +159,20 @@ private File downloadZipIfNeeded(String zipLocation, String destinationDir)
...
+
+    try {
+      httpResponse = httpClient.send(httpRequest, BodyHandlers.ofFile(destinationFile.toPath()));
+      destinationFile = httpResponse.body().toFile();
     } catch (IOException e) {
-      throw new SchemaProviderException("Unable to store schema file", e);
+      LOGGER.info(String.format("There was some trouble sending a request to %s", schemasLocation));
+    } catch (InterruptedException e){
+      Thread.currentThread().interrupt();
+      LOGGER.info(String.format("There was some trouble sending a request to %s", schemasLocation));
```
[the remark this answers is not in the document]
JoanaCMS: Will change it

metis-media-service/src/main/java/eu/europeana/metis/mediaprocessing/http/AbstractHttpClient.java
```diff
@@ -106,71 +82,153 @@ private void cleanConnections(PoolingHttpClientConnectionManager connectionManag
...
+      throw new IOException("There was some trouble retrieving the uri");
+    }
 
-      // Shouldn't really happen.
-      throw new IOException("An unexpected exception occurred.", e);
 
-    } catch (IOException e) {
+    // Do first check redirection and analysis
+    if(Family.familyOf(statusCode) == Family.REDIRECTION){
+      final String redirectUris = Optional.ofNullable(httpResponse.headers().firstValue(HttpHeaders.LOCATION))
```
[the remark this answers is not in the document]
JoanaCMS: Hmm that's a good point, thanks for bringing it up

metis-media-service/src/main/java/eu/europeana/metis/mediaprocessing/http/AbstractHttpClient.java
```diff
@@ -106,71 +82,153 @@ private void cleanConnections(PoolingHttpClientConnectionManager connectionManag
...
+    } else if (Status.fromStatusCode(statusCode) != Status.OK) {
+      throw new IOException(
+          String.format("Download failed of resource %s. Status code %s", resourceUri, statusCode));
+    }
 
-      // If aborted, provide a nicer message. Otherwise, just rethrow.
-      if (httpGet.isAborted()) {
-        throw new IOException("The request was aborted: it exceeded the time limit.", e);
+    // Obtain header information.
+    final String mimeType = Optional.ofNullable(httpResponse.headers().firstValue(HttpHeaders.CONTENT_TYPE))
```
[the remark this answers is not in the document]
JoanaCMS: Will do it

PR #391, opened 2020-10-28
MET-3125 Make unit tests in Depublish service in Metis Core

JoanaCMS:


metis-core/metis-core-service/src/test/java/eu/europeana/metis/core/dao/TestDepublishRecordIdDao.java
```diff
@@ -0,0 +1,178 @@
...
+    verify(provider, times(1)).getDatastore();
+    assertEquals(1L, result);
+  }
+
+  @Test
+  void getDepublishRecordIdsTest() {
+    final String datasetId = Integer.toString(TestObjectFactory.DATASETID);
+    final Set<String> setTest = new HashSet<>();
+    setTest.add("1004");
+    setTest.add("1005");
```
jochen-vermeulen, line 122: You can also do `final Set<String> setTest = Set.of("1004", "1005");`
The same goes I guess for the other methods.

metis-core/metis-core-service/src/test/java/eu/europeana/metis/core/dao/TestDepublishRecordIdDao.java
```diff
@@ -0,0 +1,178 @@
...
+
+  @Test
+  void markRecordIdsWithDepublicationStatusTest() {
+    final String datasetId = Integer.toString(TestObjectFactory.DATASETID);
+    final Set<String> setTest = new HashSet<>();
+    setTest.add("1007");
+
+    depublishRecordIdDao
+        .addRecords(setTest, datasetId, DepublicationStatus.PENDING_DEPUBLICATION, Instant.now());
+    reset(provider, depublishRecordIdDao);
```
jochen-vermeulen, line 163: This shouldn't be necessary!

metis-core/metis-core-service/src/test/java/eu/europeana/metis/core/service/TestDepublishRecordIdService.java
```diff
@@ -0,0 +1,153 @@
...
+    reset(depublishRecordIdService);
+  }
+
+  @Test
+  void addRecordIdsToBeDepublishedTest() throws GenericMetisException {
+    depublishRecordIdService.addRecordIdsToBeDepublished(metisUser, datasetId, "1002");
+
+    verify(authorizer, times(1)).authorizeWriteExistingDatasetById(metisUser, datasetId);
+    verify(depublishRecordIdService, times(1)).checkAndNormalizeRecordIds(any(), any());
+    verify(depublishRecordIdDao, times(1)).createRecordIdsToBeDepublished(any(), any());
```
jochen-vermeulen, line 71: Perhaps you can check here not just that the method is called, but also that it's called with the right data? 
Same in the mathods below.

jochen-vermeulen (changes requested):

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/service/DepublishRecordIdService.java
```diff
@@ -239,12 +238,32 @@ public boolean canTriggerDepublication(MetisUser metisUser, String datasetId)
     } else {
```
jochen-vermeulen, line 238: Only the else case is covered in your unit tests. Perhaps you can add a simple test case for the if case as well?

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/dao/DepublishRecordIdDao.java
```diff
@@ -133,7 +133,7 @@ public int createRecordIdsToBeDepublished(String datasetId, Set<String> candidat
     return recordIdsToAdd.size();
   }
```
jochen-vermeulen, line 134: In this class, there are several interesting lines of code that don't have test coverage. Perhaps you could look into some of those?

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/service/DepublishRecordIdService.java
JoanaCMS: You are right, I missed the `if` statement, I will add a new unit test

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/service/DepublishRecordIdService.java
```diff
@@ -239,12 +238,32 @@ public boolean canTriggerDepublication(MetisUser metisUser, String datasetId)
...
+      String recordIdsInSeparateLines) throws BadContentException {
+    return DepublishRecordIdUtils
+        .checkAndNormalizeRecordIds(datasetId, recordIdsInSeparateLines);
+  }
+
+  ResponseListWrapper<DepublishRecordIdView> createResponseListWrapper(List<DepublishRecordIdView> records, int page){
+    final ResponseListWrapper<DepublishRecordIdView> result = new ResponseListWrapper<>();
+    result.setResultsAndLastPage(records, depublishRecordIdDao.getPageSize(), page);
+    return result;
+  }
```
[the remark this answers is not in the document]
JoanaCMS: My main intention was to mock `ResponseListWrapper` and `verify` if the method `setResultsAndLastPage` was called. This was the solution I came up with, but if you have another suggestion I would like to hear

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/service/DepublishRecordIdService.java
```diff
@@ -239,12 +238,32 @@ public boolean canTriggerDepublication(MetisUser metisUser, String datasetId)
...
+
+  Workflow createNewWorkflow(){
+    return new Workflow();
+  }
+
+  DepublishPluginMetadata createNewDepublishPluginMetadata(){
+    return new DepublishPluginMetadata();
+  }
```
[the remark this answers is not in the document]
JoanaCMS: Same as what I said before, may intention was to mock these objects and `verify` if specific method from them were called. Any suggestion as to what I could do otherwise?

metis-core/metis-core-service/src/test/java/eu/europeana/metis/core/dao/TestDepublishRecordIdDao.java
```diff
@@ -0,0 +1,178 @@
...
+
+  @Test
+  void createRecordIdsToBeDepublishedTest() throws BadContentException {
+    final String datasetId = Integer.toString(TestObjectFactory.DATASETID);
+    final Set<String> setTest = new HashSet<>();
+    setTest.add("1001");
+
+    depublishRecordIdDao.createRecordIdsToBeDepublished(datasetId, setTest);
+
+    assertEquals(1, provider.getDatastore().find(DepublishRecordId.class).count());
```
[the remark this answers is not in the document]
JoanaCMS: That is true, I should check the record, will add that

metis-core/metis-core-service/src/test/java/eu/europeana/metis/core/dao/TestDepublishRecordIdDao.java
JoanaCMS: Indeed that is much simpler, will do that

metis-core/metis-core-service/src/test/java/eu/europeana/metis/core/dao/TestDepublishRecordIdDao.java
JoanaCMS: Because we call the method ` depublishRecordIdDao.addRecords`, the verifies later have a misleading number for the parameter `times`. I wanted to check the exact number of times each object and their method are called within the method `depublishRecordIdDao.markRecordIdsWithDepublicationStatus`, and not both. However if there is another way to do it, I would like to know

metis-core/metis-core-service/src/test/java/eu/europeana/metis/core/dao/TestDepublishRecordIdDao.java
```diff
@@ -0,0 +1,178 @@
...
+        .addRecords(setTest, datasetId, DepublicationStatus.DEPUBLISHED, Instant.now());
+
+    doReturn(mockDatastore).when(provider).getDatastore();
+    doReturn(mockQuery).when(mockDatastore).find(DepublishRecordId.class);
+    doReturn(1L).when(depublishRecordIdDao).deleteRecords(mockQuery);
+    depublishRecordIdDao.deletePendingRecordIds(datasetId, setTest);
+
+    verify(mockDatastore, times(1)).find(DepublishRecordId.class);
+    verify(mockQuery, times(3)).filter(any());
+    assertEquals(0, provider.getDatastore().find(DepublishRecordId.class).count());
```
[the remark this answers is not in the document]
JoanaCMS: I have fixed this problem and it is passing

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/service/DepublishRecordIdService.java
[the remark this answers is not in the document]
JoanaCMS: Solved

metis-core/metis-core-service/src/test/java/eu/europeana/metis/core/dao/TestDepublishRecordIdDao.java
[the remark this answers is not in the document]
JoanaCMS: I have fixed it

metis-core/metis-core-service/src/test/java/eu/europeana/metis/core/service/TestDepublishRecordIdService.java
JoanaCMS: Like we talked before, some methods (like `addRecordIdsToBeDepublished` and `deletePendingRecordIdsTest`) are mostly dependent on other objects and on their maethods, more specifically the object `DepublishRecordIdDao`, and those specific methods are already being tested on a different unit test class. So I believe in this case it is not necessary to check the values.
However, I have added these checks on the values on the other methods

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/dao/DepublishRecordIdDao.java
JoanaCMS: Created new unit tests for the `if else` clauses

PR #403, opened 2020-12-11
Feat/met 3177 allow for concurrency on plugin level

stzanakis:


metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/execution/WorkflowExecutor.java
```diff
@@ -170,36 +200,95 @@ private Date runInqueueOrRunningStateWorkflowExecution() {
...
+      try {
+        //Update date for workflow execution so that it does not become stale
+        final Date updatedDate = new Date();
+        executablePlugin.setUpdatedDate(updatedDate);
+        workflowExecution.setUpdatedDate(updatedDate);
+        if (workflowExecution.getWorkflowStatus() != WorkflowStatus.RUNNING) {
+          workflowExecution.setStartedDate(updatedDate);
+          workflowExecution.setWorkflowStatus(WorkflowStatus.RUNNING);
+        }
+        workflowExecutionDao.updateMonitorInformation(workflowExecution);
```
jochen-vermeulen, line 263: Are you sure you need to do this? I think this whole block won't be necessary (especially if the semaphore acquire returns immediately), and what's more, it interferes with some logic below (for instance line 302-304). 

If you really want to signal that the execution is not stale, perhaps just set the execution update time and leave it at that?

jochen-vermeulen (changes requested):

metis-core/metis-core-rest/src/main/java/eu/europeana/metis/core/rest/config/OrchestratorConfig.java
```diff
@@ -218,24 +214,29 @@ public ProxiesService getProxiesService(WorkflowExecutionDao workflowExecutionDa
...
   public WorkflowPostProcessor workflowPostProcessor(DepublishRecordIdDao depublishRecordIdDao,
       DatasetDao datasetDao, WorkflowExecutionDao workflowExecutionDao, DpsClient dpsClient) {
-    return new WorkflowPostProcessor(depublishRecordIdDao, datasetDao, workflowExecutionDao, dpsClient);
+    return new WorkflowPostProcessor(depublishRecordIdDao, datasetDao, workflowExecutionDao,
+        dpsClient);
+  }
+
+  @Bean
+  public SemaphoresPerPluginManager semaphoresPerPluginManager() {
+    return new SemaphoresPerPluginManager(propertiesHolder.getFailsafeMarginOfInactivityInSecs(),
```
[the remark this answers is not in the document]
stzanakis: I'm removing the timeout where it's used, so I'll remove the parameter from here too

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/execution/PersistenceProvider.java
```diff
@@ -1,30 +1,37 @@
...
+import org.redisson.api.RedissonClient;
 
 class PersistenceProvider {
 
   private final Channel rabbitmqPublisherChannel;
   private final Channel rabbitmqConsumerChannel;
+  private final SemaphoresPerPluginManager semaphoresPerPluginManager;
   private final WorkflowExecutionDao workflowExecutionDao;
   private final WorkflowPostProcessor workflowPostProcessor;
   private final RedissonClient redissonClient;
```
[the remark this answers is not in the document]
stzanakis: But it is used..

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/execution/QueueConsumer.java
```diff
@@ -66,94 +70,123 @@ public QueueConsumer(Channel rabbitmqConsumerChannel, String rabbitmqQueueName,
...
-      super.getChannel().basicNack(rabbitmqEnvelope.getDeliveryTag(), false, true);
-      LOGGER.debug("NACK sent for {} with tag {}", objectId, rabbitmqEnvelope.getDeliveryTag());
-    }
+  private void sendAck(Envelope rabbitmqEnvelope, String objectId) throws IOException {
+    // Send ACK back to remove from queue asap.
+    super.getChannel().basicAck(rabbitmqEnvelope.getDeliveryTag(), false);
+    LOGGER.debug("ACK sent for {} with tag {}", objectId, rabbitmqEnvelope.getDeliveryTag());
   }
 
+  @Scheduled(fixedDelay = 5000)
```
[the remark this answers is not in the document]
stzanakis: I cannot use the previous option as a parameter on the annotation.
I don't think we need it configurable tbh, it's a very simple internal thing in the class.
If we really want it configurable I can add a Timer instead.

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/execution/QueueConsumer.java
```diff
@@ -66,94 +70,123 @@ public QueueConsumer(Channel rabbitmqConsumerChannel, String rabbitmqQueueName,
...
-      WorkflowExecutor workflowExecutor = new WorkflowExecutor(objectId, persistenceProvider,
-          workflowExecutionSettings, workflowExecutionMonitor);
-      completionService.submit(workflowExecutor);
-      threadsCounter++;
-      sendAck = true;
-    }
-    return sendAck;
+  private void submitExecution(String objectId) {
+    WorkflowExecutor workflowExecutor = new WorkflowExecutor(objectId, workflowExecutorManager,
+        workflowExecutionSettings, workflowExecutionMonitor);
```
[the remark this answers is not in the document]
stzanakis: Well the issue is that it has to go through the plugins and check if it can run per plugin and not just claim it like before. 
And to do so the semaphore should be acquired..
So even if one plugin is run the next one might not be allowed.
We can do what you say, but I thought it would complicate it more without much benefit?

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/execution/WorkflowExecutor.java
```diff
@@ -82,83 +88,107 @@
...
     // Claim the execution: if this claim is denied, we stop this execution.
-    LOGGER.info("Claiming workflow execution with id: {}", workflowExecutionId);
-    this.workflowExecution = workflowExecutionMonitor.claimExecution(this.workflowExecutionId);
-    if (this.workflowExecution == null) {
-      LOGGER.info("Discarding WorkflowExecution with id: {}, it could not be claimed.",
-          workflowExecutionId);
-      return null;
+    LOGGER.info("workflowExecutionId: {} - Claiming workflow execution", workflowExecutionId);
+    final Pair<WorkflowExecution, Boolean> workflowExecutionClaimedPair = workflowExecutionMonitor
+        .claimExecution(this.workflowExecutionId);
```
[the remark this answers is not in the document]
stzanakis: That's true but the claiming leniency of some execution being stale is quite small, so it shouldn't have a big effect here actually.
Do you still feel we need to add this state and if so where were you thinking of adding it, in the plugin level in the workflow execution level or both?

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/execution/SemaphoresPerPluginManager.java
```diff
@@ -0,0 +1,57 @@
...
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Manages a map of {@link ExecutablePluginType} keys and {@link Semaphore} values.
+ * <p>Each executable plugin type contains it's own semaphore so that access to those plugin
+ * types is controlled.</p>
+ */
+public class SemaphoresPerPluginManager {
+
+  private final int acquireTimeoutSecs;
```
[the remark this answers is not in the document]
stzanakis: I guess so indeed!

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/execution/WorkflowExecutor.java
stzanakis: The exact same thing is happening inside the claiming function.
But if the acquire timeout is removed then I guess this is not required here.

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/execution/QueueConsumer.java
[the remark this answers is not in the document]
stzanakis: Moved this to the OrchestratorConfig together with the other periodic threads

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/execution/QueueConsumer.java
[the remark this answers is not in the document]
stzanakis: So after discussing this, I moved only the claiming functionality to the QueueConsumer

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/execution/WorkflowExecutor.java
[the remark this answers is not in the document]
stzanakis: After discussion, when a plugin is not allowed to run then only setting the INQUEUE status of the workflowExecution should suffice.

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/execution/QueueConsumer.java
```diff
@@ -100,63 +98,97 @@ public QueueConsumer(Channel rabbitmqConsumerChannel, String rabbitmqQueueName,
...
+              + "could not be claimed.", workflowExecution.getId());
+      sendAck = false;
+    } else {
+      LOGGER.info("workflowExecutionId: {} - Does not have a claimable status, discarding message",
+          workflowExecution.getId());
+    }
+    if (sendAck) {
+      sendAck(rabbitmqEnvelope, objectId);
+    } else {
+      sendNack(rabbitmqEnvelope, objectId);
```
jochen-vermeulen, line 127: Should this be in a finally block so that we know it is always sent?

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/execution/SemaphoresPerPluginManager.java
```diff
@@ -38,12 +34,9 @@ public SemaphoresPerPluginManager(int acquireTimeoutSecs, int permissionsPerSema
    *
    * @param executablePluginType the provided executable plugin type
    * @return true if acquisition was successful, false otherwise
-   * @throws InterruptedException if we were interrupted during acquisition
    */
-  public boolean tryAcquireForExecutablePluginType(ExecutablePluginType executablePluginType)
-      throws InterruptedException {
-    return unmodifiableMaxThreadsPerPlugin.get(executablePluginType)
-        .tryAcquire(acquireTimeoutSecs, TimeUnit.SECONDS);
+  public boolean tryAcquireForExecutablePluginType(ExecutablePluginType executablePluginType) {
```
jochen-vermeulen, line 38: Because this no longer throws an InterruptedException, it seems like `WorkflowExecutor.runMetisPluginWithSemaphoreAllocation()` no longer throws one either. It can be removed from that method declaration, and from there possibly it has further effects?

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/execution/QueueConsumer.java
```diff
@@ -100,63 +98,97 @@ public QueueConsumer(Channel rabbitmqConsumerChannel, String rabbitmqQueueName,
...
+    LOGGER.info("workflowExecutionId: {} - Received from queue.", objectId);
+    LOGGER.info("workflowExecutionId: {} - Claiming workflow execution", objectId);
+    Pair<WorkflowExecution, Boolean> workflowExecutionClaimedPair = workflowExecutionMonitor
+        .claimExecution(objectId);
+    WorkflowExecution workflowExecution = workflowExecutionClaimedPair.getLeft();
+    final Boolean workflowClaimed = workflowExecutionClaimedPair.getRight();
 
+    boolean sendAck = true;
+    if (workflowExecution == null) {
+      // This execution no longer exists and we need to ignore it.
```
jochen-vermeulen, line 110: It could also mean that an exception occurred retrieving/claiming it (see `WorkflowExecutionMonitor.claimExecution()`). Is this to be treated differently?

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/execution/WorkflowExecutionMonitor.java
```diff
@@ -182,34 +185,38 @@ public void performFailsafe() {
...
-        //Update dates
-        final Date now = new Date();
-        workflowExecution.setUpdatedDate(now);
-        if (workflowExecution.getWorkflowStatus() != WorkflowStatus.RUNNING) {
-          workflowExecution.setStartedDate(now);
-          workflowExecution.setWorkflowStatus(WorkflowStatus.RUNNING);
+      workflowExecution = workflowExecutionDao.getById(workflowExecutionId);
+
+      if (workflowExecution != null && CLAIMABLE_STATUSES
+          .contains(workflowExecution.getWorkflowStatus())) {
```
jochen-vermeulen, line 198: I think you could leave out the second condition (the claimable statuses one)? It is the very first check in the method `mayClaimExecution` and there it actually prints a nice log statement.

stzanakis: Okay, I'm changing it slightly though, so that the Set is used more explicitly

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/execution/QueueConsumer.java
stzanakis: No. I would say like before we are assuming that it will be picked up from the failsafe if that occurs

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/execution/SemaphoresPerPluginManager.java
stzanakis: That's true, wonder why my IDE didn't say anything..
Removing it, there shouldn't be any side effect so we are good on that perspective

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/execution/QueueConsumer.java
stzanakis: Wouldn't matter much, but wouldn't hurt having it in a finally block either

PR #407, opened 2020-12-21
Feat/met 3128 record parsing and augmenting

jochen-vermeulen:


metis-enrichment/metis-enrichment-common/src/main/java/eu/europeana/enrichment/api/internal/AbstractReferenceTerm.java
```diff
@@ -1,40 +1,40 @@
...
   @Override
-  public boolean equals(Object other){
-    if(other == this){
+  public final boolean equals(Object otherObject) {
+    if (otherObject == this) {
       return true;
     }
-
-    if(other == null || getClass() != other.getClass()){
+    if (!(otherObject instanceof ReferenceTerm)) {
```
stzanakis, line 28: Why change this?
It's comparing here the interface type which it shouldn't and it's not symmetrical.
I think sonar will complain for this too

metis-enrichment/metis-enrichment-common/src/main/java/eu/europeana/enrichment/api/internal/AbstractSearchTerm.java
```diff
@@ -1,46 +1,45 @@
...
   @Override
-  public boolean equals(Object other){
-    if(other == this){
+  public final boolean equals(Object otherObject) {
+    if (otherObject == this) {
       return true;
     }
-
-    if(other == null || getClass() != other.getClass()){
+    if (!(otherObject instanceof SearchTerm)) {
```
stzanakis, line 24: Same here

metis-enrichment/metis-enrichment-common/src/main/java/eu/europeana/enrichment/api/internal/SearchTerm.java
```diff
@@ -1,14 +1,35 @@
...
 public interface SearchTerm {
 
-  List<EntityType> getCandidateTypes();
+  Set<EntityType> getCandidateTypes();
 
   String getTextValue();
 
   String getLanguage();
 
+  /**
```
stzanakis, line 19: Equals and hashCode shouldn't be in the interface imo.
The implementer should decide if the equals should be applied on the implementing class or not.
Btw they were removed on a previous review.

metis-enrichment/metis-enrichment-common/src/main/java/eu/europeana/enrichment/api/internal/ReferenceTerm.java
```diff
@@ -2,11 +2,32 @@
...
+ * types. Two reference terms are equal if and only if they have the same reference and candidate
+ * types.
+ */
 public interface ReferenceTerm {
 
-  List<EntityType> getCandidateTypes();
+  Set<EntityType> getCandidateTypes();
 
   URL getReference();
+
```
stzanakis, line 17: Same here 

"Equals and hashCode shouldn't be in the interface imo.
The implementer should decide if the equals should be applied on the implementing class or not.
Btw they were removed on a previous review."

metis-enrichment/metis-enrichment-common/src/main/java/eu/europeana/enrichment/api/internal/ReferenceTermContext.java
```diff
@@ -17,22 +20,11 @@ public ReferenceTermContext(URL reference, Set<FieldType> fieldTypes) {
   }
 
   @Override
-  public List<EntityType> getCandidateTypes() {
-    return fieldTypes.stream().map(FieldType::getEntityType).collect(Collectors.toList());
+  public Set<EntityType> getCandidateTypes() {
+    return fieldTypes.stream().map(FieldType::getEntityType).collect(Collectors.toSet());
   }
 
-  @Override
```
stzanakis, line 24: I think equals and hashcode should be declared here too, preferably with using the new fields or just by using super

metis-enrichment/metis-enrichment-common/src/main/java/eu/europeana/enrichment/api/internal/SearchTermContext.java
```diff
@@ -17,27 +19,11 @@ public SearchTermContext(String textValue, String language, Set<FieldType> field
...
+  public Set<EntityType> getCandidateTypes() {
+    return fieldTypes.stream().map(FieldType::getEntityType).collect(Collectors.toSet());
   }
 
   public Set<FieldType> getFieldTypes() {
-    return new HashSet<>(fieldTypes);
+    return Collections.unmodifiableSet(fieldTypes);
   }
-
-  @Override
```
stzanakis, line 28: Same here 

"I think equals and hashcode should be declared here too, preferably with using the new fields or just by using super"

metis-enrichment/metis-enrichment-common/src/main/java/eu/europeana/enrichment/api/internal/SearchTermImpl.java
```diff
@@ -0,0 +1,23 @@
...
+
+  public SearchTermImpl(String textValue, String language, Set<EntityType> entityTypes) {
+    super(textValue, language);
+    this.entityTypes = Set.copyOf(entityTypes);
+  }
+
+  @Override
+  public Set<EntityType> getCandidateTypes() {
+    return Collections.unmodifiableSet(entityTypes);
+  }
```
stzanakis, line 22: Same here

"I think equals and hashcode should be declared here too, preferably with using the new fields or just by using super"

metis-enrichment/metis-enrichment-common/src/main/java/eu/europeana/enrichment/api/internal/ReferenceTermImpl.java
```diff
@@ -0,0 +1,24 @@
...
+
+  public ReferenceTermImpl(URL reference, Set<EntityType> entityTypes) {
+    super(reference);
+    this.entityTypes = Set.copyOf(entityTypes);
+  }
+
+  @Override
+  public Set<EntityType> getCandidateTypes() {
+    return Collections.unmodifiableSet(entityTypes);
+  }
```
stzanakis, line 23: Same here:

"I think equals and hashcode should be declared here too, preferably with using the new fields or just by using super"

metis-enrichment/metis-enrichment-common/src/main/java/eu/europeana/enrichment/api/internal/AbstractReferenceTerm.java
```diff
@@ -1,40 +1,40 @@
...
   @Override
-  public URL getReference(){
+  public URL getReference() {
     return reference;
   }
 
   @Override
-  public boolean equals(Object other){
-    if(other == this){
+  public final boolean equals(Object otherObject) {
```
stzanakis, line 24: Why final here, were are enforcing equals/hashcode in the hierarchy which I believe we should not

metis-enrichment/metis-enrichment-common/src/main/java/eu/europeana/enrichment/api/internal/AbstractSearchTerm.java
```diff
@@ -1,46 +1,45 @@
...
+  public AbstractSearchTerm(String textValue, String language) {
     this.textValue = textValue;
-    this.language = LanguageCodes.convert(language);
+    this.language = language;
   }
 
   @Override
-  public boolean equals(Object other){
-    if(other == this){
+  public final boolean equals(Object otherObject) {
```
stzanakis, line 20: Same here:

"Why final here, were are enforcing equals/hashcode in the hierarchy which I believe we should not"

stzanakis (changes requested):
My comments mostly reflect code principles, more specifically for equals/hashcode where I think we should not transparently enforce implementations of it. But rather give freedom to the implementer. The fact that we use/implement them in a specific manner shouldn't need to be enforced to another implementation.
Any strong opinions on this?

jochen-vermeulen:
The implementation of equals and hashcode were made to comply with an old version of the EntityResolver interface (where it was important for the input objects to strictly satisfy equality conditions). But I changed the interface later to use generics, so that these strict conditions are no longer needed (but they were kept in place).

I now removed them, in this case that's fine. But perhaps in other situations there may be cases like this... anyway, that's not important now.

PR #410, opened 2021-01-08
Bug/met 3196 fix fail status when restart core

stzanakis:


metis-common/metis-common-network/src/main/java/eu/europeana/metis/network/ExternalRequestUtil.java
```diff
@@ -249,7 +249,7 @@ public static boolean isNullOrEmpty(final Map<?, ?> m) {
     return m == null || m.isEmpty();
   }
 
-  private static Throwable getCause(Throwable e) {
+  public static Throwable getCause(Throwable e) {
```
jochen-vermeulen, line 252: Should have javadoc?

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/execution/QueueConsumer.java
```diff
@@ -214,6 +214,8 @@ private void checkCollectedWorkflowExecution(
    */
   public void close() {
     threadPool.shutdown();
+    //Interrupt running threads
+    threadPool.shutdownNow();
```
jochen-vermeulen, line 218: I think this is overkill. The first call does not block, so if you're going to interrupt everything, might as well do it straight away.

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/execution/WorkflowExecutor.java
```diff
@@ -355,26 +368,37 @@ private void periodicCheckingLoop(long sleepTime, AbstractExecutablePlugin plugi
...
-        }
-        consecutiveCancelOrMonitorFailures++;
-        LOGGER.warn(String.format(
-            "workflowExecutionId: %s, pluginType: %s - Monitoring of external task failed %s "
-                + "consecutive times. After exceeding %s retries, pending status will be set",
-            workflowExecution.getId(), plugin.getPluginType(), consecutiveCancelOrMonitorFailures,
-            MAX_CANCEL_OR_MONITOR_FAILURES), e);
-        if (consecutiveCancelOrMonitorFailures >= MAX_CANCEL_OR_MONITOR_FAILURES) {
-          plugin.setPluginStatusAndResetFailMessage(PluginStatus.PENDING);
+        } else {
```
jochen-vermeulen, line 380: Else is not needed here - the if has a return at the end.

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/execution/WorkflowExecutor.java
```diff
@@ -355,26 +368,37 @@ private void periodicCheckingLoop(long sleepTime, AbstractExecutablePlugin plugi
...
+          LOGGER.warn(String
+              .format("workflowExecutionId: %s, pluginType: %s - ExternalTaskException occurred.",
+                  workflowExecution.getId(), plugin.getPluginType()), e);
+          if (!ExternalRequestUtil.doesExceptionCauseMatchAnyOfProvidedExceptions(
+              UNMODIFIABLE_MAP_WITH_NETWORK_EXCEPTIONS_AND_UNREADABLE_MESSAGE, e)) {
```
jochen-vermeulen, line 385: Is it not simpler to add the extra condition here in the if, instead of creating a copy of the map? If we want to undo it later, to have it here explicitly as an extra check is easier to revert than take out the new map (and remembering to put the old one in its place). Just a suggestion... what do you think?

metis-common/metis-common-network/src/main/java/eu/europeana/metis/network/ExternalRequestUtil.java
stzanakis: Fixed

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/execution/QueueConsumer.java
stzanakis: Wouldn't matter much, I guess we can avoid the first call

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/execution/WorkflowExecutor.java
stzanakis: Fixed

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/execution/WorkflowExecutor.java
stzanakis: I was in between the two but then I thought I was being more concise by using the map but I guess I can add to the if instead.

PR #411, opened 2021-01-11
MET-3244: Tiffs classified as content tier 0 bug

JoanaCMS:


metis-media-service/src/main/java/eu/europeana/metis/mediaprocessing/extraction/ThumbnailGenerator.java
```diff
@@ -270,6 +270,10 @@ private static void closeAllThumbnailsSilently(List<ThumbnailWithSize> thumbnail
     final String colorResultFormat = "\n" + contentMarker + "\n%c\n" + contentMarker;
     command.addAll(Arrays.asList("-colorspace", "sRGB", "-dither", "Riemersma", "-remap",
         colormapFile, "-format", colorResultFormat, "histogram:info:"));
+
+    // To suppress warnings that can come up. This flag needs
+    // to be at the beginning of the command to work
+    command.add(1, "-quiet");
```
jochen-vermeulen, line 276: I think the command should be compiled in order, so the quiet flag should be added to line 250. That makes this code a lot more understandable, and robust. Imagine trying to add something else in the beginning without realizing that this thing changes the command afterwards?

JoanaCMS: That is a good point, didn't think of that. Will change it

PR #424, opened 2021-02-16
MET-3282: Remove web resource meta info from mongo if web resource ha…

jochen-vermeulen:
…s none.

### DIFF
--- a/metis-indexing/src/main/java/eu/europeana/indexing/mongo/WebResourceMetaInfoUpdater.java
+++ b/metis-indexing/src/main/java/eu/europeana/indexing/mongo/WebResourceMetaInfoUpdater.java
@@ -7,6 +7,7 @@ import eu.europeana.corelib.definitions.edm.model.metainfo.ImageMetaInfo;
 import eu.europeana.corelib.definitions.edm.model.metainfo.TextMetaInfo;
 import eu.europeana.corelib.definitions.edm.model.metainfo.VideoMetaInfo;
 import eu.europeana.corelib.edm.model.metainfo.WebResourceMetaInfoImpl;
+import eu.europeana.indexing.mongo.property.MongoObjectManager;
 import eu.europeana.indexing.mongo.property.MongoPropertyUpdater;
 import eu.europeana.indexing.mongo.property.MongoPropertyUpdaterFactory;
 import eu.europeana.metis.mongo.dao.RecordDao;
@@ -25,8 +26,9 @@ import org.slf4j.LoggerFactory;
 /**
  * Field updater for instances of {@link WebResourceMetaInfoImpl}.
  */
-public class WebResourceMetaInfoUpdater
-    extends AbstractMongoObjectUpdater<WebResourceMetaInfoImpl, WebResourceInformation> {
+public class WebResourceMetaInfoUpdater extends
+        AbstractMongoObjectUpdater<WebResourceMetaInfoImpl, WebResourceInformation> implements
+        MongoObjectManager<WebResourceMetaInfoImpl, WebResourceInformation> {
 
   private static final Logger LOGGER = LoggerFactory.getLogger(WebResourceMetaInfoUpdater.class);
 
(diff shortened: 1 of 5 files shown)


metis-indexing/src/main/java/eu/europeana/indexing/mongo/property/MongoObjectUpdater.java
```diff
@@ -6,20 +6,21 @@
 /**
  * Updater of mongo objects.
  *
- * @param <R> The type of the record to update.
+ * @param <R> The type of the object to update.
  * @param <A> The type of the ancestor information (information from parents).
  */
 public interface MongoObjectUpdater<R, A> {
```
stzanakis, line 12: Why not just add a default method here for the delete that does nothing and it's only implemented on the classes that needed, in this case WebResourceMetaInfoUpdater?

metis-indexing/src/main/java/eu/europeana/indexing/mongo/WebResourceMetaInfoUpdater.java
```diff
@@ -25,8 +26,9 @@
 /**
  * Field updater for instances of {@link WebResourceMetaInfoImpl}.
  */
-public class WebResourceMetaInfoUpdater
-    extends AbstractMongoObjectUpdater<WebResourceMetaInfoImpl, WebResourceInformation> {
+public class WebResourceMetaInfoUpdater extends
```
stzanakis, line 29: Please check my comment on MongoObjectUpdater interface.
I think there is another approach which will avoid this extra layer of interface, which I think provides a cleaner approach..
What do you think?

metis-indexing/src/main/java/eu/europeana/indexing/mongo/property/MongoPropertyUpdaterImpl.java
```diff
@@ -266,12 +266,14 @@ private static boolean equals(AbstractEdmEntity entity1, AbstractEdmEntity entit
...
       Function<T, WebResourceInformation> ancestorInfoGetter,
-      Supplier<MongoObjectUpdater<WebResourceMetaInfoImpl, WebResourceInformation>> updaterSupplier) {
+      Supplier<MongoObjectManager<WebResourceMetaInfoImpl, WebResourceInformation>> updaterSupplier) {
     final WebResourceMetaInfo entity = Optional.of(updated).map(getter).orElse(null);
     final WebResourceInformation ancestorInformation = ancestorInfoGetter.apply(updated);
-    if (entity != null) {
-      updaterSupplier.get()
-          .update((WebResourceMetaInfoImpl) entity, ancestorInformation, null, null, mongoServer);
+    if (entity == null) {
+      updaterSupplier.get().delete(ancestorInformation, mongoServer);
```
stzanakis, line 273: Should we be checking nullity on ancestorInformation too?
I know that what is provided is not null, but for extra safety?

metis-indexing/src/main/java/eu/europeana/indexing/mongo/property/MongoObjectUpdater.java
jochen-vermeulen: Here are my reasons:
1. Not all objects can ever be removed (we already have removal code for whole records).
2. It would go against consistency. We would not call those remove methods for all subclasses. This specific one is called from code that deals with metadata meta info only, and I think it would be overkill to implement object removal everywhere just on the chance that someone later implements a functional removal method. It follows that it would create the false impression that by actually implementing the remove method, we could enable removal. In summary: there is a lack of consistency in that situation.
3. I also wanted to keep the change relatively minimal (small). Perhaps this may be a hotfix.
Does this make sense? Or would you still prefer to merge the two interfaces?

metis-indexing/src/main/java/eu/europeana/indexing/mongo/property/MongoPropertyUpdaterImpl.java
jochen-vermeulen: I could, but it would be redundant, I think? Currently, if it is null it would result in a NullPointerException in the code that is called (either deletion or update). And that's what we'd expect, I think. So all we would do is create some lines of code to check and create an exception that you would get anyway.
Or am I missing something?

metis-indexing/src/main/java/eu/europeana/indexing/mongo/property/MongoObjectUpdater.java
stzanakis: But you don't have to implement it anywhere, that's the idea of the default method on the interface.
Did you mean implementing it on all subclasses?

metis-indexing/src/main/java/eu/europeana/indexing/mongo/property/MongoPropertyUpdaterImpl.java
stzanakis: No you  are not, I just thought for extra security but I guess you're right and we are safe here.

PR #428, opened 2021-02-26
MET-3245: Fixing the bug

JoanaCMS:


metis-enrichment/metis-enrichment-common/src/main/java/eu/europeana/enrichment/api/external/SearchValue.java
```diff
@@ -36,9 +36,11 @@
   /**
    * Internally duplicates are removed when value is set or unmarshalled.
    */
+  //TODO: 26-02-2021 We initialized the List to fix the bug MET-3245. A error message would show
+  //TODO: when this field was missing. This is not the best fix and we should figure out another solution
   @XmlElement(name = "entityType")
   @JsonProperty("entityType")
-  private List<EntityType> entityTypes;
+  private List<EntityType> entityTypes = new ArrayList<>();
```
stzanakis, line 43: The required attribute on the @XmlElement and @JsonProperty is not working for this case?

stzanakis: Never mind by comment, it seems to be by default to false so that wouldn't work

PR #429, opened 2021-03-01
MET-2812: Add application name to Mongo configuration options.

jochen-vermeulen:


metis-common/metis-common-mongo/src/main/java/eu/europeana/metis/mongo/connection/MongoClientProvider.java
```diff
@@ -101,6 +107,8 @@ public MongoClientProvider(MongoProperties<E> properties, Builder clientSettings
     if (mongoCredential != null) {
       clientSettingsBuilder.credential(mongoCredential);
     }
+    Optional.ofNullable(properties.getApplicationName())
+            .ifPresent(clientSettingsBuilder::applicationName);
```
stzanakis, line 111: I think we should also validation for blankness of this value, otherwise it can be mistakenly overwrite the default value

jochen-vermeulen: Sounds good. Fixed.

PR #431, opened 2021-03-09
Feat/met 2215 incremental oai harvesting option

jochen-vermeulen:


metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/dao/DataEvolutionUtils.java
```diff
@@ -517,6 +355,81 @@ private static boolean pluginHasSuccessfulRecords(
...
+      final boolean incrementalHarvest = (rootHarvest.getPlugin() instanceof OaipmhHarvestPlugin)
+              && ((OaipmhHarvestPlugin) rootHarvest.getPlugin()).getPluginMetadata()
+              .isIncrementalHarvest();
+      if (!incrementalHarvest) {
+        break;
+      }
+    }
+
+    // Done. Sort and return the harvests in the right order.
+    return resultHarvests.values().stream()
```
stzanakis, line 428: Why are we ordering here?
Also they are already inversely ordered from the previous list

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/dao/DataEvolutionUtils.java
```diff
@@ -517,6 +355,81 @@ private static boolean pluginHasSuccessfulRecords(
...
+
+    // Compile a list of all associated harvests (that led to one of these publish operations).
+    // Note: we assume that workflows don't cross each other (i.e. an earlier publish cannot have a
+    // later harvest). We stop when we find a full harvest (the latest full harvest).
+    final Map<ExecutedMetisPluginId, PluginWithExecutionId<? extends ExecutablePlugin<?>>> resultHarvests = new HashMap<>();
+    for (Pair<IndexToPublishPlugin, WorkflowExecution> publishOperation : publishOperations) {
+
+      // If the publish is not available, we have detected an anomaly. We are done.
+      if (MetisPlugin.getDataStatus(publishOperation.getLeft()) != DataStatus.VALID &&
+              MetisPlugin.getDataStatus(publishOperation.getLeft()) != DataStatus.DEPRECATED) {
```
stzanakis, line 403: So DEPRECATED plugins are ok?
Shouldn't this be allowed for only VALID ones?

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/dao/DataEvolutionUtils.java
```diff
@@ -517,6 +355,81 @@ private static boolean pluginHasSuccessfulRecords(
...
+   * harvest that is the root predecessor of an index to publish).
+   * </p>
+   *
+   * @param datasetId The dataset ID for which to obtain the chain.
+   * @return The chain, in the form of plugin-execution pairs that are ordered chronologically. Is
+   * never null, but can be empty if no such chain exists (i.e. the dataset does not have a
+   * published harvest or we find an index after the last full harvest that is invalid or did
+   * somehow not originate from a harvest).
+   */
+  public List<PluginWithExecutionId<? extends ExecutablePlugin<?>>> getPublishedHarvestIncrements(
```
stzanakis, line 375: Don't we need a unit test for this?

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/dao/WorkflowValidationUtils.java
```diff
@@ -0,0 +1,216 @@
...
+      if (pluginMetadata instanceof OaipmhHarvestPluginMetadata) {
+        final OaipmhHarvestPluginMetadata oaipmhMetadata = (OaipmhHarvestPluginMetadata) pluginMetadata;
+        final URI validatedUri = validateUrl(oaipmhMetadata.getUrl());
+        oaipmhMetadata
+                .setUrl(new URIBuilder(validatedUri).removeQuery().setFragment(null).toString());
+        oaipmhMetadata.setMetadataFormat(oaipmhMetadata.getMetadataFormat() == null ? null
+                : oaipmhMetadata.getMetadataFormat().trim());
+        oaipmhMetadata.setSetSpec(
+                oaipmhMetadata.getSetSpec() == null ? null : oaipmhMetadata.getSetSpec().trim());
+        if (oaipmhMetadata.isIncrementalHarvest() && !isIncrementalHarvestingAllowed(datasetId)) {
```
stzanakis, line 158: We should capture this case in a unit test

metis-core/metis-core-rest/src/main/java/eu/europeana/metis/core/rest/OrchestratorController.java
```diff
@@ -309,6 +309,46 @@ public WorkflowExecution getWorkflowExecutionByExecutionId(
...
+  @ResponseBody
+  public IncrementalHarvestingAllowedView isIncrementalHarvestingAllowed(
+          @RequestHeader("Authorization") String authorization,
+          @PathVariable("datasetId") String datasetId) throws GenericMetisException {
+    final MetisUser metisUser = authenticationClient.getUserByAccessTokenInHeader(authorization);
+    return new IncrementalHarvestingAllowedView(
+            orchestratorService.isIncrementalHarvestingAllowed(metisUser, datasetId));
+  }
+
+  static class IncrementalHarvestingAllowedView {
```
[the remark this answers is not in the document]
jochen-vermeulen: OK.

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/dao/WorkflowExecutionDao.java
```diff
@@ -814,13 +821,13 @@ private Pagination createPagination(int firstPage, Integer pageCount,
     return new Pagination(skip, limit, maxRequested);
   }
 
-  private static class Pagination {
+  static class Pagination {
```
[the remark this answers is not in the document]
jochen-vermeulen: Yes, the class can be final, but we don't typically mark all classes as final that can be. Perhaps to facilitate unit tests we should leave it as is? What do you think?

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/dao/WorkflowExecutionDao.java
```diff
@@ -284,7 +284,7 @@ public String existsAndNotCompleted(String datasetId) {
    *                    values.
    * @return the first plugin found
    */
-  public PluginWithExecutionId<MetisPlugin> getFirstSuccessfulPlugin(String datasetId,
+  public PluginWithExecutionId<MetisPlugin<?>> getFirstSuccessfulPlugin(String datasetId,
```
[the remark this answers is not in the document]
jochen-vermeulen: That's an interesting point, as I did some of this to get rid of some warnings showing up in my IDE. But some others have now appeared, so it didn't quite work:)
There was another reason though, I got an untyped cast exception somewhere, which set me on the path to work on this in the first place.
I decided to keep it this way because it forces us to think more about the typing in our code, but I didn't expect that it would affect our sonarQube. What do you think?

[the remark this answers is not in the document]
jochen-vermeulen: Actually, perhaps I can do some cleverness with untyped interfaces to solve this issue.

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/dao/DataEvolutionUtils.java
```diff
@@ -517,6 +355,81 @@ private static boolean pluginHasSuccessfulRecords(
...
+      // If the publish is not available, we have detected an anomaly. We are done.
+      if (MetisPlugin.getDataStatus(publishOperation.getLeft()) != DataStatus.VALID &&
+              MetisPlugin.getDataStatus(publishOperation.getLeft()) != DataStatus.DEPRECATED) {
+        return Collections.emptyList();
+      }
+
+      // Get the root harvest and add it to the map.
+      final PluginWithExecutionId<? extends ExecutablePlugin<?>> rootHarvest = getRootAncestor(
+              new PluginWithExecutionId<>(publishOperation.getRight(), publishOperation.getLeft()));
+      resultHarvests.put(ExecutedMetisPluginId.forPlugin(rootHarvest.getPlugin()), rootHarvest);
```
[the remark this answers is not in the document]
jochen-vermeulen: Ah, I see now. Yes, you're right.

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/dao/DataEvolutionUtils.java
jochen-vermeulen: You'll notice that they are in a map (which is just easier to use when filtering out duplicates). So they are not sorted at all.

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/dao/DataEvolutionUtils.java
jochen-vermeulen: Sure, we could. I'll see if I can add one within reasonable time.

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/dao/WorkflowValidationUtils.java
jochen-vermeulen: Yes, I can add this.

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/dao/WorkflowExecutionDao.java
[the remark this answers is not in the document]
stzanakis: Hmm not sure I follow, this is a good case were this class could be declared final, we don't expect other classes to inherit from it. Also why do you think there issue with unit tests? They should still work in they way they are now right?

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/dao/WorkflowExecutionDao.java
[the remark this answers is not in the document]
stzanakis: I think that the current sonarqube issue is "worse" than the previous issue.. What I mean is, I think before it was more an IDE issue while now it has become a "Major code smell".
I'm okay with both tbh but we should take action for at least the latter if we leave as is.
For the scope if this ticket I guess we can move forward, but we should try to take action with the idea you had as a follow up, and/or somehow mark those cases in Sonar.
My first intuition is to do an immediate follow up with your idea, what do you think?

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/dao/DataEvolutionUtils.java
stzanakis: Sorry I think I didn't express myself properly, so maybe I'll try another way:
So there is a list which is first sorted publishOperations.
Then that list is iterated and there are keys inserted on the map basically in order since the publishOperations is sorted previously.
So if we simply change the type of map from HashMap to LinkedHashMap, that means that the map will already be sorted at the return statement right? (It's inversely sorted though so not sure if that matter, but in this case it doesn't matter if I'm not mistaken?)

Hope I expressed it better this time around:)

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/dao/DataEvolutionUtils.java
stzanakis: Yes exactly so why would we allow an incremental harvest to happen if the datase status for the plugin execution is deprecated? I understand this method returns the chain of harvests but, it is used with the purpose to identify if an incremental harvest is allowed or not here eu.europeana.metis.core.dao.WorkflowValidationUtils#isIncrementalHarvestingAllowed
Or I'm I understanding this wrongly?

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/dao/WorkflowExecutionDao.java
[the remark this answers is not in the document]
jochen-vermeulen: They should, yes. But mocking or spying on a final class does not work. Anyway, I'll change it, it is really not a problem at this point, I think.

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/dao/DataEvolutionUtils.java
jochen-vermeulen: Imagine this sequence of events:
1. Full harvest followed by index
2. Incremental harvest followed by index
3. Reindex (no longer really belongs to either of the previous index operations, but 'deprecates' the last index)
4. Incremental harvest

This last incremental harvest SHOULD be allowed, right? There is no reason to stop an incremental harvest from happening after a reindex. We know that the exact same records are published as before (with possibly different properties), so we should just be able to go ahead with an incremental harvest here.
If we don't allow DEPRECATED plugins, we have a problem here.

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/dao/DataEvolutionUtils.java
jochen-vermeulen: OK, I replaced it with a LinkedHashMap.

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/dao/DataEvolutionUtils.java
stzanakis: I see you are right we should allow this!
I think in my mind was more executions other than publish that were deprecated during reindex and that this method would allow indexing without re-running those(perhaps for previous increments, if those previous revision increments were used in combination with the "new" ones), but I guess that doesn't apply..

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/dao/WorkflowExecutionDao.java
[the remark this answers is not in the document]
jochen-vermeulen: Sure, sounds good. You're right, I don't think we should keep it this way, and I'd like to find the nicest way to solve this. If I have time this sprint I will do a tech debt ticket on this. But given the time constraints, it is not currently part of this ticket. Does that sound acceptable?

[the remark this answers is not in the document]
stzanakis: Yeah sounds good!

PR #432, opened 2021-03-11
Feat/met 2707 maven version upgrades

stzanakis:


metis-harvesting/src/test/java/eu/europeana/metis/harvesting/oaipmh/CloseableHttpOaiClientTest.java
```diff
@@ -1,73 +1,102 @@
...
-public class CloseableHttpOaiClientTest extends WiremockHelper {
+public class CloseableHttpOaiClientTest {
 
+  private static int portForWireMock = 9999;
+
+  static {
+    try {
+      portForWireMock = NetworkUtil.getAvailableLocalPort();
+    } catch (IOException e) {
+      e.printStackTrace();
```
JoanaCMS, line 30: Could you remove this `printStackTrace` and write a log instead?
Even for a class test, I believe we should avoid using `printStackTrace` and make it easier for everyone with a log

stzanakis: True that could be an improvement.
Although I found a better way to initialize wiremock, have a look!

PR #433, opened 2021-03-12
MET-3419 Use new configuration in Metis

JoanaCMS:


metis-common/metis-common-mongo/src/main/java/eu/europeana/metis/mongo/connection/MongoProperties.java
```diff
@@ -78,14 +78,19 @@ public void setAllProperties(String[] hosts, int[] ports, String authenticationD
    * @param password The password. Can be null, in which case no authentication takes place.
    * @param enableSsl Whether to enable SSL connections.
    * @param readPreferenceValue The read preference. Can be null, where then the default applies
+   * @param applicationName
    * @throws E In case either of the arrays is null, or their lengths don't match.
    */
   public void setAllProperties(String[] hosts, int[] ports, String authenticationDatabase,
-      String username, String password, boolean enableSsl, ReadPreferenceValue readPreferenceValue)
+      String username, String password, boolean enableSsl, ReadPreferenceValue readPreferenceValue,
+      String applicationName)
```
stzanakis, line 86: We need to be  cautious here, this is a library used by others.
I think it there might be a conflict in corelib at least.
It would be better to create another method and keep the already existent that calls that "new" method with null as the applicationName to get the default.
We could even deprecate the old method if we want to, but we should have follow up tickets to update the corresponding projects

metis-enrichment/metis-enrichment-rest/src/main/java/eu/europeana/enrichment/rest/config/Application.java
```diff
@@ -56,6 +56,8 @@
   @Value("${enrichment.mongo.database}")
   private String enrichmentMongoDatabase;
 
+  @Value("${mongo.application.name}")
```
stzanakis, line 59: Maybe follow the naming pattern here(with the enrichment prefix)?

metis-common/metis-common-mongo/src/main/java/eu/europeana/metis/mongo/connection/MongoProperties.java
```diff
@@ -78,14 +78,19 @@ public void setAllProperties(String[] hosts, int[] ports, String authenticationD
    * @param password The password. Can be null, in which case no authentication takes place.
    * @param enableSsl Whether to enable SSL connections.
    * @param readPreferenceValue The read preference. Can be null, where then the default applies
+   * @param applicationName
```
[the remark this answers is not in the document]
JoanaCMS: Just added it

metis-common/metis-common-mongo/src/main/java/eu/europeana/metis/mongo/connection/MongoProperties.java
```diff
@@ -78,14 +78,19 @@ public void setAllProperties(String[] hosts, int[] ports, String authenticationD
...
-      String username, String password, boolean enableSsl, ReadPreferenceValue readPreferenceValue)
+      String username, String password, boolean enableSsl, ReadPreferenceValue readPreferenceValue,
+      String applicationName)
       throws E {
     setAllProperties(hosts, ports, authenticationDatabase, username, password);
     this.mongoEnableSsl = enableSsl;
     setReadPreferenceValue(readPreferenceValue);
+
+    // Set the application name
+    setApplicationName(applicationName);
```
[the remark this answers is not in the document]
JoanaCMS: Removed

metis-enrichment/metis-enrichment-rest/src/main/java/eu/europeana/enrichment/rest/config/Application.java
JoanaCMS: Good point, I'm going to take care of it

metis-enrichment/metis-enrichment-rest/src/main/resources/enrichment.properties.example
```diff
@@ -8,4 +8,5 @@ socks.proxy.password=
 #Mongo
 enrichment.mongo.host=
 enrichment.mongo.database=
-enrichment.mongo.port=
\ No newline at end of file
+enrichment.mongo.port=
+mongo.application.name=
```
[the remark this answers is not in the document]
JoanaCMS: Taken care of

metis-common/metis-common-mongo/src/main/java/eu/europeana/metis/mongo/connection/MongoProperties.java
JoanaCMS: That's a good catch, I was not aware.
I made a "new" method of `setAllProperties` without the `applicationName` input variable. I also added the `@Deprecated` annotation to it and set the default value of application name as empty.

metis-common/metis-common-mongo/src/main/java/eu/europeana/metis/mongo/connection/MongoProperties.java
```diff
@@ -80,12 +80,39 @@ public void setAllProperties(String[] hosts, int[] ports, String authenticationD
...
    * @throws E In case either of the arrays is null, or their lengths don't match.
    */
+  @Deprecated
   public void setAllProperties(String[] hosts, int[] ports, String authenticationDatabase,
       String username, String password, boolean enableSsl, ReadPreferenceValue readPreferenceValue)
       throws E {
     setAllProperties(hosts, ports, authenticationDatabase, username, password);
     this.mongoEnableSsl = enableSsl;
     setReadPreferenceValue(readPreferenceValue);
+    setApplicationName("");
```
stzanakis, line 90: This method should call the "new" method below with "null" applicationName parameter, as it is not it's overwriting the default value with the empty String

metis-enrichment/metis-enrichment-rest/src/main/java/eu/europeana/enrichment/rest/config/Application.java
```diff
@@ -55,6 +55,8 @@
   private int enrichmentMongoPort;
   @Value("${enrichment.mongo.database}")
   private String enrichmentMongoDatabase;
+  @Value("${enrichment.mongo.application.name}")
+  private String mongoApplicationName;
```
stzanakis, line 59: How about aligning the name of the field too?:)

metis-common/metis-common-mongo/src/main/java/eu/europeana/metis/mongo/connection/MongoProperties.java
JoanaCMS: Done

metis-enrichment/metis-enrichment-rest/src/main/java/eu/europeana/enrichment/rest/config/Application.java
JoanaCMS: Oops, will do that :)

PR #437, opened 2021-03-19
MET-3448 Review Indexing according to API/Search requirements

JoanaCMS:


metis-indexing/src/main/java/eu/europeana/indexing/solr/SolrDocumentPopulator.java
```diff
@@ -159,4 +160,21 @@ public void populateWithFacets(SolrInputDocument document, RdfWrapper rdf) {
       document.addField(EdmLabel.FACET_VALUE_CODES.toString(), code);
     }
   }
+
+  private List<AggregationImpl> getDataProviderAggregations(FullBeanImpl fullBean){
+
+    List<String> proxyInResult = fullBean.getProxies()
+        .stream()
+        .filter(x -> !x.isEuropeanaProxy() && x.getLineage().length == 0)
```
stzanakis, line 168: I think there is a chance here a null pointer might be thrown that we should cover..
Also there is this new Predicate.not() in java 11.
How about replacing this filter with:

```suggestion
        .filter(not(ProxyImpl::isEuropeanaProxy))
        .filter(x -> ArrayUtils.isEmpty(x.getLineage()))
```

metis-indexing/src/main/java/eu/europeana/indexing/solr/SolrDocumentPopulator.java
```diff
@@ -21,6 +21,7 @@
 import eu.europeana.indexing.utils.WebResourceLinkType;
 import eu.europeana.indexing.utils.WebResourceWrapper;
 import eu.europeana.metis.schema.model.MediaType;
+import java.util.ArrayList;
```
stzanakis, line 24: Unused import

JoanaCMS: Removed

metis-indexing/src/main/java/eu/europeana/indexing/solr/SolrDocumentPopulator.java
JoanaCMS: Good point, I'll add this change

PR #439, opened 2021-03-25
MET-3479 Change Exception from Enrichment Service

JoanaCMS:


metis-enrichment/metis-enrichment-rest/src/main/java/eu/europeana/enrichment/rest/EnrichmentController.java
```diff
@@ -76,8 +78,13 @@ public EnrichmentResultList search(
   @ResponseBody
   @ApiResponses(value = {@ApiResponse(code = 400, message = "Error processing the result")})
   public EnrichmentBase equivalence(@ApiParam("uri") @RequestParam("uri") String uri) {
-    List<EnrichmentBase> result =  enrichmentService
-        .enrichByEquivalenceValues(new ReferenceValue(uri, Collections.emptySet()));
+    List<EnrichmentBase> result;
+    try {
+      result = enrichmentService
+          .enrichByEquivalenceValues(new ReferenceValue(uri, Collections.emptySet()));
+    } catch (IllegalArgumentException e) {
```
stzanakis, line 85: It would be better to not have a try catch here and handle it with the ControllerAdvice/RestResponseExceptionHandler that is in the project, similarly to what we do with other projects.
Same for the method below

JoanaCMS: I thought that it would be best to return a `BAD_REQUEST` since the issue for that exception to be thrown is incorrect input.
Nevertheless, I will remove the try catch from both methods

stzanakis: Yes you are right, the idea is to extend the ControllerAdvice to handle this case properly.
So what I meant is remove the try catch, let the ControllerAdvice handle it but update it properly so that it can recognize it and configure it as BAD_REQUEST as you had in the try block

JoanaCMS: There is already a `RestResponseExceptionHandler` class for `EnrichmentController` with the `@ControllerAdvice` annotation that handles every exception by throwing a `BAD_REQUEST`. Since my scenario is already handled by this one, there is no need to add a new method to do the same thing but for `IllegalArgumentException`

stzanakis: Btw I was not referring on creating a new @ControllerAdvice, but rather on extending the already existent with a new @ExceptionHandler to separate cases based on error and map them on per case basis to an http code and/or message, similarly to what we do with metis-core for example.
But this case seems to be ok to be handled from the generic one that is already there, so I'll merge this as is now

PR #442, opened 2021-04-06
MET-3531 Common web resources added for aggregations added only to pr…

stzanakis:
…ovider aggregation

### DIFF
--- a/metis-indexing/src/main/java/eu/europeana/indexing/fullbean/AggregationFieldInput.java
+++ b/metis-indexing/src/main/java/eu/europeana/indexing/fullbean/AggregationFieldInput.java
@@ -10,9 +10,8 @@ import java.util.Locale;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
+import java.util.Set;
 import java.util.function.Function;
-import java.util.stream.Collectors;
-import org.apache.commons.collections.CollectionUtils;
 
 /**
  * Converts a {@link Aggregation} from an {@link eu.europeana.metis.schema.jibx.RDF} to a {@link
@@ -20,35 +19,31 @@ import org.apache.commons.collections.CollectionUtils;
  */
 final class AggregationFieldInput implements Function<Aggregation, AggregationImpl> {
 
-  private final List<WebResourceImpl> recordWebResources;
-  private final List<WebResourceImpl> referencedWebResources;
+  private final Map<String, WebResourceImpl> recordWebResourcesMap;
+  private final Set<String> referencedWebResourceAbouts;
 
-  AggregationFieldInput(final List<WebResourceImpl> recordWebResources,
-      List<WebResourceImpl> referencedWebResources) {
-    this.recordWebResources = recordWebResources;
-    this.referencedWebResources = referencedWebResources;
+  AggregationFieldInput(final Map<String, WebResourceImpl> recordWebResourcesMap,
+      Set<String> referencedWebResourceAbouts) {
+    this.recordWebResourcesMap = recordWebResourcesMap;
+    this.referencedWebResourceAbouts = referencedWebResourceAbouts;
   }
 
   private String processResource(List<WebResourceImpl> aggregationWebResources,
       ResourceType resource) {
     String resourceString = Optional.ofNullable(resource).map(ResourceType::getResource)
         .map(String::trim).orElse(null);
-    final boolean isNotAlreadyReferenced = referencedWebResources.stream().map(WebResourceImpl::getAbout)
-        .noneMatch(about -> about.equals(resourceString));
-    if (resourceString != null && isNotAlreadyReferenced) {
-      final List<WebResourceImpl> matchingWebResources = recordWebResources.stream()
-          .filter(webResource -> webResource.getAbout().equals(resourceString))
-          .collect(Collectors.toList());
+    if (resourceString != null && !referencedWebResourceAbouts.contains(resourceString)) {
+      final WebResourceImpl matchingWebResource = recordWebResourcesMap.get(resourceString);
 
-      if (CollectionUtils.isNotEmpty(matchingWebResources)) {
-        aggregationWebResources.addAll(matchingWebResources);
-        referencedWebResources.addAll(matchingWebResources);
-        recordWebResources.removeAll(matchingWebResources);
+      if (matchingWebResource != null) {
+        aggregationWebResources.add(matchingWebResource);
+        referencedWebResourceAbouts.add(matchingWebResource.getAbout());
+        recordWebResourcesMap.remove(matchingWebResource.getAbout());
       } else {
         WebResourceImpl webResource = new WebResourceImpl();
         webResource.setAbout(resourceString);
         aggregationWebResources.add(webResource);
-        referencedWebResources.add(webResource);
+        referencedWebResourceAbouts.add(webResource.getAbout());
       }
     }
     return resourceString;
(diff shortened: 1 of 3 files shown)


metis-indexing/src/main/java/eu/europeana/indexing/fullbean/RdfToFullBeanConverter.java
```diff
@@ -63,15 +66,44 @@ public FullBeanImpl convertRdfToFullBean(RdfWrapper record) {
...
   }
 
+  private List<Aggregation> convertAggregations(RdfWrapper record) {
+    final Supplier<List<WebResourceImpl>> webResourcesExtractor = new WebResourcesExtractor(record);
+    //Convert the provider aggregations
+    final List<AggregationImpl> providerAggregations = convertList(record.getProviderAggregations(),
+        new AggregationFieldInput(webResourcesExtractor), false);
+
+    //Compute the referenced web resources of the provider aggregation
+    final List<String> providerWebResourceReferences = Optional.ofNullable(providerAggregations)
```
jochen-vermeulen, line 87: Again: is it not better to put the IDs in a set?

metis-indexing/src/main/java/eu/europeana/indexing/fullbean/RdfToFullBeanConverter.java
```diff
@@ -63,15 +66,44 @@ public FullBeanImpl convertRdfToFullBean(RdfWrapper record) {
...
+        .orElse(Collections.emptyList()).stream().map(AggregationImpl::getWebResources)
+        .flatMap(List::stream).map(WebResource::getAbout).collect(Collectors.toList());
+
+    //Create supplier for the web resources of the aggregator aggregation
+    final Supplier<List<WebResourceImpl>> aggregatorAggregationWebResources = () -> webResourcesExtractor
+        .get().stream().filter(
+            webResourceImpl -> !providerWebResourceReferences.contains(webResourceImpl.getAbout()))
+        .collect(Collectors.toList());
```
jochen-vermeulen, line 95: I think there is a problem here, in your current implementation. I think that the code is such that all web resources are always added to the provider aggregation, so this supplier always returns an empty list, and empty web resources will be added to the aggregator aggregation if it references something.

I think the algorithm may be fixed by providing AggregationFieldInput with a lookup for webresources (by about value) instead of a supplier. And then make sure that it only adds what it references (taking care to remove web resources from the lookup if it has added it to an aggregation). Then you call this for all aggregations in the right order (provider aggregation first). And then at the end you  add all remaining (unclaimed) web resources to the (first) provider aggregation.

Does that sound reasonable?

metis-indexing/src/main/java/eu/europeana/indexing/utils/RdfWrapper.java
```diff
@@ -167,6 +168,35 @@ private static boolean isEuropeanaProxy(ProxyType proxy) {
...
+   * @return the list of aggregations
+   */
+  public List<Aggregation> getAggregatorAggregations() {
+    return getAggregations(proxyType -> !CollectionUtils.isEmpty(proxyType.getLineageList()));
+  }
+
+  private List<Aggregation> getAggregations(Predicate<? super ProxyType> proxyTypePredicate) {
+    final List<String> proxyInList = record.getProxyList().stream().filter(
+        proxyType -> proxyType.getEuropeanaProxy() == null || !proxyType.getEuropeanaProxy()
+            .isEuropeanaProxy()).filter(proxyTypePredicate).map(ProxyType::getProxyInList)
```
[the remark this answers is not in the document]
stzanakis: Done

metis-indexing/src/main/java/eu/europeana/indexing/utils/RdfWrapper.java
```diff
@@ -167,6 +168,35 @@ private static boolean isEuropeanaProxy(ProxyType proxy) {
...
+        .flatMap(List::stream).map(ProxyIn::getResource).collect(Collectors.toList());
```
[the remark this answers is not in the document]
stzanakis: Not relevant anymore

metis-indexing/src/main/java/eu/europeana/indexing/utils/RdfWrapper.java
```diff
@@ -167,6 +168,35 @@ private static boolean isEuropeanaProxy(ProxyType proxy) {
     return getPropertyList(record.getAggregationList());
   }
 
+  /**
+   * Extract provider aggregations
+   *
+   * @return the list of aggregations
+   */
```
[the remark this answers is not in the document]
stzanakis: Ok added some jdoc

metis-indexing/src/main/java/eu/europeana/indexing/fullbean/RdfToFullBeanConverter.java
stzanakis: Not relevant anymore

metis-indexing/src/main/java/eu/europeana/indexing/fullbean/RdfToFullBeanConverter.java
stzanakis: Yeah I see! I also see another issue with this..
The new implementation should cover also duplicate references for potential double provider aggregations.
Hope now it's better, plz have a look!

metis-indexing/src/main/java/eu/europeana/indexing/fullbean/AggregationFieldInput.java
```diff
@@ -20,22 +20,36 @@
  */
 final class AggregationFieldInput implements Function<Aggregation, AggregationImpl> {
 
-  private final Supplier<List<WebResourceImpl>> webResourcesSupplier;
+  private final List<WebResourceImpl> recordWebResources;
+  private final List<WebResourceImpl> referencedWebResources;
```
jochen-vermeulen, line 24: The code below would read a lot better if these were maps (with the about values as keys). Do you think that would be a useful change?

metis-indexing/src/main/java/eu/europeana/indexing/fullbean/RdfToFullBeanConverter.java
```diff
@@ -78,26 +78,28 @@ public FullBeanImpl convertRdfToFullBean(RdfWrapper record) {
   }
 
   private List<Aggregation> convertAggregations(RdfWrapper record) {
-    final Supplier<List<WebResourceImpl>> webResourcesExtractor = new WebResourcesExtractor(record);
+    //The record web resources is reduced every time one of it's web resources gets referenced
+    final List<WebResourceImpl> recordWebResources = new ArrayList<>(new WebResourcesExtractor(record).get());
+    //The reference list is being extended every time a new web resource is referenced from an aggregator
+    final List<WebResourceImpl> referencedWebResources = new ArrayList<>(recordWebResources.size());
     //Convert the provider aggregations
```
jochen-vermeulen, line 85: It's important that these go first, right? So that they get first claim for web resources. Should we make a note of this?

metis-indexing/src/main/java/eu/europeana/indexing/fullbean/AggregationFieldInput.java
jochen-vermeulen: In fact, on further inspection, we never need the actual referenced web resources again, so `referencedWebResources` could even be a set of about values.

stzanakis: Okay so I don't see much of a benefit for maps but I changed the first one to a map to do some extra work, why not :D
We need to keep in mind now though that duplicate web resources with the same about would be discarded and only the first encountered will be kept. That was happening anyway automatically from mongo but now it will be "cleared" before hand.(In general that shouldn't be allowed at the first place anyway)
For the second I opted out with a set of strings instead.

metis-indexing/src/main/java/eu/europeana/indexing/fullbean/RdfToFullBeanConverter.java
stzanakis: yes sure I've added some short explanation

PR #448, opened 2021-05-04
Feat/met 3478 upgrade zoho library to 3 0 1

stzanakis:


metis-common/metis-common-zoho/src/main/java/eu/europeana/metis/zoho/ZohoAccessClient.java
```diff
@@ -100,20 +114,50 @@ public ZohoAccessClient(String grantToken, Map<String, String> zcrmConfiguration
...
+      RecordOperations recordOperations = new RecordOperations();
+      ParameterMap paramInstance = new ParameterMap();
+      paramInstance.add(SearchRecordsParam.CRITERIA, String
           .format(ZohoConstants.ZOHO_OPERATION_FORMAT_STRING, ZohoConstants.ACCOUNT_NAME_FIELD,
               ZohoConstants.EQUALS_OPERATION, organizationName));
-    } catch (ZCRMException e) {
+
+      APIResponse<ResponseHandler> response = recordOperations
+          .searchRecords(ZohoConstants.ACCOUNTS_MODULE_NAME, paramInstance);
+      return getZohoRecord(response);
```
jochen-vermeulen, line 128: It looks like multiple methods in this class are almost the same apart from the specific configuration of the paramInstance. Is there some generalization that we can do?

metis-common/metis-common-zoho/src/main/java/eu/europeana/metis/zoho/ZohoUtils.java
```diff
@@ -22,8 +26,28 @@ private ZohoUtils() {
...
+  /**
+   * Method that would check if the object provided is of type {@link JSONObject.Null} and will
+   * return a correct representation of {@link List} with {@link String} items or empty list.
+   *
+   * @param object the object to be checked
+   * @return the List of strings representation of the object or empty list
+   */
+  public static List<String> stringListSupplier(Object object) {
+    if (!JSONObject.NULL.equals(object) && object instanceof List<?> && ((List<?>) object)
+        .get(0) instanceof Choice<?>) {
```
jochen-vermeulen, line 47: Can this list be empty? Should we check that?

metis-authentication/metis-authentication-common/src/main/java/eu/europeana/metis/authentication/utils/MetisZohoOAuthPSQLHandler.java
```diff
@@ -70,60 +79,79 @@ public static void initializeWithRefreshToken(@Nullable String userEmailId,
...
+    LOGGER.warn("Returning null, this method is not supported");
+    return Collections.emptyList();
+  }
+
+  @Override
+  public void deleteTokens() {
+    //Nothing to do, not used method
+    LOGGER.warn("Did nothing, this method is not supported");
+  }
+
```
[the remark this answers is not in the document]
stzanakis: These are calls that are used from Zoho libs internally so I wouldn't throw an exception here cause I don't know how it would behaved..

metis-common/metis-common-zoho/src/main/java/eu/europeana/metis/zoho/ZohoUtils.java
stzanakis: Yes it's possible, I've added a check

PR #449, opened 2021-05-17
MET-3659: Include header in result of OAI GetRecord request.

jochen-vermeulen:


metis-harvesting/src/main/java/eu/europeana/metis/harvesting/oaipmh/OaiRecord.java
```diff
@@ -0,0 +1,42 @@
...
+
+  /**
+   * Constructor.
+   *
+   * @param header The OAI header of the record.
+   * @param record The record itself (the embedded record).
+   */
+  public OaiRecord(OaiRecordHeader header, byte[] record) {
+    this.header = header;
+    this.record = record;
```
stzanakis, line 23: Sonar says to store a copy of the array.
We've decided to try and store and retrieve copies of such items.

metis-harvesting/src/main/java/eu/europeana/metis/harvesting/oaipmh/OaiRecord.java
```diff
@@ -0,0 +1,42 @@
...
+    return header;
+  }
+
+  /**
+   * Makes the embedded record available.
+   *
+   * @return An input stream containing the record. The caller needs to close it after use.
+   * @throws HarvesterException In case the record is marked as deleted.
+   */
+  public InputStream getRecord() throws HarvesterException {
```
stzanakis, line 36: I was thinking, is there a way to protect the system by somehow auto closing the InputStream if the caller didn't. 
Even in the tests they are not closed I think, not that it matters that much..
E.g. holding a map of items for the class and close them at the end or something like that? Or is it an overkill, what do you think?

metis-harvesting/src/main/java/eu/europeana/metis/harvesting/oaipmh/OaiRecordParser.java
```diff
@@ -79,58 +81,48 @@ private static void initializeExpressions() throws HarvesterException {
    * @param oaiRecord The record to parse as a string value.
    * @throws HarvesterException In case there was a problem with setting up the parser.
    */
-  OaiRecordParser(String oaiRecord) throws HarvesterException {
+  OaiRecordParser(byte[] oaiRecord) throws HarvesterException {
```
stzanakis, line 84: The oaiRecord shouldn't be part of the class or constructor.
It should be used/passed with the methods further down, right?
Also the XPathExpression doesn't have to be computed for every record, just once. Computing them each time I believe has a significant impact on the performance.

metis-harvesting/src/main/java/eu/europeana/metis/harvesting/oaipmh/OaiRecordParser.java
```diff
@@ -79,58 +81,48 @@ private static void initializeExpressions() throws HarvesterException {
...
-   * @throws HarvesterException in case there is a problem with the expression.
-   */
-  boolean recordIsDeleted() throws HarvesterException {
-    try {
-      // Note that this is created safely, so this is a false positive by SonarQube.
-      @SuppressWarnings("findsecbugs:XXE_XPATH")
-      final String result = isDeletedExpression.evaluate(getInputSource());
-      return "deleted".equalsIgnoreCase(result);
-    } catch (XPathExpressionException e) {
+      final byte[] headerBytes = getAsBytes(headerExpression);
```
stzanakis, line 97: If the constructor is updated to use the same static XPathExpression for all records I think those XPathExpression should be used with synchronization to avoid weird issues we had before with those classes(I think it depends on the  exact implementation of the library that is used, but we might want to stay on the safe side).

metis-harvesting/src/main/java/eu/europeana/metis/harvesting/oaipmh/OaiRecordParser.java
```diff
@@ -45,26 +48,25 @@
           "/*[local-name()='metadata']" +
           "/child::*";
 
-  private static final String IS_DELETED_XPATH = "string(/*[local-name()='OAI-PMH']" +
+  private static final String HEADER_XPATH = "/*[local-name()='OAI-PMH']" +
           "/*[local-name()='GetRecord']" +
           "/*[local-name()='record']" +
-          "/*[local-name()='header']" +
-          "/@status)";
+          "/*[local-name()='header']";
```
stzanakis, line 54: Should we compute the constant in one go, removing the "+"?
(Same for the to other xpath)

metis-harvesting/src/main/java/eu/europeana/metis/harvesting/oaipmh/OaiRecordParser.java
jochen-vermeulen: OK. Done.

metis-harvesting/src/main/java/eu/europeana/metis/harvesting/oaipmh/OaiRecord.java
stzanakis: Indeed the byte stream is safer! I guess we can stick to the javadoc notification

metis-harvesting/src/main/java/eu/europeana/metis/harvesting/oaipmh/OaiRecordParser.java
stzanakis: True, but at least we avoid that extra call to the initialization

metis-harvesting/src/main/java/eu/europeana/metis/harvesting/oaipmh/OaiRecordParser.java
stzanakis: Okay makes sense

PR #451, opened 2021-05-25
Feat/met 2215 add incremental parameter for ecloud

jochen-vermeulen:


metis-core/metis-core-common/src/main/java/eu/europeana/metis/core/workflow/plugins/AbstractExecutablePlugin.java
```diff
@@ -122,13 +122,20 @@ DpsTask createDpsTaskForHarvestPlugin(EcloudBasePluginParameters ecloudBasePlugi
     if (extraParameters != null) {
       parameters.putAll(extraParameters);
     }
-    parameters.put("PROVIDER_ID", ecloudBasePluginParameters.getEcloudProvider());
-    parameters.put("OUTPUT_DATA_SETS", String
+
+    if (incrementalProcessing) {
```
stzanakis, line 126: Why the check, can't we just set the value always?

jochen-vermeulen: It's just that I don't know if eCloud is already expecting the parameter for http harvesting and, if it is not, if it will throw an error for an unexpected parameter.

PR #454, opened 2021-06-02
MET-3690: Technical debt

JoanaCMS:


metis-authentication/metis-authentication-common/src/main/java/eu/europeana/metis/authentication/utils/MetisZohoOAuthPSQLHandler.java
```diff
@@ -95,6 +94,7 @@ public Token getToken(UserSignature userSignature, Token token) {
         }
         final MetisZohoOAuthToken foundToken = (MetisZohoOAuthToken) query.list().get(0);
         String potentialErrorMessage = "Exception while retrieving zoho oauth user tokens";
+        Transaction tx = session.beginTransaction();
```
jochen-vermeulen, line 97: This is quite wrong! You should be starting the transaction before creating and executing the query. 

Sonarqube probably told you to declare the variable closer to where it is used. But this is in fact a very valid exception to that rule!

PR #456, opened 2021-06-14
MET-3671 Organization indexing support

stzanakis:
PR build will fail build since it's dependent on the corelib [pull request](https://github.com/europeana/corelib/pull/237)
Once corelib is merged, the build here could be re-triggered for verification

### DIFF
--- a/metis-indexing/src/main/java/eu/europeana/indexing/mongo/FullBeanUpdater.java
+++ b/metis-indexing/src/main/java/eu/europeana/indexing/mongo/FullBeanUpdater.java
@@ -96,7 +96,7 @@ public class FullBeanUpdater extends AbstractMongoObjectUpdater<FullBeanImpl, Vo
     propertyUpdater.updateReferencedEntities("aggregations", FullBeanImpl::getAggregations,
         FullBeanUpdater::createRootAbout, new AggregationUpdater());
     propertyUpdater.updateReferencedEntities("organizations", FullBeanImpl::getOrganizations,
-        FullBeanUpdater::createRootAbout, new OrganizationUpdater());
+        fullBean -> null, new OrganizationUpdater());
     propertyUpdater.updateReferencedEntity("europeanaAggregation",
         FullBeanUpdater::getEuropeanaAggregationFromFullBean, FullBeanUpdater::createRootAbout,
         new EuropeanaAggregationUpdater());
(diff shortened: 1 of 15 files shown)


stzanakis:
PR will fail build since it's dependent on the corelib [pull request](https://github.com/europeana/corelib/pull/237)

metis-indexing/src/main/java/eu/europeana/indexing/solr/property/AggregationSolrCreator.java
```diff
@@ -15,38 +26,113 @@
...
+    return new ImmutablePair<>(uriList, literalsMap);
+  }
+
+  private void extractUris(Map<String, List<String>> urisLiteralsMap,
+      List<String> uriList, Map<String, List<String>> literalsMap) {
+    for (Map.Entry<String, List<String>> entry : urisLiteralsMap.entrySet()) {
+      final List<String> literals = new ArrayList<>();
+      for (String value : entry.getValue()) {
+        if (organizationMap.containsKey(value)) {
+          uriList.add(value);
```
jochen-vermeulen, line 114: So we 'lose' the key (i.e. the language) of these links? There is no danger with this?

metis-indexing/src/main/java/eu/europeana/indexing/mongo/FullBeanUpdater.java
```diff
@@ -95,6 +95,8 @@ protected void update(MongoPropertyUpdater<FullBeanImpl> propertyUpdater,
         fullBean -> null, new ProvidedChoUpdater());
     propertyUpdater.updateReferencedEntities("aggregations", FullBeanImpl::getAggregations,
         FullBeanUpdater::createRootAbout, new AggregationUpdater());
+    propertyUpdater.updateReferencedEntities("organizations", FullBeanImpl::getOrganizations,
+        FullBeanUpdater::createRootAbout, new OrganizationUpdater());
```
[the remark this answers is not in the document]
stzanakis: Indeed, good catch!

metis-indexing/src/main/java/eu/europeana/indexing/solr/property/AggregationSolrCreator.java
stzanakis: Correct! But we don't really lose anything, since the links shouldn't have a language in the first place. (well with the conversion they are getting the "def" language which is then cleared if it's empty..
So I don't think there is danger in that perspective.

PR #458, opened 2021-06-22
Fix xml:lang in rdf:resource refs

hugomanguinhas:
This is a fix to make the XSLT more resilient to cases when both the xml:lang and rdf:resource are present. This issue happens on the JHN dataset.


stzanakis:
This has been here for a while, I'm assuming it's irrelevant now since there has been any action on this.
Closing.

PR #460, opened 2021-06-30
MET-3707: Manage New Task state for Post Processing

JoanaCMS:


metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/execution/WorkflowExecutor.java
```diff
@@ -345,6 +346,11 @@ private void periodicCheckingLoop(long sleepTime, AbstractExecutablePlugin plugi
         plugin.setPluginStatusAndResetFailMessage(
             monitorResult.getTaskState() == TaskState.REMOVING_FROM_SOLR_AND_MONGO
                 ? PluginStatus.CLEANING : PluginStatus.RUNNING);
+
+        plugin.setPluginStatusAndResetFailMessage(
+            monitorResult.getTaskState() == TaskState.IN_POST_PROCESSING
+                ? PluginStatus.IDENTIFYING_DELETED_RECORDS : PluginStatus.RUNNING);
```
jochen-vermeulen, line 352: I think it is better to integrate these two statements so that you don't first set it to running and then to identifying deleted records. This should make the code more readable.

jochen-vermeulen: Also, have you verified with PSNC which task types may get this TaskState value? The description 'identifying deleted records' applies only to the two harvesting task types. If they think it is (or will be) also applicable to other task types, we should distinguish by plugin type here.

JoanaCMS: I will check this with PSNC and apply the changes

PR #464, opened 2021-07-15
Feat/met 3748 metis repository first version

jochen-vermeulen:


metis-repository/src/main/java/eu/europeana/metis/repository/rest/RecordController.java
```diff
@@ -0,0 +1,64 @@
...
+   * It creates a record to be put into the database
+   *
+   * @param recordId - A unique record id
+   * @param datasetId - The id of the dataset which the record belongs to
+   * @param dateStamp - Last time the record was updated. It can also be the date of creation
+   * @param edmRecord - The record itself
+   * @return The record just saved in the database
+   */
+  @PostMapping(value = RestEndpoints.ADD_RECORD_TO_DATABASE, consumes = {
+      MediaType.APPLICATION_JSON_VALUE, MediaType.APPLICATION_XML_VALUE}, produces = {
```
jochen-vermeulen, line 51: Should it consume json? We only support XML records.

metis-repository/src/main/java/eu/europeana/metis/repository/dao/RecordDao.java
```diff
@@ -0,0 +1,61 @@
...
+  /**
+   * Create record in the database
+   *
+   * @param record - The record to be saved in the database
+   * @return The record just saved in the database
+   */
+  public Record createRecord(Record record) {
+
+    Optional<Record> recordFound = datastore.find(Record.class).stream()
+        .filter(x -> x.getRecordId().equals(record.getRecordId())).findFirst();
```
jochen-vermeulen, line 47: This query is not really efficient! This retrieves all objects (downloads all) and then locally find the one you need. Much better to perform a filter on the query before calling stream(), that way the filter will be performed on the database server side.

metis-repository/src/main/java/eu/europeana/metis/repository/dao/RecordDao.java
```diff
@@ -0,0 +1,61 @@
...
+
+    if(recordFound.isPresent()) {
+      record.setId(recordFound.get().getId());
+    } else {
+      final ObjectId objectId = Optional.ofNullable(record.getId()).orElseGet(ObjectId::new);
+      record.setId(objectId);
```
jochen-vermeulen, line 53: I think if you create a new object better to not create a new one: the database will take care of it.

metis-repository/src/main/java/eu/europeana/metis/repository/rest/RecordController.java
JoanaCMS: That's a good point, I will remove the JSON consumption

metis-repository/src/main/java/eu/europeana/metis/repository/rest/RecordController.java
```diff
@@ -0,0 +1,64 @@
...
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RequestParam;
+import org.springframework.web.bind.annotation.ResponseBody;
+import org.springframework.web.bind.annotation.RestController;
+
+/**
+ * Controller for record management.
+ */
+@RestController
+@RequestMapping("records")
```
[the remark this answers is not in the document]
JoanaCMS: I'll remove this then

metis-repository/src/main/java/eu/europeana/metis/repository/dao/RecordDao.java
JoanaCMS: I'll change the query then

metis-repository/src/main/java/eu/europeana/metis/repository/dao/RecordDao.java
JoanaCMS: I'll remove these lines of code then

metis-repository/src/main/java/eu/europeana/metis/repository/rest/RecordController.java
jochen-vermeulen: I think you actually removed XML consumption by accident?

metis-repository/src/main/java/eu/europeana/metis/repository/rest/RecordController.java
```diff
@@ -0,0 +1,63 @@
...
+import io.swagger.annotations.ApiResponse;
+import io.swagger.annotations.ApiResponses;
+import io.swagger.v3.oas.annotations.tags.Tag;
+import io.swagger.v3.oas.annotations.tags.Tags;
+import java.util.Date;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.http.MediaType;
+import org.springframework.web.bind.annotation.PostMapping;
+import org.springframework.web.bind.annotation.RequestBody;
+import org.springframework.web.bind.annotation.RequestMapping;
```
jochen-vermeulen, line 17: I think there is an unused import here.

metis-repository/src/main/java/eu/europeana/metis/repository/dao/RecordDao.java
```diff
@@ -0,0 +1,73 @@
...
+import dev.morphia.Morphia;
+import dev.morphia.mapping.DiscriminatorFunction;
+import dev.morphia.mapping.MapperOptions;
+import dev.morphia.mapping.NamingStrategy;
+import dev.morphia.query.experimental.filters.Filters;
+import eu.europeana.metis.network.ExternalRequestUtil;
+import java.util.List;
+import java.util.Optional;
+import java.util.stream.Collectors;
+import org.bson.types.ObjectId;
```
jochen-vermeulen, line 14: Looks like an unused import.

metis-repository/src/main/java/eu/europeana/metis/repository/rest/HttpHarvestController.java
```diff
@@ -0,0 +1,86 @@
...
+    ByteArrayOutputStream byteArrayOutputStream = new ByteArrayOutputStream();
+    BufferedOutputStream bufferedOutputStream = new BufferedOutputStream(byteArrayOutputStream);
+    ZipOutputStream zipOutputStream = new ZipOutputStream(bufferedOutputStream);
+
+    for (Record record : recordList) {
+      zipOutputStream.putNextEntry(new ZipEntry(record.getRecordId() + ".xml"));
+      ObjectOutputStream objectOutputStream = new ObjectOutputStream(zipOutputStream);
+      objectOutputStream.writeObject(record);
+      objectOutputStream.flush();
+      objectOutputStream.close();
```
jochen-vermeulen, line 74: We shouldn't write the whole object to the output stream. Metis is not going to know what to do with this. Only the actual EDM record should be zipped (so the one property of the Record object). It is a string, so no need to use an object stream.

metis-repository/src/main/java/eu/europeana/metis/repository/rest/HttpHarvestController.java
```diff
@@ -0,0 +1,86 @@
...
+      zipOutputStream.closeEntry();
+    }
+
+    zipOutputStream.finish();
+    zipOutputStream.flush();
+    IOUtils.closeQuietly(zipOutputStream);
+    IOUtils.closeQuietly(bufferedOutputStream);
+    IOUtils.closeQuietly(byteArrayOutputStream);
```
jochen-vermeulen, line 82: I think if there is a problem, it should report it (in the form of a 500 status code). So that the receiver knows that there is a problem.

metis-repository/src/main/java/eu/europeana/metis/repository/dao/RecordDao.java
```diff
@@ -0,0 +1,73 @@
...
+  /**
+   * Creates  list of records that belong to the given dataset id
+   *
+   * @param datasetId - The id of the dataset the records belong to
+   * @return A List of records that are part of the dataset with datasetId
+   */
+  public List<Record> getAllRecordsFromDataset(String datasetId) {
+    return datastore.find(Record.class).filter(Filters.eq(DATASET_ID_FIELD, datasetId)).stream()
+        .collect(Collectors.toList());
+  }
```
jochen-vermeulen, line 72: I think it is better to return a stream here, instead of loading everything in memory first.

metis-repository/src/main/java/eu/europeana/metis/repository/rest/HttpHarvestController.java
```diff
@@ -0,0 +1,86 @@
...
+ */
+@RestController
+@RequestMapping("http")
+@Tags(@Tag(name = HttpHarvestController.CONTROLLER_TAG_NAME,
+    description = "Controller providing access to HTTP (zip) harvesting functionality."))
+@Api(tags = HttpHarvestController.CONTROLLER_TAG_NAME)
+public class HttpHarvestController {
+
+  public static final String CONTROLLER_TAG_NAME = "HttpHarvestController";
+  private static final Logger LOGGER = LoggerFactory.getLogger(HttpHarvestController.class);
```
jochen-vermeulen, line 41: LOGGER object is unused (although maybe you will start using it when you resolve the review comments).

metis-repository/src/main/java/eu/europeana/metis/repository/rest/HttpHarvestController.java
```diff
@@ -0,0 +1,86 @@
...
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.ResponseBody;
+import org.springframework.web.bind.annotation.ResponseStatus;
+import org.springframework.web.bind.annotation.RestController;
+
+/**
+ * Controller for HTTP harvesting.
+ */
+@RestController
+@RequestMapping("http")
```
jochen-vermeulen, line 34: Also this line should be removed because you already define the full mapping below (in the individual request(s)).

metis-repository/src/main/java/eu/europeana/metis/repository/rest/RecordController.java
JoanaCMS: Removed

metis-repository/src/main/java/eu/europeana/metis/repository/dao/RecordDao.java
JoanaCMS: Removed

metis-repository/src/main/java/eu/europeana/metis/repository/rest/HttpHarvestController.java
JoanaCMS: Removed

metis-repository/src/main/java/eu/europeana/metis/repository/rest/HttpHarvestController.java
JoanaCMS: Yes, I just changed so that it is used somewhere in the code

metis-repository/src/main/java/eu/europeana/metis/repository/rest/HttpHarvestController.java
JoanaCMS: I changed the close and added a try catch block

metis-repository/src/main/java/eu/europeana/metis/repository/dao/RecordDao.java
JoanaCMS: Also works, I just changed the method

metis-repository/src/main/java/eu/europeana/metis/repository/rest/RecordController.java
JoanaCMS: Oops, corrected that

metis-repository/src/main/java/eu/europeana/metis/repository/rest/HttpHarvestController.java
jochen-vermeulen: I ended up making some changes to the code. Making sure we use spring code only (no underlying java response object), returning a 500 code in case of issues, and using the try with closable scheme for the zip output stream. Have a look when you're back.

metis-repository/src/main/java/eu/europeana/metis/repository/rest/OaiPmhController.java
```diff
@@ -0,0 +1,140 @@
...
+    } else {
+      throw new ResponseStatusException(HttpStatus.BAD_REQUEST, "Unsupported verb: " + verb);
+    }
+
+    // Compile the result
+    final OAIPMH result = new OAIPMH().withVerb(verbResult).withResponseDate(new Date())
+            .withRequest(new Request(RestEndpoints.OAI_ENDPOINT).withVerbType(verbResult.getType()));
+    try {
+      ByteArrayOutputStream outputStream = new ByteArrayOutputStream();
+      final WriterContext context = new WriterContext(Granularity.Day, new SimpleResumptionTokenFormat());
```
JoanaCMS, line 94: I'm not familiar with `WriterContext`, what's its purpose in here?

jochen-vermeulen: It's an OAI specific term. I'm really only using it to specify that we want to send the datestamps as dates only (and not include a time component). I think that will be more beneficial to our testing.

metis-harvesting/src/main/java/eu/europeana/metis/harvesting/http/HttpHarvesterImpl.java
```diff
@@ -34,6 +42,55 @@
 
   private static final Logger LOGGER = LoggerFactory.getLogger(HttpHarvesterImpl.class);
 
+  @Override
+  public void harvestRecords(InputStream inputStream, CompressedFileExtension compressedFileType,
+          Consumer<ArchiveEntry> action) throws HarvesterException {
```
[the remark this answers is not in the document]
jochen-vermeulen: The existing harvesting functionality works on a temporary file, not on an input stream. It is due to the fact that the zip file might be large. So the idea is to first save it to the file system and then we can do all operations on it without risking memory issues. Of course for our test application this is not really a concern, but it is easiest to just do this so that we can use the functionality already implemented below. And you never know...

PR #468, opened 2021-07-27
#MET-3780 fixed pagination for ZohoAccessClient

gsergiu:
the fix for pagination when retrieving zoho organizations was implemented

### DIFF
--- a/metis-common/metis-common-zoho/src/main/java/eu/europeana/metis/zoho/ZohoAccessClient.java
+++ b/metis-common/metis-common-zoho/src/main/java/eu/europeana/metis/zoho/ZohoAccessClient.java
@@ -247,14 +247,14 @@ public class ZohoAccessClient {
           new IllegalArgumentException(
               String.format("Provided page: %s, and pageSize: %s", page, pageSize)));
     }
-    int start = ((page - 1) * pageSize) + 1;
+//    int start = ((page - 1) * pageSize) + 1;
 
     try {
       APIResponse<ResponseHandler> response;
       RecordOperations recordOperations = new RecordOperations();
       ParameterMap paramInstance = new ParameterMap();
       if (isNullOrEmpty(searchCriteria)) {//No searchCriteria available
-        paramInstance.add(GetRecordsParam.PAGE, start);
+        paramInstance.add(GetRecordsParam.PAGE, page);
         paramInstance.add(GetRecordsParam.PER_PAGE, pageSize);
         HeaderMap headerInstance = new HeaderMap();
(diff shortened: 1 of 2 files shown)


metis-common/metis-common-zoho/src/main/java/eu/europeana/metis/zoho/ZohoAccessClient.java
```diff
@@ -247,31 +248,40 @@ public ZohoAccessClient(TokenStore tokenStore, String zohoEmail, String clientId
...
         paramInstance.add(SearchRecordsParam.PER_PAGE, pageSize);
         paramInstance.add(SearchRecordsParam.CRITERIA,
             createZohoCriteriaString(searchCriteria, criteriaOperator));
+        
         response = recordOperations
             .searchRecords(ZohoConstants.ACCOUNTS_MODULE_NAME, paramInstance);
       }
-      return getZohoRecords(response);
+      
+      if(response.getStatusCode() == 200) {
```
stzanakis, line 275: Is this if else block needed?
Doesn't the getZohoRecords method cover the 204 code and all other codes with 
eu/europeana/metis/zoho/ZohoAccessClient.java:141?

metis-enrichment/metis-enrichment-service/src/main/java/eu/europeana/enrichment/service/Converter.java
```diff
@@ -233,9 +238,33 @@ static EnrichmentTerm organizationImplToEnrichmentTerm(
     enrichmentTerm.setCreated(Objects.requireNonNullElseGet(created, Date::new));
     enrichmentTerm.setUpdated(updated);
 
+    //generate LabelInfos
+    final Map<String, List<String>> combinedLabels = new HashMap<>();
+    copyToCombinedLabels(combinedLabels, organizationEnrichmentEntity.getPrefLabel());
+    copyToCombinedLabels(combinedLabels, organizationEnrichmentEntity.getAltLabel());
+    copyToCombinedLabels(combinedLabels, organizationEnrichmentEntity.getEdmAcronym());
```
stzanakis, line 245: These two blocks seem to be a copy from the tool that we have to populate enrichment but with the addition of EdmAcronym. Is this field a particularity only for organization entities?

stzanakis:
@gsergiu On your absence, I've updated the code to reflect my thoughts.
When you check this I would probably be on a leave, so please check if everything is fine as it is now and perhaps coordinate the merge with someone from the metis team.
(This is not an urgent merge)

PR #471, opened 2021-08-27
MET-2214: Add new counters to execution progress.

jochen-vermeulen:


metis-core/metis-core-common/src/main/java/eu/europeana/metis/core/workflow/plugins/ExecutionProgress.java
```diff
@@ -10,28 +9,29 @@
  * @author Simon Tzanakis (Simon.Tzanakis@europeana.eu)
  * @since 2017-06-01
  */
-@Embedded
+@Entity
```
stzanakis, line 12: Reverted this, to take care of the deprecation on all objects at once with another ticket.

PR #473, opened 2021-09-03
MET-3841 Counters calculation fix

stzanakis:


metis-core/metis-core-common/src/main/java/eu/europeana/metis/core/workflow/plugins/ExecutionProgress.java
```diff
@@ -91,6 +91,6 @@ public void setStatus(TaskState status) {
 
   public void recalculateProgressPercentage() {
     this.progressPercentage = this.expectedRecords == 0 ? 0
-        : Math.round(100f * this.processedRecords / this.expectedRecords);
+        : Math.round(100f * (this.processedRecords + this.deletedRecords)/ (this.expectedRecords + this.deletedRecords));
```
jochen-vermeulen, line 94: This could cause confusion if there are large numbers of deleted records, right? The user could see something like 1 record done out of 100, and the bar is at 95%.

metis-core/metis-core-common/src/main/java/eu/europeana/metis/core/workflow/plugins/AbstractExecutablePlugin.java
```diff
@@ -250,21 +228,46 @@ public MonitorResult monitor(DpsClient dpsClient) throws ExternalTaskException {
...
+      //ignoredRecordsCount never used
+      //expectedPostProcessedRecordsNumber and postProcessedRecordsCount represent deleted records
+      //The deletedRecordsCount is always 0
+      expectedRecordCount = taskInfo.getExpectedRecordsNumber();
+      processedRecordCount = taskInfo.getProcessedRecordsCount();
+      deletedRecordCount = taskInfo.getPostProcessedRecordsCount();
+    } else {
+      //Other plugins including incremental indexing
+      //expectedPostProcessedRecordsNumber, postProcessedRecordsCount and ignoredRecordsCount not used
+      expectedRecordCount = taskInfo.getExpectedRecordsNumber() - taskInfo.getDeletedRecordsCount();
```
jochen-vermeulen, line 263: I think the problem here is that the expected records count decreases gradually as the task progresses and it finds more and more deleted records. Are we sure we want this? To do this properly, we would need the number of expected deleted records.

stzanakis: That is indeed true for all executions now(It used to be just for oai).
We discussed this and the solution would be another counter, but for now we decided to keep it this way(since it was somewhat possible for oai already) and see what the user experience would be.
Unless we decide otherwise.

metis-core/metis-core-common/src/main/java/eu/europeana/metis/core/workflow/plugins/ExecutionProgress.java
stzanakis: I see the point.
The way I thought of this is that if there are many deleted records being "processed" the task will seem stuck in the progress bar until those deleted go through.
I'm fine with either, it's more a matter of display, let me know what you think.

jochen-vermeulen: Ok, maybe we should leave it like this. I think we may wish to eventually get the expected deleted records count and solve both the problems/comments at once.

PR #479, opened 2021-10-13
met-3953

adolfopeixinho:
Fix for file paths using slash/, replace with File.separator.
Change for access method from private to public


JoanaCMS:
This PR has been reviewed. Will merge once the release is done

PR #484, opened 2021-11-11
MET-4004 Removed any references to commons-lang library

JoanaCMS:


stzanakis:
I see one StringUtils(non apache) non-used import in `src/test/java/eu/europeana/enrichment/utils/ItemExtractorUtilsTest.java`.
Are there others like that?

Also since the `commons-lang` is not used anymore and we are expected to use the `commons-lang3` perhaps we can go the extra mile here and remove the `commons-lang` transitive dependency from the dependencies that contain it?

PR #485, opened 2021-11-12
Feat/met 4007 handle total records indexin plugin

stzanakis:


metis-core/metis-core-common/src/main/java/eu/europeana/metis/core/workflow/plugins/AbstractIndexPlugin.java
```diff
@@ -0,0 +1,77 @@
...
+
+  @Override
+  public DpsTask prepareDpsTask(String datasetId, EcloudBasePluginParameters ecloudBasePluginParameters) {
+    return createDpsTaskForIndexPlugin(ecloudBasePluginParameters, datasetId,
+        getPluginMetadata().isIncrementalIndexing(),
+        getPluginMetadata().getHarvestDate(),
+        getPluginMetadata().isUseAlternativeIndexingEnvironment(),
+        getPluginMetadata().isPreserveTimestamps(),
+        getPluginMetadata().getDatasetIdsToRedirectFrom(),
+        getPluginMetadata().isPerformRedirects(), getTargetIndexingDatabase().name());
```
JoanaCMS, line 50: I don't know where the method `createDpsTaskForIndexPlugin` comes from, but I see a lot of the input variables come from the `getPluginMetadata`, wouldn't be easier to just use a `pluginMetadata` as one of the inputs?

stzanakis: The `createDpsTaskForIndexPlugin` comes from the `AbstractExecutablePlugin` which is higher in the hierarchy.
It makes sense what you are asking so I gave it some thought before replying.
Using a `pluginMetadata` is possible as long as it's a specific type. So to pass it further we'd need to cast it to the right type either when calling the method or inside it.
Where it is now, it's basically transparent and automatically recognized.
Also the `createDpsTaskForIndexPlugin` is there for other potential uses(even though we are not using it) without having an explicit plugin required.
What I don't like is that the higher class need to explicitly know the below class..
What we could do though is something similar to `TransformationPlugin`.
Having said that, we might need more time to think about it, since we should have a look at the other plugins at the same time.
So I'll leave it as is for now.

JoanaCMS: I see your point, makes sense for the higher class to not know explicitly the below class. 
The approach done in `TransformationPlugin` seems to be the more feasible solution for this.
Perhaps we could create a Technical Debt ticket to tackle this?

stzanakis: Sure, we can work with investigative td ticket.

PR #487, opened 2021-12-07
MET-4065 Added different constructor to support metis-sandbox functio…

adolfopeixinho:
…nality. Todo in pom.xml with missing dependency. Added better gitignore


stzanakis:
This branch has been squash merged.

PR #493, opened 2022-01-10
Feat/met 4156 tier calculation overview

stzanakis:


metis-indexing/src/main/java/eu/europeana/indexing/tiers/view/LanguageBreakdown.java
```diff
@@ -0,0 +1,64 @@
+package eu.europeana.indexing.tiers.view;
+
+import java.util.List;
+
+public class LanguageBreakdown {
+
+  private int potentialLanguageQualifiedElements;
+  private int actualLanguageQualifiedElements;
+  private int actualLanguageQualifiedElementsPercentage;
```
JoanaCMS, line 9: As a percentage wouldn't we want this as a double instead of int?

metis-indexing/src/main/java/eu/europeana/indexing/tiers/view/LanguageBreakdown.java
```diff
@@ -0,0 +1,64 @@
...
+  private int actualLanguageUnqualifiedElements;
+  private List<String> actualLanguageUnqualifiedElementsList;
+  private String metadataTier;
```
JoanaCMS, line 12: Same thing as above

metis-indexing/src/main/java/eu/europeana/indexing/tiers/view/EnablingElements.java
```diff
@@ -0,0 +1,55 @@
...
+
+import java.util.List;
+
+public class EnablingElements {
+
+  private int distinctEnablingElements;
+  private List<String> distinctEnablingElementsList;
+  private int metadataGroups;
+  private List<String> metadataGroupsList;
+  private String metadataTier;
```
JoanaCMS, line 11: Same as above

metis-indexing/src/main/java/eu/europeana/indexing/tiers/view/ContextualClasses.java
```diff
@@ -0,0 +1,46 @@
+package eu.europeana.indexing.tiers.view;
+
+import java.util.List;
+
+public class ContextualClasses {
+
+  private int completeContextualResources;
+  private int distinctClassesOfCompleteContextualResources;
+  private List<String> distinctClassesList;
+  private String metadataTier;
```
JoanaCMS, line 10: Maybe instead of using a String we could use the values from an enum, like the ones being used in `SolrTier` (`RdfTier` and `EdmLabel`),  it would be one way to ensure there are no errors when writing the value as a String

stzanakis: Not possible because:
- This is a view class and the tier class will be serialized as a whole, which is not what we want
- Also the "0" classification cannot be exactly represented in java because fields cannot be named as just a number.

It could be possible to do but I would need to merge the tiers somehow and write a custom (de)serializer I thnk.

stzanakis: I think the closest we can get is `MediaTier` and `MetadataTier`, but still a (de)serializer would probably be required.
I'll give it some thought.

metis-indexing/src/main/java/eu/europeana/indexing/tiers/view/LanguageBreakdown.java
stzanakis: Not sure how useful the decimal fractions would be in this case but I could update this to float, perhaps not double.

metis-indexing/src/main/java/eu/europeana/indexing/tiers/view/MediaResourceTechnicalMetadata.java
```diff
@@ -0,0 +1,64 @@
...
+
+  public String getVerticalResolution() {
+    return verticalResolution;
+  }
+
+  public void setVerticalResolution(String verticalResolution) {
+    this.verticalResolution = verticalResolution;
+  }
+
+  public LicenseType getLicenceType() {
```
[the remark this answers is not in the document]
stzanakis: Indeed, thanks!

metis-indexing/src/main/java/eu/europeana/indexing/tiers/view/ContextualClasses.java
stzanakis: I've managed to make something nice out of this.

metis-indexing/src/main/java/eu/europeana/indexing/tiers/view/EnablingElements.java
stzanakis: Same as https://github.com/europeana/metis-framework/pull/493#discussion_r781343224

metis-indexing/src/main/java/eu/europeana/indexing/tiers/RecordTierCalculationViewGenerator.java
```diff
@@ -0,0 +1,91 @@
...
+import java.util.List;
+import java.util.stream.Collectors;
+
+public class RecordTierCalculationViewGenerator {
+
+  private final String europeanaId;
+  private final String providerId;
+  private final String stringRdf;
+
+  public RecordTierCalculationViewGenerator(String europeanaId, String providerId, String stringRdf) {
```
jeortizquan: I think these parameters can be final ```RecordTierCalculationViewGenerator(final String europeanaId, final String providerId, final String stringRdf)``` since they are not being changed in the class.

metis-indexing/src/main/java/eu/europeana/indexing/tiers/RecordTierCalculationViewGenerator.java
```diff
@@ -0,0 +1,91 @@
...
+    recordTierCalculationSummary.setProviderRecordId(providerId);
+    recordTierCalculationSummary.setContentTier(tierValues.getContentTier());
+    recordTierCalculationSummary.setMetadataTier(tierValues.getMetadataTier());
+    recordTierCalculationSummary.setPortalLink("https://example.com"); // We need the root link from the configuration
+    recordTierCalculationSummary.setHarvestedRecordLink("https://example.com"); // We need a proxy controller for this
+
+    return recordTierCalculationView;
+  }
+
+  private TierValues calculateTierValues(String xml) {
```
jeortizquan: same here. ```final``` parameters.

metis-indexing/src/main/java/eu/europeana/indexing/tiers/RecordTierCalculationViewGenerator.java
```diff
@@ -0,0 +1,91 @@
...
+      throw new TierCalculationException("Error during calculation of tiers", e);
+    }
+  }
+
+  static class TierValues {
+
+    private final String contentTier;
+    private final String metadataTier;
+
+    public TierValues(String contentTier, String metadataTier) {
```
jeortizquan: same here. ```final``` parameters.

metis-indexing/src/main/java/eu/europeana/indexing/tiers/RecordTierCalculationViewGenerator.java
stzanakis: Although in favor of the final keyword, I try not to over use it because I find it pollutes the code and it becomes less readable especially for simple methods where it's clear what they are doing..
Furthermore, we as developers should expect(in almost all cases) that a parameter won't change inside a method, unless explicitly defined that it can change.
In fact I'm curious why java doesn't have all parameters final by default and the developer only define the ones that wants them changeable(so the reversion of what is available now)..

Talking about this specific case I won't add the final since this is a simple constructor.

metis-indexing/src/main/java/eu/europeana/indexing/tiers/RecordTierCalculationViewGenerator.java
stzanakis: Check https://github.com/europeana/metis-framework/pull/493#discussion_r781886073
Here it makes a bit more sense to add it which I will do, but still not hugely impact-full since it's quite clear what's going on I think(obviously always depends on interpretation :smile: ).

metis-indexing/src/main/java/eu/europeana/indexing/tiers/RecordTierCalculationViewGenerator.java
stzanakis: Not adding it here.
Check https://github.com/europeana/metis-framework/pull/493#discussion_r781886073

metis-indexing/src/main/java/eu/europeana/indexing/tiers/RecordTierCalculationViewGenerator.java
jeortizquan: Yep, I remember in a previous experience I had, was sonarqube rule in place in favor / promote inmutability (functional programming) whenever was possible. The advantage of using ```final``` is that it reduces the mental load that the value / reference is never altered later on. So developers need not pay attention to modifications in order to understand the computation. But also understand your point in the sense that it will pollute the code. It something to keep in balance.

metis-indexing/src/main/java/eu/europeana/indexing/tiers/RecordTierCalculationViewGenerator.java
jeortizquan: in this case if it helps to keep inmutability. is fine.

PR #499, opened 2022-01-21
Feat/met 4158 media tier breakdown

stzanakis:


metis-schema/src/main/java/eu/europeana/metis/schema/model/MediaType.java
```diff
@@ -12,50 +13,67 @@
...
-   * Obtains the media type of a given mime type. This method accepts media types with subtypes
-   * and/or parameters.
+   * Obtains the media type of a given mime type. This method accepts mime types with subtypes and/or parameters.
    *
-   * @param mediaType The media type.
+   * @param mimeType The media type.
    * @return The media type to which the mime type belongs.
    */
-  public static MediaType getMediaType(String mediaType) {
+  public static MediaType getMediaType(String mimeType) {
```
JoanaCMS, line 66: Just a small detail but why change from `mediaType` to `mimeType`? For me makes makes more sense to use `mediaType`

metis-indexing/src/main/java/eu/europeana/indexing/tiers/media/TextClassifier.java
```diff
@@ -41,16 +42,28 @@ MediaTier classifyWebResource(WebResourceWrapper webResource, boolean hasLanding
...
+      mediaTier = MediaTier.T0;
     }
-    return result;
+
+    return mediaTier;
+  }
+
+  @Override
+  ResolutionTierMetadata extractResolutionTierMetadata(WebResourceWrapper webResource, MediaTier mediaTier) {
+    return new ResolutionTierMetadataBuilder().setImageResolution(webResource.getSize())
```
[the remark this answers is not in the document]
stzanakis: Because it can be a pdf which is considered as image.
At least that's what the specification says.

metis-schema/src/main/java/eu/europeana/metis/schema/model/MediaType.java
stzanakis: I'll disagree.
We call it MediaType as being a media of a type group we have defined.
That media type group is selected based on the mime type, and the mime type comes from the http protocol in fact and this is what is supposed to be passed in this method, so that we can classify it.

See https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types

PR #507, opened 2022-02-16
MET-4233 Secure Dereference service from url request attacks

JoanaCMS:


metis-dereference/metis-dereference-import/src/main/java/eu/europeana/metis/dereference/vocimport/utils/DereferenceValidationUtils.java
```diff
@@ -0,0 +1,64 @@
...
+import java.util.Properties;
+
+/**
+ * Utils class focused on validating values used in dereferencing
+ */
+public class DereferenceValidationUtils {
+
+    private static final String PROPERTY_FIELD_NAME = "valid.directories.values";
+    private String dereferencingPropertiesLocation =
+            "metis-dereference/metis-dereference-rest/src/main/resources/dereferencing.properties";
```
stzanakis, line 21: We shouldn't be doing that here. We are hardcoding the location of the file.
Instead, the property should be read where it's used, in this case the dereferencing rest(like the rest of the values). 
And the white listed url prefixes should be propagated here.

metis-dereference/metis-dereference-import/src/main/java/eu/europeana/metis/dereference/vocimport/utils/DereferenceValidationUtils.java
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    /**
+     * Method that verifies if the given directory is valid
+     * @param directoryToEvaluate The directory to evaluate as a string
+     * @return True if is valid; False otherwise
+     */
+    public boolean isDirectoryValid(String directoryToEvaluate) {
+        if (CollectionUtils.isEmpty(validDirectoriesValues)) {
+            return false;
```
stzanakis, line 45: Try to reduce returns by having a variable and then returning that once at the end.

metis-dereference/metis-dereference-import/src/main/java/eu/europeana/metis/dereference/vocimport/utils/DereferenceValidationUtils.java
```diff
@@ -0,0 +1,64 @@
+package eu.europeana.metis.dereference.vocimport.utils;
+
+import edu.emory.mathcs.backport.java.util.Arrays;
```
stzanakis, line 3: Wrong import here

stzanakis (changes requested):

metis-dereference/metis-dereference-import/src/main/java/eu/europeana/metis/dereference/vocimport/utils/DereferenceValidationUtils.java
JoanaCMS: Thanks for pointing it out! I didn't realize this, should pay more attention to this. I will change this

metis-dereference/metis-dereference-import/src/main/java/eu/europeana/metis/dereference/vocimport/VocabularyCollectionImporterFactory.java
```diff
@@ -12,25 +15,43 @@
...
    * Note that this method can only be used for locations that are also a valid {@link
    * java.net.URL}.
    *
    * @param directoryLocation The location of the directory to import.
+   * @throws VocabularyImportException if a problem occurs when verifying directory
    * @return A vocabulary importer.
    */
-  public VocabularyCollectionImporter createImporter(URI directoryLocation) {
+  public VocabularyCollectionImporter createImporter(URI directoryLocation) throws VocabularyImportException {
+    if(!dereferenceValidationUtils.isDirectoryValid(directoryLocation.toString())){
```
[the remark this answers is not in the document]
JoanaCMS: At least sonar did not show any warning related to this for every push, I've been checking on it while working on this ticket. I can run in jenkins to see if it shows up there

metis-dereference/metis-dereference-import/src/main/java/eu/europeana/metis/dereference/vocimport/utils/DereferenceValidationUtils.java
```diff
@@ -0,0 +1,64 @@
...
+     * @param directoryToEvaluate The directory to evaluate as a string
+     * @return True if is valid; False otherwise
+     */
+    public boolean isDirectoryValid(String directoryToEvaluate) {
+        if (CollectionUtils.isEmpty(validDirectoriesValues)) {
+            return false;
+        } else if (validDirectoriesValues.size() == 1 && StringUtils.equals(validDirectoriesValues.get(0), "*")) {
+            return true;
+        } else {
+            return validDirectoriesValues.stream().anyMatch(directoryToEvaluate::contains);
```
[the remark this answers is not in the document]
JoanaCMS: I wasn't sure that the valid values would be written as "it should start like this" or that "it should be from this website", so I wrote `contains` to play it safe. But thanks for clarifying that, I'll change this to starts with

metis-dereference/metis-dereference-import/src/main/java/eu/europeana/metis/dereference/vocimport/utils/DereferenceValidationUtils.java
JoanaCMS: Done

metis-dereference/metis-dereference-import/src/main/java/eu/europeana/metis/dereference/vocimport/utils/DereferenceValidationUtils.java
JoanaCMS: I was struggling in the beginning on how to do that, but after doing some search, I believe I found a way to do exactly this. I will changes this

metis-dereference/metis-dereference-import/src/main/java/eu/europeana/metis/dereference/vocimport/utils/DereferenceValidationUtils.java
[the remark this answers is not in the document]
stzanakis: The problem with contains is that it can be bypassed by setting it as a path or query parameter e.g.
exploiter.com/whitelisted.domain.com/more/path
exploiter.com?query=whitelisted.domain.com

metis-dereference/metis-dereference-import/src/main/java/eu/europeana/metis/dereference/vocimport/utils/DereferenceValidationUtils.java
JoanaCMS: I have changed this part of the code. I've used spring annotations to handle this

metis-dereference/metis-dereference-rest/src/main/resources/dereferencing.properties.example
```diff
@@ -12,4 +12,6 @@ mongo.username=
 mongo.password=
 mongo.application.name=
 entity.db=
-vocabulary.db=
\ No newline at end of file
+vocabulary.db=
+
+valid.directories.values=
```
stzanakis, line 17: Exactly, that was what I meant!
But isn't the naming confusing? We are passing here url prefixes and not the actual "values". I think same applies for the naming convention down the scope.

metis-dereference/metis-dereference-import/src/main/java/eu/europeana/metis/dereference/vocimport/VocabularyCollectionImporterFactory.java
```diff
@@ -12,25 +15,44 @@
...
   /**
    * Create a vocabulary importer for local files, indicated by instances of {@link Path}.
    *
    * @param directoryLocation The location of the directory file to import.
+   * @throws VocabularyImportException if a problem occurs when verifying directory
    * @return A vocabulary importer.
    */
-  public VocabularyCollectionImporter createImporter(Path directoryLocation) {
+  public VocabularyCollectionImporter createImporter(Path directoryLocation) throws VocabularyImportException {
+    if(!dereferenceValidationUtils.isDirectoryValid(directoryLocation.toString())){
```
stzanakis, line 53: How about inverting this?
So check for true instead of "!" and return in the if, and outside of if throw.
Makes it slightly more readable.
(Same for the other occasion)

metis-dereference/metis-dereference-import/src/main/java/eu/europeana/metis/dereference/vocimport/utils/DereferenceValidationUtils.java
```diff
@@ -0,0 +1,40 @@
...
+
+import org.apache.commons.collections.CollectionUtils;
+
+import java.util.Arrays;
+import java.util.List;
+
+/**
+ * Utils class focused on validating values used in dereferencing
+ */
+public class DereferenceValidationUtils {
```
stzanakis, line 11: Isn't this a little over engineered? I mean it's one simple method that could simply be part of the VocabularyCollectionImporterFactory class instead.

JoanaCMS: When I first created this class, it made sense to separate the "responsibilities".
But I agree with you, now it does look over engineered. I will fix this

metis-dereference/metis-dereference-import/src/main/java/eu/europeana/metis/dereference/vocimport/VocabularyCollectionImporterFactory.java
JoanaCMS: I made it has it is so it would be simple for me to follow. But that is a good point, since it is always used to check the negative of it. I will change this

stzanakis (approved):
We'll wait for the new snapshot

PR #510, opened 2022-03-03
MET-4237 Align de-publication Status

adolfopeixinho:
https://europeana.atlassian.net/browse/MET-4237

### DIFF
--- a/metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/execution/WorkflowPostProcessor.java
+++ b/metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/execution/WorkflowPostProcessor.java
@@ -211,7 +211,6 @@ public class WorkflowPostProcessor {
     final List<SubTaskInfo> subTasks = new ArrayList<>();
     List<SubTaskInfo> subTasksBatch;
     do {
-      // need to change dpsCline call
       subTasksBatch = retryableExternalRequestForNetworkExceptionsThrowing(
           () -> dpsClient.getDetailedTaskReportBetweenChunks(
               depublishPlugin.getTopologyName(), externalTaskId, subTasks.size(),
--- a/metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/execution/WorkflowPostProcessor.java
+++ b/metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/execution/WorkflowPostProcessor.java
@@ -18,6 +18,8 @@ import eu.europeana.metis.core.dataset.Dataset.PublicationFitness;
 import eu.europeana.metis.core.dataset.DepublishRecordId.DepublicationStatus;
 import eu.europeana.metis.core.exceptions.InvalidIndexPluginException;
 import eu.europeana.metis.core.service.OrchestratorService;
+import eu.europeana.metis.core.util.DepublishRecordIdSortField;
+import eu.europeana.metis.core.util.SortDirection;
 import eu.europeana.metis.core.workflow.WorkflowExecution;
 import eu.europeana.metis.core.workflow.plugins.AbstractExecutablePlugin;
 import eu.europeana.metis.core.workflow.plugins.AbstractMetisPlugin;
@@ -27,6 +29,7 @@ import eu.europeana.metis.core.workflow.plugins.IndexToPreviewPlugin;
 import eu.europeana.metis.core.workflow.plugins.IndexToPublishPlugin;
 import eu.europeana.metis.core.workflow.plugins.MetisPlugin;
 import eu.europeana.metis.core.workflow.plugins.PluginType;
+import eu.europeana.metis.exception.BadContentException;
 import java.util.ArrayList;
 import java.util.Date;
 import java.util.List;
@@ -40,7 +43,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 /**
- * This object can perform post processing for workflows.
+ * This object can perform post-processing for workflows.
  */
 public class WorkflowPostProcessor {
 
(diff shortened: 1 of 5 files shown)


metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/execution/WorkflowPostProcessor.java
```diff
@@ -105,8 +115,21 @@ private void indexPostProcess(AbstractExecutablePlugin<?> indexPlugin, String da
     } else if (indexPlugin instanceof IndexToPublishPlugin) {
       targetIndexingDatabase = ((IndexToPublishPlugin) indexPlugin).getTargetIndexingDatabase();
       targetIndexingEnvironment = ((IndexToPublishPlugin) indexPlugin).getTargetIndexingEnvironment();
-      //Reset depublish status
-      depublishRecordIdDao.markRecordIdsWithDepublicationStatus(datasetId, null,
+
+      // get all tasks from dataset id and topology name
+      List<SubTaskInfo> taskReport = dpsClient.getDetailedTaskReport(indexPlugin.getTopologyName(),
```
stzanakis, line 120: I don't think we should be doing it this way.
We are requesting here all the datasetIds with one call from dps, right?
What if the ids are in the millions or hundreds of thousands?
This not only can have a huge impact in performance, even timeout on the request but also memory consumption.

A better way would be to send the ids that we are interested in and dps should tell us if those were affected on it's response.

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/execution/WorkflowPostProcessor.java
```diff
@@ -91,12 +98,15 @@ void performPluginPostProcessing(AbstractExecutablePlugin<?> plugin, String data
   /**
    * Performs post-processing for indexing plugins
    *
-   * @param indexPlugin the index plugin
-   * @param datasetId the dataset id
-   * @throws DpsException if communication with ecloud dps failed
+   * @param indexPlugin The index plugin
+   * @param datasetId   The dataset id
+   * @throws DpsException                If communication with e-cloud dps failed
```
stzanakis, line 103: I don't think we have this spacing in our code style(I'm not sure though).
Are you using the correct one?

stzanakis (changes requested):

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/execution/WorkflowPostProcessor.java
```diff
@@ -40,13 +43,13 @@
...
 /**
- * This object can perform post processing for workflows.
+ * This object can perform post-processing for workflows.
  */
 public class WorkflowPostProcessor {
 
   private static final Logger LOGGER = LoggerFactory.getLogger(WorkflowPostProcessor.class);
 
-  private static final int ECLOUD_REQUEST_BATCH_SIZE = 100;
+  private static final int ECLOUD_REQUEST_BATCH_SIZE = 1000;
```
jeortizquan, line 52: This amount was according to what was discussed with eCloud team.

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/execution/WorkflowPostProcessor.java
```diff
@@ -152,32 +192,45 @@ private void depublishDatasetPostProcess(String datasetId) {
...
 
     // Retrieve the successfully depublished records.
     final long externalTaskId = Long.parseLong(depublishPlugin.getExternalTaskId());
     final List<SubTaskInfo> subTasks = new ArrayList<>();
     List<SubTaskInfo> subTasksBatch;
     do {
-      subTasksBatch = retryableExternalRequestForNetworkExceptionsThrowing(() -> dpsClient.getDetailedTaskReportBetweenChunks(
-          depublishPlugin.getTopologyName(), externalTaskId, subTasks.size(),
-          subTasks.size() + ECLOUD_REQUEST_BATCH_SIZE));
+      // need to change dpsCline call
```
jeortizquan, line 214: remove code note.

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/execution/WorkflowPostProcessor.java
```diff
@@ -105,9 +115,36 @@ private void indexPostProcess(AbstractExecutablePlugin<?> indexPlugin, String da
     } else if (indexPlugin instanceof IndexToPublishPlugin) {
       targetIndexingDatabase = ((IndexToPublishPlugin) indexPlugin).getTargetIndexingDatabase();
       targetIndexingEnvironment = ((IndexToPublishPlugin) indexPlugin).getTargetIndexingEnvironment();
-      //Reset depublish status
-      depublishRecordIdDao.markRecordIdsWithDepublicationStatus(datasetId, null,
-          DepublicationStatus.PENDING_DEPUBLICATION, null);
+
+      final long totalRecords = dpsClient.getTotalMetisDatabaseRecords(indexPlugin.getExternalTaskId(),
```
jeortizquan, line 119: the totalRecords is type ```long```
create a ticket for ```getDetailedTaskReportBetweenChunks``` receives an ```int``` if the amount it will fail because of variable type. For this time we are skipping this to get the environment running again.

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/execution/WorkflowPostProcessor.java
```diff
@@ -105,9 +115,36 @@ private void indexPostProcess(AbstractExecutablePlugin<?> indexPlugin, String da
...
+          ((IndexToPublishPlugin) indexPlugin).getTargetIndexingDatabase());
+      List<SubTaskInfo> subTaskInfoList;
+
+      // get chunked tasks from dataset id and topology name
+      for (int i = 0; i < totalRecords; i = +ECLOUD_REQUEST_BATCH_SIZE) {
```
jeortizquan, line 124: fix this expression because it will always be a 1000 ```i = +ECLOUD_REQUEST_BATCH_SIZE```

jeortizquan (changes requested):
There is some small changes to review.

jeortizquan (approved):
This is will have a continuation story with MET-4321 in ecloud side, so this will need to be rewrited.

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/execution/WorkflowPostProcessor.java
jeortizquan: Discussed that it will be a new implementation MET-4321, we will use this one as first approach to watch how it performs.

PR #511, opened 2022-03-04
MET-4159 Make Http and Zip file harvesting Asynchronous

JoanaCMS:


metis-harvesting/src/main/java/eu/europeana/metis/harvesting/http/HttpHarvesterImpl.java
```diff
@@ -206,6 +188,17 @@ public FileIterator(Path extractedDirectory) {
...
+    @Override
+    public void deleteIteratorContent() {
+      if (extractedDirectory != null) {
+        try {
+          FileUtils.deleteDirectory(extractedDirectory.toFile());
+        } catch (IOException e) {
+          LOGGER.warn("Could not delete directory.", e);
+        }
+      }
+    }
```
jeortizquan, line 200: add logger when ```extractedDirectory``` is ```null```

metis-harvesting/src/main/java/eu/europeana/metis/harvesting/http/HttpHarvesterImpl.java
```diff
@@ -126,6 +87,27 @@ public HttpRecordIterator harvestRecords(String archiveUrl, String downloadDirec
...
+
+      tempDir = Files.createTempDirectory(prefix);
+      tempFile = Files.createTempFile(tempDir, prefix, compressedFileType.getExtension());
+      FileUtils.copyInputStreamToFile(input, tempFile.toFile());
+
+      return harvestRecords(tempFile);
+      } catch (IOException e) {
+        throw new HarvesterException("Problem saving archive.", e);
+      }
+
```
jeortizquan, line 108: close ```input``` stream to prevent memory leakage. if applies use try-with-resources to close stream or do it manually.

metis-harvesting/src/main/java/eu/europeana/metis/harvesting/http/HttpHarvesterImpl.java
```diff
@@ -43,68 +42,30 @@
```
jeortizquan, line 42: unit tests for this file are missing.

jeortizquan (changes requested):
There are small changes to include. See comments. 
Also sonar quality gate fails. changes need to be included also.

metis-harvesting/src/main/java/eu/europeana/metis/harvesting/http/HttpHarvesterImpl.java
stzanakis: As we discussed, we allow this in this iteration

metis-harvesting/src/main/java/eu/europeana/metis/harvesting/http/HttpHarvesterImpl.java
stzanakis: I don't see why that would be useful here.

metis-harvesting/src/main/java/eu/europeana/metis/harvesting/http/HttpHarvesterImpl.java
stzanakis: The input stream is closed insdie the method copyInputStreamToFile()

metis-harvesting/src/main/java/eu/europeana/metis/harvesting/http/HttpHarvesterImpl.java
jeortizquan: It is the feedback that nothing was removed when method was called.

stzanakis: I understand, but why would we care to know if there was nothing to remove? :D
Also normally there should be something to remove if someone is using this method.

stzanakis (approved):

metis-harvesting/src/main/java/eu/europeana/metis/harvesting/http/HttpHarvesterImpl.java
jeortizquan: In case of reading logs we can detect why the things are not being deleted, and  we can determine that this path of code is being followed. perhaps ```LOGGER.debug or warn("Nothing has been deleted")``` I had this idea when we read logs later that is the reason why I think.

PR #518, opened 2022-03-16
Ea-2891 add profiling

SrishtiSingh-eu:


SrishtiSingh-eu:
Will not merge

SrishtiSingh-eu:
it's obsolete as well because we've been measuring the time with APM essentially now as well.

PR #521, opened 2022-03-29
MET-4374 Reproduce error

stzanakis:


metis-dereference/metis-dereference-common/src/main/java/eu/europeana/metis/dereference/IncomingRecordToEdmTransformer.java
```diff
@@ -0,0 +1,148 @@
...
+
+  /**
+   * Validates if the provided xml can be parsed.
+   *
+   * @param xml the xml string
+   * @throws ParserConfigurationException if xml parsing failed
+   * @throws IOException if xml parsing failed
+   * @throws SAXException if xml parsing failed
+   */
+  private void isValidXml(String xml) throws ParserConfigurationException, IOException, SAXException {
```
JoanaCMS, line 131: The naming for this type of method is a bit weird. At first I thought it was a `boolean` method type, but then I had to look at it again to see that it was actually `void`. Also the javadoc gives the impression that this is to return a `boolean` type. 
Looking at how it is right now, how will the user know it is valid or not? Perhaps we could rename this to something else?

JoanaCMS (changes requested):

PR #522, opened 2022-03-30
Ea 2890 entity client implmentation

SrishtiSingh-eu:


metis-enrichment/metis-enrichment-common/src/main/java/eu/europeana/enrichment/utils/EntityResolverUtils.java
```diff
@@ -0,0 +1,105 @@
...
+
+    public static <T extends ReferenceTerm> Function<T, EntityClientRequest> inputFunctionForRefSearch() {
+        return term -> new EntityClientRequest(term.getReference().toString(),
+                term.getCandidateTypes()
+                        .stream()
+                        .map(entityType -> entityType.name().toLowerCase())
+                        .collect(Collectors.joining(",")), true);
+    }
+
+    public static <T>  Map<T, EnrichmentBase> getIdResults(Map<T, List<EnrichmentBase>> results) {
```
gsergiu, line 57: this seems to convert Map<T, List<EnrichmentBase>> results to Map<T, EnrichmentBase> by picking first in the list. Better naming and javadoc would be needed

metis-enrichment/metis-enrichment-rest/.gitignore
```diff
@@ -1,2 +1,5 @@
 ##Add to ignore to not commit by mistake
-/src/main/resources/enrichment.properties
\ No newline at end of file
+/src/main/resources/enrichment.properties
+
+# properties file
+**/entity-client.properties
```
gsergiu, line 5: this can be added to metis-enrichment/.gitignore it will match in modules as well

metis-enrichment/metis-enrichment-rest/pom.xml
```diff
@@ -66,12 +66,7 @@
     </dependency>
     <dependency>
       <groupId>io.springfox</groupId>
-      <artifactId>springfox-swagger2</artifactId>
-      <version>${version.swagger}</version>
-    </dependency>
-    <dependency>
-      <groupId>io.springfox</groupId>
-      <artifactId>springfox-swagger-ui</artifactId>
+      <artifactId>springfox-boot-starter</artifactId>
```
gsergiu, line 69: is this required, need to check with metis team on how the service is deployed

metis-enrichment/metis-enrichment-rest/src/main/resources/log4j2.xml
```diff
@@ -9,5 +9,7 @@
 		<Root level="INFO">
 			<AppenderRef ref="Console"/>
 		</Root>
+		<!-- Set to DEBUG to log all requests to backend systems -->
+		<Logger name="eu.europeana.entity.client" level="DEBUG"/>
```
gsergiu, line 13: need to discuss with Metis team, but I assume that log4j should be submitted only as template

gsergiu (changes requested):
There would be a couple of renamings to do and a cross check with the metis team for the mapping of entitites fields (some seem to be missing)

metis-enrichment/metis-enrichment-client/pom.xml
```diff
@@ -44,15 +44,15 @@
...
-    </dependency>
     <dependency>
       <groupId>org.springframework</groupId>
       <artifactId>spring-web</artifactId>
       <version>${version.spring}</version>
     </dependency>
+    <dependency>
+      <groupId>org.apache.logging.log4j</groupId>
+      <artifactId>log4j-api</artifactId>
+      <version>2.17.1</version>
```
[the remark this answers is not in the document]
SrishtiSingh-eu: The dependency should be removed as it's not used by the client module anymore. the code moved to the common and rest module. hence removing it.

metis-enrichment/metis-enrichment-common/src/main/java/eu/europeana/enrichment/api/external/impl/EntityClientResolver.java
```diff
@@ -0,0 +1,188 @@
...
+    public <T extends ReferenceTerm> Map<T, List<EnrichmentBase>> resolveByUri(Set<T> referenceTerms) {
+        final Function<T, EntityClientRequest> inputFunction = EntityResolverUtils.inputFunctionForRefSearch();
+        return performInBatches(referenceTerms, inputFunction);
+    }
+
+    /**
+     * Gets the Enrichment for Multiple values in batches
+     *
+     * @param inputValues
+     * @param identity
```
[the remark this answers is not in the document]
SrishtiSingh-eu: done

metis-enrichment/metis-enrichment-common/src/main/java/eu/europeana/enrichment/api/external/impl/EntityClientResolver.java
```diff
@@ -0,0 +1,188 @@
...
+            List<Entity> parentEntities = new ArrayList<>();
+            entities.stream().forEach(entity -> {
+                if(EntityResolverUtils.isParentEntityRequired(entity, clientRequest) && entity.getIsPartOfArray() != null) {
+                    getParentEntities(entity, parentEntities);
+                }
+            });
+            // 3. add the parent entities
+            entities.addAll(parentEntities);
+            // 4. convert entity to EnrichmentBase
+            List<EnrichmentBase> enrichmentBaseList = entities.stream().map(entity -> {
```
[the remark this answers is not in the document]
SrishtiSingh-eu: done

metis-enrichment/metis-enrichment-common/src/main/java/eu/europeana/enrichment/api/external/impl/EntityClientResolver.java
```diff
@@ -0,0 +1,188 @@
...
+     */
+    private void getParentEntities(Entity entity, List<Entity> parentEntities) {
+        entity.getIsPartOfArray().stream().forEach(parentEntityId -> {
+            if (!EntityResolverUtils.checkIfEntityAlreadyExists(parentEntityId, parentEntities)) {
+                getParent(parentEntityId, parentEntities);
+            }
+        });
+    }
+
+    private void getParent(String parentEntityId, List<Entity> parentEntities) {
```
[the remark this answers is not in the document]
SrishtiSingh-eu: renamed to appendParent

metis-enrichment/metis-enrichment-common/src/main/java/eu/europeana/enrichment/api/external/impl/EntityClientResolver.java
```diff
@@ -0,0 +1,188 @@
...
+    }
+
+    /**
+     * Converts the EM model class to Metis XML Entity class
+     *
+     * @param entity EM Entity
+     * @param <T>    class that extends EnrichmentBase
+     * @return
+     */
+    private static <T extends EnrichmentBase> T createXmlEntity(Entity entity) {
```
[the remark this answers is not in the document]
SrishtiSingh-eu: Moved to Utils class. renamed to **convertEntityToEnrichmentBase**

metis-enrichment/metis-enrichment-common/src/main/java/eu/europeana/enrichment/api/external/model/AgentBase.java
```diff
@@ -204,4 +219,63 @@ public List<Part> getSameAs() {
   public void setSameAs(List<Part> sameAs) {
     this.sameAs = cloneListAcceptingNull(sameAs);
   }
+
+  private void init(eu.europeana.entitymanagement.definitions.model.Agent agent){
+    if (agent.getHiddenLabel() != null) {
```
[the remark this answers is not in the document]
SrishtiSingh-eu: I have verified that all the fields if present are mapped. I have for now covered all the fields in AgentBase and other classes. If there is something else that needs to be added after discussion, let me know

metis-enrichment/metis-enrichment-common/src/main/java/eu/europeana/enrichment/api/external/model/Concept.java
```diff
@@ -131,4 +139,40 @@ public List<Resource> getRelatedMatch() {
   public void setRelatedMatch(List<Resource> relatedMatch) {
     this.relatedMatch = cloneListAcceptingNull(relatedMatch);
   }
+
+  private void init(eu.europeana.entitymanagement.definitions.model.Concept concept){
+    if (concept.getHiddenLabel() != null) {
```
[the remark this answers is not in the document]
SrishtiSingh-eu: yes, maybe in another ticket we can verify fields for all entity types and add what's missing

metis-enrichment/metis-enrichment-common/src/main/java/eu/europeana/enrichment/api/external/model/EntityClientRequest.java
```diff
@@ -0,0 +1,76 @@
+package eu.europeana.enrichment.api.external.model;
+
+// Request class for sending request to Entity API client
+public class EntityClientRequest {
```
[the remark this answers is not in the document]
SrishtiSingh-eu: renamed

metis-enrichment/metis-enrichment-common/src/main/java/eu/europeana/enrichment/api/external/model/Place.java
```diff
@@ -76,4 +86,25 @@ public void setAlt(String alt) {
     this.alt = alt;
   }
 
+  private void init(eu.europeana.entitymanagement.definitions.model.Place place) {
+    if (place.getIsPartOfArray() != null) {
```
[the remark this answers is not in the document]
SrishtiSingh-eu: yes, maybe in another ticket we can verify fields for all entity types and add what's missing

metis-enrichment/metis-enrichment-common/src/main/java/eu/europeana/enrichment/api/external/model/TimeSpan.java
```diff
@@ -84,4 +94,28 @@ public Part getIsNextInSequence() {
   public void setIsNextInSequence(Part isNextInSequence) {
     this.isNextInSequence = isNextInSequence;
   }
+
+  private void init(eu.europeana.entitymanagement.definitions.model.TimeSpan timeSpan) {
+    if (timeSpan.getIsPartOfArray() != null) {
```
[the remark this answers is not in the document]
SrishtiSingh-eu: yes, maybe in another ticket we can verify fields for all entity types and add what's missing

metis-enrichment/metis-enrichment-common/src/main/java/eu/europeana/enrichment/utils/EntityResolverUtils.java
```diff
@@ -0,0 +1,105 @@
...
+            if (currentPartition.size() >= batchSize) {
+                currentPartition = new ArrayList<>();
+                partitions.add(currentPartition);
+            }
+            currentPartition.add(item);
+        });
+        return partitions;
+    }
+
+    public static <T extends SearchTerm> Function<T, EntityClientRequest> inputFunctionForTextSearch() {
```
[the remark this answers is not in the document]
SrishtiSingh-eu: more like we are creating Enrichment Query with the inputs of SearchTerm. 
 So createEnrichmentQueryForTextSearch

metis-enrichment/metis-enrichment-common/src/main/java/eu/europeana/enrichment/utils/EntityResolverUtils.java
```diff
@@ -0,0 +1,105 @@
...
+        return term -> new EntityClientRequest(term.getTextValue(), term.getLanguage(),
+                term.getCandidateTypes()
+                        .stream()
+                        .map(entityType -> entityType.name().toLowerCase())
+                        .collect(Collectors.joining(",")), false);
+    }
+
+    public static <T extends ReferenceTerm> Function<T, EntityClientRequest> inputFunctionForRefSearch() {
```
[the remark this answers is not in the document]
SrishtiSingh-eu: createEnrichmentQueryForRefSearch

metis-enrichment/metis-enrichment-common/src/main/java/eu/europeana/enrichment/utils/EntityResolverUtils.java
SrishtiSingh-eu: done

metis-enrichment/metis-enrichment-rest/pom.xml
SrishtiSingh-eu: I added this as a fix for swagger, as I was getting bean exception for swagger and the application was not starting. Now the application works fine expect the swagger mapping. For now swagger doesn't work anymore, there is a mapping dis ambiguity in this branch. But that should be fixed in a separate ticket.

metis-enrichment/metis-enrichment-rest/src/main/resources/log4j2.xml
SrishtiSingh-eu: yes, but for now I will comment this. Usually, I use this for local testing.

metis-enrichment/metis-enrichment-rest/.gitignore
SrishtiSingh-eu: done

EuropeanaDevelopment (commented):
I've started checking and will fix things one the way.
Please do not push any changes.
I would leave single comments/questions I'll see on the way, so please answer them individually when you can.
(Simon)

metis-enrichment/metis-enrichment-common/src/main/java/eu/europeana/enrichment/api/external/impl/EntityClientResolver.java
```diff
@@ -0,0 +1,174 @@
...
+
+    private List<EnrichmentBase> executeEntityClientRequest(EnrichmentQuery clientRequest) {
+        // 1. get Entities
+        List<Entity> entities = getEntities(clientRequest);
+        if (!entities.isEmpty()) {
+            // 2. get the parent entities
+            List<Entity> parentEntities = new ArrayList<>();
+            entities.stream().forEach(entity -> {
+                if (entity.getIsPartOfArray() != null && EntityResolverUtils.isTextOrUriSearch(clientRequest)) {
+                    fetchParentEntities(entity, parentEntities);
```
stzanakis, line 89: So this is done "remotely". In metis is done on the Persistent resolver which inside the service.
Could be quite a performance impact here.

metis-enrichment/metis-enrichment-common/src/main/java/eu/europeana/enrichment/api/external/model/EnrichmentQuery.java
```diff
@@ -0,0 +1,76 @@
+package eu.europeana.enrichment.api.external.model;
+
+// Request class for sending request to Entity API client
+public class EnrichmentQuery {
```
stzanakis, line 4: Is there a particular reason for this class?
We have a whole set of objects that represent the search.
EnrichmentReference
EnrichmentSearch
ReferenceValue
SearchValue

metis-enrichment/metis-enrichment-common/src/main/java/eu/europeana/enrichment/api/external/impl/EntityClientResolver.java
```diff
@@ -0,0 +1,174 @@
...
+
+
+    public EntityClientResolver(int batchSize) {
+        this.batchSize = batchSize;
+        entityClientApi = new EntityClientApiImpl();
+    }
+
+    @Override
+    public <T extends SearchTerm> Map<T, List<EnrichmentBase>> resolveByText(Set<T> searchTerms) {
+        final Function<T, EnrichmentQuery> inputFunction = EntityResolverUtils.createEnrichmentQueryForTextSearch();
```
stzanakis, line 39: How about the 3 and 2 code language check.
In contrary to the fetching parent entities, I don't see the language mapping implemented anyway in this code.

metis-enrichment/metis-enrichment-rest/src/test/java/eu/europeana/enrichment/rest/enrichment/EntityClientResolverITTest.java
```diff
@@ -0,0 +1,82 @@
...
+import java.net.URL;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+// TODO - This is a Integration Test. Once we have measured performance of the new implementation,
+//  proper junit test will be added.
+@Disabled
+public class EntityClientResolverITTest {
```
stzanakis, line 23: We don't do those kind of integration tests in metis.
At least not yet.
If you want this kind of test, it should be implemented with Wiremock, otherwise we'll leave it disabled for now.

metis-enrichment/metis-enrichment-common/src/main/java/eu/europeana/enrichment/api/external/impl/EntityClientResolver.java
SrishtiSingh-eu: I haven't added the ISO language check here. But the languages are validated in entity-api. There is no harm in adding a language validation here, but it will be redundant then.

metis-enrichment/metis-enrichment-common/src/main/java/eu/europeana/enrichment/api/external/impl/EntityClientResolver.java
```diff
@@ -0,0 +1,174 @@
...
+    }
+
+    private List<EnrichmentBase> executeEntityClientRequest(EnrichmentQuery clientRequest) {
+        // 1. get Entities
+        List<Entity> entities = getEntities(clientRequest);
+        if (!entities.isEmpty()) {
+            // 2. get the parent entities
+            List<Entity> parentEntities = new ArrayList<>();
+            entities.stream().forEach(entity -> {
+                if (entity.getIsPartOfArray() != null && EntityResolverUtils.isTextOrUriSearch(clientRequest)) {
```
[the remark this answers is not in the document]
SrishtiSingh-eu: The second check is to avoid executing the parent entity code for ID search.

metis-enrichment/metis-enrichment-common/src/main/java/eu/europeana/enrichment/api/external/impl/EntityClientResolver.java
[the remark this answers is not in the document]
SrishtiSingh-eu: Here is the ticket for Parent entity logic - [EA-2984](https://europeana.atlassian.net/browse/EA-2984). It is mentioned to fetch parent entities for both text and Uri search. One thing I am getting confused about is the service side and client side. Are the functionality of enrichment different in service and client ??

metis-enrichment/metis-enrichment-rest/src/test/java/eu/europeana/enrichment/rest/enrichment/EntityClientResolverITTest.java
SrishtiSingh-eu: Yes, we can leave it as disabled or we can remove it as well.

metis-enrichment/metis-enrichment-rest/src/main/java/eu/europeana/enrichment/rest/config/Application.java
```diff
@@ -59,6 +48,9 @@ public class Application implements WebMvcConfigurer, InitializingBean {
   @Value("${enrichment.mongo.application.name}")
   private String enrichmentMongoApplicationName;
 
+  @Value("${enrichment.batch.size}")
+  private int enrichmentBatchSize;
```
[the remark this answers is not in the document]
SrishtiSingh-eu: okay.

metis-enrichment/metis-enrichment-common/src/main/java/eu/europeana/enrichment/api/external/impl/EntityClientResolver.java
stzanakis: So when you say validated does that mean that either a 2 letter or 3 letter code will produce a result?
If the answer is yes then indeed it would be redundant to add a check here.

metis-enrichment/metis-enrichment-common/src/main/java/eu/europeana/enrichment/api/external/impl/EntityClientResolver.java
stzanakis: How about this one @SrishtiSingh-eu ?

metis-enrichment/metis-enrichment-common/src/main/java/eu/europeana/enrichment/api/external/model/EnrichmentQuery.java
stzanakis: Ok I'll see what I can do.

metis-enrichment/metis-enrichment-common/src/main/java/eu/europeana/enrichment/api/external/impl/EntityClientResolver.java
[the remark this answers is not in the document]
stzanakis: Oh I see! So because the method is used by both cases, for one of the cases(the **resolveById**) you don't want the parents.
From what I see the check allows for references that their value start with **EntityApiConstants.BASE_URL**.

metis-enrichment/metis-enrichment-common/src/main/java/eu/europeana/enrichment/api/external/impl/EntityClientResolver.java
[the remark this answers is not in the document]
stzanakis: Ok. But should this be done here or inside the service?
I'm asking because for each items the parents have to be resolved remotely while it could be done much faster inside the service.
Also how about co-referencing logic??

metis-enrichment/metis-enrichment-rest/src/test/java/eu/europeana/enrichment/rest/enrichment/EntityClientResolverITTest.java
stzanakis: Okay.

metis-enrichment/metis-enrichment-common/src/main/java/eu/europeana/enrichment/api/external/impl/EntityClientResolver.java
```diff
@@ -0,0 +1,174 @@
...
+            throw new UnknownException("Entity Client call to " + (entityClientRequest.isReference() ? "getEntityByUri" : "getSuggestions")
+                    + " failed for" + entityClientRequest + ".", e);
+        } catch (RuntimeException e) {
+            throw new UnknownException("Entity Client call failed for : " + entityClientRequest + ".", e);
+        }
+        return Collections.emptyList();
+    }
+
+    private List<Entity> resolveReferences(EnrichmentQuery entityClientRequest) {
+        if (entityClientRequest.getValueToEnrich().startsWith(EntityApiConstants.BASE_URL)) {
```
stzanakis, line 127: So id checks are only considered values that start with **EntityApiConstants.BASE_URL**?
Metis does not have this distintion.
For example in metis when a client requests **resolveById** which does **not** start with **EntityApiConstants.BASE_URL** it will be checked as **resolveById** and not as **resolveByUri**.

With this change here it means that **resolveById** with values that do **not** start with **EntityApiConstants.BASE_URL** will be checked as **resolveByUri** which includes parent checking etc.
So a client is requesting a **resolveById** but gets essentially a **resolveByUri**

Is this really the intended behavior?

Also the **EntityApiConstants.BASE_URL**="http://data.europeana.eu". How about https? Perhaps we should be checking with a regex here to account for https too.

metis-enrichment/metis-enrichment-common/src/main/java/eu/europeana/enrichment/api/external/impl/EntityClientResolver.java
SrishtiSingh-eu: I had a thought of having a single class with all functionality. But if it will have a performance impact here. This could be moved to Service class

metis-enrichment/metis-enrichment-common/src/main/java/eu/europeana/enrichment/api/external/impl/EntityClientResolver.java
SrishtiSingh-eu: The check is to validate ISO language ([here](https://github.com/europeana/entity-api-v2/blob/e6c3df4ed92cb457f574de54750c38dc624114a9/entity-web/src/main/java/eu/europeana/entity/web/controller/BaseRest.java#L236)) . I think that validates the two or three-letter code already.

metis-enrichment/metis-enrichment-common/src/main/java/eu/europeana/enrichment/api/external/impl/EntityClientResolver.java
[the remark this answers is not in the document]
SrishtiSingh-eu: Yes, as it's done in the same method. The check is -> if the value is a reference and **does not starts** with EntityApiConstants.BASE_URL OR is not a reference (ie; text)

metis-enrichment/metis-enrichment-common/src/main/java/eu/europeana/enrichment/api/external/impl/EntityClientResolver.java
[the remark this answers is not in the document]
SrishtiSingh-eu: Well, for us performance is important. So if we will gain better results doing this remotely, we should do this remotely in the service. 
For co-referencing logic, probably best to ask Hugo this.  From the ticket, I only know that we needed to fetch the parents for the entities.

metis-enrichment/metis-enrichment-common/src/main/java/eu/europeana/enrichment/api/external/impl/EntityClientResolver.java
SrishtiSingh-eu: So, you mean that the resolveByID can be requested without the full id like this '/agent/base/1' instead of this 'http://data.europeana.eu/agent/base/1' ??, 
I feel the simple solution should be to append the bases when we are requesting resolveByID. 
As it will be much faster to just fetch the ID directly rather than executing resolveByURI which is a two-step process

metis-enrichment/metis-enrichment-common/src/main/java/eu/europeana/enrichment/api/external/impl/EntityClientResolver.java
```diff
@@ -0,0 +1,174 @@
...
+
+    private List<Entity> getEntities(EnrichmentQuery entityClientRequest) {
+        try {
+            if (entityClientRequest.isReference()) {
+                return resolveReferences(entityClientRequest);
+            } else {
+                // if the enrich method returns more than 1 entity, then no enrichment should be (for now) considered.
+                // It works as if no entity was returned. The reason for this is that Metis does not have a disambiguation
+                // mechanism in place that can judge which entity should be chosen out of the list of options.
+                List<Entity> entities = entityClientApi.getEnrichment(entityClientRequest.getValueToEnrich(), entityClientRequest.getLanguage(), entityClientRequest.getType(), null) ;
```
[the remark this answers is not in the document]
SrishtiSingh-eu: This check when we receive the results from Entity management. So if we have **paris** as a text search, and we fetch 10 entities. The enrichment would not be considered, as it returns more than 1 entity and Metis does not have a disambiguation mechanism in place that can judge which entity should be chosen out of the list of options.

Here is the ticket [EA-2983](https://europeana.atlassian.net/browse/EA-2983).

metis-enrichment/metis-enrichment-common/src/main/java/eu/europeana/enrichment/api/external/impl/EntityClientResolver.java
stzanakis: When I say Service I don't mean the EnrichmentService class in metis code. I mean your Entity application service where the EntityClientApi point at.

SrishtiSingh-eu: Okay. Hmmm now that I think of this, a method could be added to retrieve the Parents in entity API or EM. But as I discussed it with Hugo, we were under the impression to add the code in the metis. We should discuss this with Hugo if we want to move the code there.

metis-enrichment/metis-enrichment-common/src/main/java/eu/europeana/enrichment/api/external/impl/EntityClientResolver.java
stzanakis: I don't think it does the same way metis-enrichment handles it.

Metis does the following:
- If the provided language is a 3 letter code it then converts it to a 2 letter code
- If the provided language is a 2 letter code it checks if it a valid 2 letter code otherwise uses null(which means do not check for a specific language)
- On any other case use null. A null language provided is okay as well.

While the link you mentioned just validates if the language is not null, it does not contain a comma, if it is an exact match to 'all' or it's a valid 2 letter code. If that fails it throws an exception.

So clearly the logic is different.
One of the potential issues here is that 3 letter codes will now fail.
Is this what we want?

metis-enrichment/metis-enrichment-common/src/main/java/eu/europeana/enrichment/api/external/impl/EntityClientResolver.java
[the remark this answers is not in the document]
stzanakis: I resolved the other similar comment.
So we need some decisions on the following:
- fetch parents implementation location. Now it's remotely performed, but could be in the Entity API application.
- Is the timespam broadness check implemented in the Entity API application? If not, should it? (metis checks that)
- Is the co-referencing implemented for URI searches in the Entity API application? If not, should it? (metis checks that)

metis-enrichment/metis-enrichment-common/src/main/java/eu/europeana/enrichment/api/external/impl/EntityClientResolver.java
SrishtiSingh-eu: No, this check should be added: If the provided language is a 3 letter code it then converts it to a 2 letter code. For invalid language codes, send null (no specific language). 

null language is accepted in entity  API.

metis-enrichment/metis-enrichment-common/src/main/java/eu/europeana/enrichment/api/external/impl/EntityClientResolver.java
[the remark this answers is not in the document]
stzanakis: I'm not sure I completely understand this so I might need some more context.
Why are enrichments that yield more than 1 entities are not valid? Won't we be losing a lot of enrichment information that way?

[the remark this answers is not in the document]
SrishtiSingh-eu: Let's discuss this as well.

metis-enrichment/metis-enrichment-common/src/main/java/eu/europeana/enrichment/api/external/impl/EntityClientResolver.java
stzanakis: Ok should I implement this here, or do you want this on the Entity API application?

stzanakis: After discussion:
The language conversion will be implemented in the ClientEntityResolver in the metis-enrichment libraries.

metis-enrichment/metis-enrichment-common/src/main/java/eu/europeana/enrichment/api/external/impl/EntityClientResolver.java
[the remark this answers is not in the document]
stzanakis: After discussion:
- Fetching parents should be applied for text and uri searches but not id searches to match the current metis-enrichment implementation. (Id parent fetching can be discussed/implemented later)
And the implementation should, for now, stay in the metis-enrichment code(if the re-processing operation shows a big decline on speed, we should then try to move the parent fetching on the Entity API application).
- Timespan broadness check is not required.
- For URI searches, the about check and then owlSameAs check is implemented with the reference.startsWith(EntityApiConstants.BASE_URL) where if it matches  it will use the entityClientApi.getEntityById(referenceValue) otherwise the entityClientApi.getEntityByUri(referenceValue);

metis-enrichment/metis-enrichment-common/src/main/java/eu/europeana/enrichment/api/external/impl/EntityClientResolver.java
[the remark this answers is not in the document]
stzanakis: After discussion:
- This check is valid for the case where the parent fetching is performed here(in the ClientEntityResolver).

metis-enrichment/metis-enrichment-common/src/main/java/eu/europeana/enrichment/api/external/impl/EntityClientResolver.java
stzanakis: After discussion:
- Semium requests are not expected anymore.
- The resolveById should return empty result if the id is not a europeana identifier and should not perform a resolveByUri.

metis-enrichment/metis-enrichment-common/src/main/java/eu/europeana/enrichment/api/external/impl/EntityClientResolver.java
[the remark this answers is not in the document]
stzanakis: I've added the parent searching for all search cases (id, text, uri)

PR #532, opened 2022-05-13
MET-4457 Set default truncate on elements for report

stzanakis:


metis-pattern-analysis/src/main/java/eu/europeana/patternanalysis/ProblemPatternAnalyzer.java
```diff
@@ -132,7 +134,8 @@ private List<ProblemOccurrence> checkP2(List<String> titles, List<String> descri
     equalTitlesAndDescriptions.retainAll(uniqueDescriptions);
 
     return equalTitlesAndDescriptions.stream().map(
-        value -> new ProblemOccurrence(format("Equal(lower cased) title and description: %s", value))
+        value -> new ProblemOccurrence(format("Equal(lower cased) title and description: %s(...)",
+            truncate(value, DEFAULT_MAX_CHARACTERS_ELEMENT_LENGTH_FOR_REPORT)))
```
jeortizquan, line 138: I see in database the max length definition is  ```message_report  varchar(255)``` should we use like a range min 50 chars max 255 chars?

jeortizquan: ooh I see. thanks for the explanation.

PR #550, opened 2022-05-31
MET-4572 Added test for AgentSolrCreator

adolfopeixinho:


jeortizquan (approved):
Here is possible to expand to cover more scenarios.

PR #551, opened 2022-06-01
MET-4440_MET-4575 LicenseSolrCreator unit tests

jeortizquan:


metis-indexing/src/test/java/eu/europeana/indexing/solr/property/LicenseSolrCreatorTest.java
```diff
@@ -0,0 +1,135 @@
...
+
+  @Test
+  void addToDocument_WithAggregationAndDeprecatedOn() {
+    license.setAbout("license2");
+    license.setCcDeprecatedOn(Date.from(Instant.parse("2022-06-01T12:22:55.888Z")));
+    license.setOdrlInheritFrom("OdrlInheritFrom");
+
+    licenseSolrCreator.addToDocument(solrInputDocument, license);
+
+    assertTrue(solrInputDocument.containsKey(EdmLabel.PROVIDER_AGGREGATION_CC_LICENSE.toString()) &&
```
stzanakis, line 42: Do you need to check this?
You are checking their values below, so I don't think it's needed to check first if they exist and then their value.
If they don't exist their value will be null therefore the equals assertion below will complain, right?
I believe this applies to the rest of the file as well.

stzanakis (commented):
All looks good, just one observation!

metis-indexing/src/test/java/eu/europeana/indexing/solr/property/LicenseSolrCreatorTest.java
stzanakis: I'm not sure I follow.
There is a call to insert the fields in the solr document and then the checks in order:
- Do the fields exist in the document?
- Do the fields(that we just checked their existence) have the expected values?

So I don't see what the benefit of that is, as opposed to check:
- Do the fields have the expected values? (this essentially checks their existence as well)

jeortizquan: This is a rough example to ilustrate my idea about the code that is why I came with this approach.

```
@Override
  public void addToDocument(SolrInputDocument doc, License license) {

    final boolean isAggregation = isAggregationResolver.test(license);

    final EdmLabel licenseLabel =
        isAggregation ? EdmLabel.PROVIDER_AGGREGATION_CC_LICENSE : EdmLabel.WR_CC_LICENSE;
    SolrPropertyUtils.addValue(doc, licenseLabel, **transform(license.getAbout())**);

    if (license.getCcDeprecatedOn() != null) {
      final EdmLabel deprecatedLabel = isAggregation ?
          EdmLabel.PROVIDER_AGGREGATION_CC_DEPRECATED_ON : EdmLabel.WR_CC_DEPRECATED_ON;
      final Date ccDeprecatedOnDate = new Date(license.getCcDeprecatedOn().getTime());
      doc.addField(deprecatedLabel.toString(), ccDeprecatedOnDate);
    }

    if (isAggregation) {
      SolrPropertyUtils.addValue(doc, EdmLabel.PROVIDER_AGGREGATION_ODRL_INHERITED_FROM,
          license.getOdrlInheritFrom());
    }
  }

  private String **transform**(String test) {
    if (test.length() > 5) {
      return test.substring(5);
    } else {
      return null;
    }
  }
```

jeortizquan: it would be ok then just to leave the value assertion then?

stzanakis: Well in that case the check you have will fail for a license value of less than 5 because you expect the field to be there but it is not added at all.

In that case, if we know our code does that, the test should check the expected value which would be:
e.g.
assertEquals("license2".substring(5), solrInputDocument.getFieldValue(EdmLabel.PROVIDER_AGGREGATION_CC_LICENSE.toString()));

And another test for the second case:
assertNull(solrInputDocument.getFieldValue(EdmLabel.PROVIDER_AGGREGATION_CC_LICENSE.toString()));

Either way, it's also okay how it is now!
What I'm saying is that we are doing a double check of essentially the same thing.

jeortizquan: I removed the double check. Simple is better as you pointed out. :smile:

PR #567, opened 2022-07-29
MET-4672: Meaningful response if a requested record is not found.

jochen-vermeulen:


jeortizquan (approved):
Tested local by removing and adding the `UNKNOWN ID`, the unit test covers when is included or not throws the expected exception in the ProxiesService.

PR #568, opened 2022-08-08
Feat/MET-4417_MET-4667 Indexing returns tier results

JoanaCMS:


metis-indexing/src/main/java/eu/europeana/indexing/IndexerImpl.java
```diff
@@ -86,6 +88,30 @@ public void index(InputStream recordInputStream, IndexingProperties indexingProp
...
+      throw new SetupRelatedIndexingException(
+              "Record redirect dao has not been initialized and performing redirects is requested");
+    }
+    LOGGER.info("Processing record to obtain tier calculations...");
+    final FullBeanPublisher publisher =
+            connectionProvider.getFullBeanPublisher(indexingProperties.isPreserveUpdateAndCreateTimesFromRdf());
+
+    final RDF rdfRecord = stringToRdfConverterSupplier.get().convertToRdf(recordContent);
+    TierResults result = preprocessRecord(rdfRecord, indexingProperties);
+    if (indexingProperties.isPerformRedirects()) {
```
jeortizquan, line 104: This name is a little confusing because one is with redirection and the other without in the publisher method
```indexingProperties.getDatasetIdsForRedirection()```

jeortizquan: maybe ```getDatasetIdsForPublication()``` or ```getDatasetIdsForPublishing()```

jeortizquan (commented):
It looks good, now an object returns tier and metadata information.

metis-indexing/src/main/java/eu/europeana/indexing/IndexerImpl.java
jeortizquan: Unit test for this new method perhaps?

JoanaCMS: The name of the method `.getDatasetIdsForRedirection` already existed before I changed this code. 
I'm not sure why the decision of this name but I would prefer to keep it since I don't know its purpose and usage.
I could work on the unit test, but I would have to create a new test class for this since it does not exist. It's true that we should try to create unit to have great code coverage, but I believe it is *not* part of the ticket to create a new test class for indexer

jochen-vermeulen: @JoanaCMS, @jeortizquan The method name `getDatasetIdsForRedirection` is a name that should probably not change. Redirection is a (fairly complex) functionality that was built before and the calling code needs to provide the dataset IDs for it, hence the method name.

There is an issue with this code that I think we should fix. The code in this method is almost identical to that in the private method a couple of lines further down. I think the new method should be calling that private method, and the private method changed to always return the `TierResults` instance.

PR #569, opened 2022-08-10
MET-4711-4722 Solve todo sections on date normalization

stzanakis:


metis-normalization/src/main/java/eu/europeana/normalization/dates/DateNormalizationResult.java
```diff
@@ -69,6 +69,7 @@ public String toString() {
         + edtfDate + "]";
   }
 
+
   // TODO: 22/07/2022 Double check if this is correct, it has not been tested since it's used in the untested generic property method
```
stzanakis, line 73: That's the method for generic property.
A range of only years(without a month and day), it should be considered valid.

From documentation:
`A normalised date is only accepted if it is considered reliable by verifying the following conditions (this verification is implemented in the method isCompleteDate() of the Match class):
The date contains the year, month and day.
If it is a range, the start and end dates both contain the year, month and day, or both contain only the year. For example, ‘1980-05-06/1980-06-20’ or ‘1980/1981’.`

metis-normalization/src/main/java/eu/europeana/normalization/dates/DateNormalizationResult.java
```diff
@@ -95,6 +96,7 @@ private boolean isEitherDateNonPrecise(EdtfDatePart startEdtfDatePart, EdtfDateP
         || startEdtfDatePart.getYearPrecision() != null || endEdtfDatePart.getYearPrecision() != null;
   }
 
+
   // TODO: 22/07/2022 Is this calculation correct??
```
stzanakis, line 100: So this computation should be with "equals" to null for both the days and months(instead of not equals as it is now).
The order is not significant.

metis-normalization/src/main/java/eu/europeana/normalization/dates/cleaning/CleanOperation.java
```diff
@@ -36,6 +36,7 @@ public enum CleanOperation {
       Matcher::find, matcher -> matcher.replaceAll("$1"), s -> true),
   CIRCA(compile("^\\s*(circa|CA\\.?|C\\.)\\s*", Pattern.CASE_INSENSITIVE),
       Matcher::find, matcher -> matcher.replaceAll(""), s -> true),
+
   // TODO: 22/07/2022 Need to fix this regex, it is reported as dangerous
```
stzanakis, line 40: For metis to investigate

metis-normalization/src/main/java/eu/europeana/normalization/dates/edtf/EdtfDatePart.java
```diff
@@ -18,6 +18,7 @@ public class EdtfDatePart implements Serializable {
   private boolean uncertain;
   private boolean approximate;
   private boolean unknown;
+
   // TODO: 25/07/2022 What is unspecified? It's not in the documentation, is it the same as unknown?
```
stzanakis, line 22: It relates to edtf format. This case is if there is an edtf formatted date already in the document and we are trying to parse it. This field will indicate if that date is an "unspecified" one.

Examples:
`1900/` (unspecified end)
`1900/?` (unknown)

metis-normalization/src/main/java/eu/europeana/normalization/dates/edtf/EdtfParser.java
```diff
@@ -12,6 +12,7 @@
  */
 public class EdtfParser {
 
+
   // TODO: 19/07/2022 Simplify regex by potentially splitting it
```
stzanakis, line 16: For Metis to investigate

metis-normalization/src/main/java/eu/europeana/normalization/dates/edtf/EdtfTimePart.java
```diff
@@ -61,6 +61,7 @@ public String toString() {
     final DecimalFormat decimalFormat = new DecimalFormat("00");
     final DecimalFormat millisFormat = new DecimalFormat("000");
     StringBuilder stringBuilder = new StringBuilder();
+
     // TODO: 25/07/2022 Original checks where checking for hour, minute, second being non-zero.
```
stzanakis, line 65: This seems okay.

metis-normalization/src/main/java/eu/europeana/normalization/dates/edtf/EdtfValidator.java
```diff
@@ -42,6 +42,7 @@ private static boolean validateNotInFuture(AbstractEdtfDate edtfDate) {
     return validateIntervalNotInFuture((IntervalEdtfDate) edtfDate);
   }
 
+
   // TODO: 20/07/2022 It checks interval of date parts but not time parts??
```
stzanakis, line 46: This is valid. It has been decided that the time part is not relevant during validation.
The Instant date contains the time part though, which might not be used elsewhere.

metis-normalization/src/main/java/eu/europeana/normalization/dates/edtf/EdtfValidator.java
```diff
@@ -72,6 +73,7 @@ private static boolean validateInterval(IntervalEdtfDate intervalEdtfDate) {
   }
 
   private static boolean validateSpecificIntervalDates(EdtfDatePart startDatePart, EdtfDatePart endDatePart) {
+
     // TODO: 20/07/2022 Should we be using the java.time classes Year, YearMonth, LocalDate etc?
```
stzanakis, line 77: This can be relevant, to be investigated.

metis-normalization/src/main/java/eu/europeana/normalization/dates/edtf/EdtfValidator.java
```diff
@@ -72,6 +73,7 @@ private static boolean validateInterval(IntervalEdtfDate intervalEdtfDate) {
...
     // TODO: 20/07/2022 No check for null year but we check for null month and day?
```
stzanakis, line 78: For more accurate checks we should check a null check for the year as well.
That means that we probably need to throw an exception as well.

metis-normalization/src/main/java/eu/europeana/normalization/dates/edtf/EdtfValidator.java
```diff
@@ -90,6 +92,7 @@ private static boolean validateSpecificIntervalDates(EdtfDatePart startDatePart,
   }
 
   private static Integer adjustYearWithPrecision(Integer year, YearPrecision yearPrecision) {
+
     // TODO: 25/07/2022 Is this precision adjustment correct? How should the rounding be for positive and negatives?
```
stzanakis, line 96: This is accurate and expected.
We are discarding the right most digits and we don't perform flooring.
Write a good javadoc explanation for this method.

metis-normalization/src/main/java/eu/europeana/normalization/dates/edtf/EdtfValidator.java
```diff
@@ -111,6 +114,7 @@ private static boolean validateInstant(InstantEdtfDate instantEdtfDate, boolean
     if (validateDatePart(edtfDatePart)) {
       EdtfTimePart edtfTimePart = instantEdtfDate.getEdtfTimePart();
       if (validateTimePart(edtfTimePart)) {
+
         // TODO: 20/07/2022 Does this mean that if it's not standalone, it is then allowed to have both null/unknown??
```
stzanakis, line 118: For non-standalone(interval dates) this check is not required since the date can be empty(which is considered valid).

metis-normalization/src/main/java/eu/europeana/normalization/dates/edtf/EdtfValidator.java
```diff
@@ -181,6 +185,7 @@ private static boolean validateIntervalNotInFuture(IntervalEdtfDate intervalEdtf
     return validateInstantNotInFuture(intervalEdtfDate.getStart()) && validateInstantNotInFuture(intervalEdtfDate.getEnd());
   }
 
+
   // TODO: 20/07/2022 This only calculates years and not other parts of the date.
```
stzanakis, line 189: We should be checking also the other parts of the date, not just the year.

metis-normalization/src/main/java/eu/europeana/normalization/dates/edtf/InstantEdtfDate.java
```diff
@@ -101,6 +101,7 @@ && getEdtfDatePart().getYear() < THRESHOLD_4_DIGITS_YEAR) {
           firstDay.getEdtfDatePart().setDay(1);
         }
       }
+
       // TODO: 25/07/2022 What about > THRESHOLD_4_DIGITS_YEAR??
```
stzanakis, line 105: The part where > THRESHOLD_4_DIGITS_YEAR is not possible because it's in the future, so we don't have to check it.
Verify though that the contents of this class are always considered valid before the call of this method.

metis-normalization/src/main/java/eu/europeana/normalization/dates/edtf/InstantEdtfDate.java
```diff
@@ -158,6 +159,7 @@ public void removeTime() {
 
   public Integer getCentury() {
     final int century;
+
     // TODO: 25/07/2022 getEdtfDatePart() or getEdtfDatePart().getYear() might be null??
```
stzanakis, line 163: Better to check both for nullity and if they are null we then throw an exception.

metis-normalization/src/main/java/eu/europeana/normalization/dates/extraction/dateextractors/DcmiPeriodDateExtractor.java
```diff
@@ -16,6 +16,7 @@
  */
 public class DcmiPeriodDateExtractor implements DateExtractor {
 
+
   // TODO: 22/07/2022 Need to fix this regex, it is reported as dangerous
```
stzanakis, line 20: For Metis to investigate.

metis-normalization/src/main/java/eu/europeana/normalization/normalizers/DatesNormalizer.java
```diff
@@ -231,6 +231,7 @@ private void normalizeElement(Document document, Element element, Namespace.Elem
     report.increment(this.getClass().getSimpleName(), ConfidenceLevel.CERTAIN);
   }
 
+
   // TODO: 25/07/2022 This can be made private
```
stzanakis, line 235: For Metis to check.

metis-normalization/src/main/java/eu/europeana/normalization/normalizers/DatesNormalizer.java
```diff
@@ -240,6 +241,7 @@ public DateNormalizationResult normalizeDateProperty(String input) {
         this::noMatchIfValidAndTimeOnly);
   }
 
+
   // TODO: 21/07/2022 To check. This method does not do dates switching like the other method
```
stzanakis, line 245: Yes that is correct. We don't need to do the switching, we are stricter on generic properties.

metis-normalization/src/main/java/eu/europeana/normalization/normalizers/DatesNormalizer.java
```diff
@@ -240,6 +241,7 @@ public DateNormalizationResult normalizeDateProperty(String input) {
...
   // TODO: 21/07/2022 Did not have unit tests originally.
   // TODO: 25/07/2022 This can be made private
```
stzanakis, line 247: For Metis to check.

metis-normalization/src/main/java/eu/europeana/normalization/normalizers/DatesNormalizer.java
```diff
@@ -240,6 +241,7 @@ public DateNormalizationResult normalizeDateProperty(String input) {
         this::noMatchIfValidAndTimeOnly);
   }
 
+
   // TODO: 21/07/2022 To check. This method does not do dates switching like the other method
   // TODO: 21/07/2022 Did not have unit tests originally.
```
stzanakis, line 246: Need to write unit tests. Might require input from @nfreire

metis-normalization/src/main/java/eu/europeana/normalization/normalizers/DatesNormalizer.java
```diff
@@ -286,6 +288,7 @@ private DateNormalizationResult normalizeProperty(
   private void noMatchIfValidAndTimeOnly(DateNormalizationResult dateNormalizationResult) {
     if (dateNormalizationResult.getDateNormalizationExtractorMatchId() != DateNormalizationExtractorMatchId.INVALID
         && dateNormalizationResult.getEdtfDate().isTimeOnly()) {
+
       // TODO: 21/07/2022 In the result only the match id is declared NO_MATCH but the contents are still present in the object. Is that okay?
```
stzanakis, line 292: This is okay, but as before we don't expect to have just the time and not the date.
To be considered on how to structure this, if at all to be changed.

metis-normalization/src/main/java/eu/europeana/normalization/normalizers/DatesNormalizer.java
```diff
@@ -304,6 +307,7 @@ private DateNormalizationResult normalizeInput(List<DateExtractor> dateExtractor
       dateNormalizationResult = normalizeInput(dateExtractors, cleanedInput.getCleanedValue());
       if (dateNormalizationResult != null) {
         dateNormalizationResult.setCleanOperationMatchId(cleanedInput.getCleanOperation());
+
         // TODO: 21/07/2022 Perhaps this should be done differently, because it pollutes the map of the extractor and its id
```
stzanakis, line 311: This can be structured differently if we want to, so we can update it as such.

metis-normalization/src/main/java/eu/europeana/normalization/normalizers/DatesNormalizer.java
```diff
@@ -373,10 +377,10 @@ private static String sanitizeCharacters(String input) {
   }
 
   private void appendTimespanEntity(Document document, AbstractEdtfDate edtfDate, String timespanId) {
+
     // TODO: 09/08/2022 All the element prefixes below are searched first and if not found then the suggested prefix is added.
```
stzanakis, line 381: For Metis to investigate.

metis-normalization/src/main/java/eu/europeana/normalization/normalizers/DatesNormalizer.java
```diff
@@ -417,6 +421,7 @@ private void appendTimespanEntity(Document document, AbstractEdtfDate edtfDate,
                                    .map(InstantEdtfDate::getCentury).orElse(null);
     Integer endCentury = Optional.ofNullable(lastDay)
                                  .map(InstantEdtfDate::getCentury).orElse(null);
+
     // TODO: 25/07/2022 What if both are null, won't the 'for' loop below fail? Is there always at least one century?
```
stzanakis, line 425: At this point, everything should be valid so that is not possible.
For a sanity check perhaps we can check for that case and throw an exception if that happens.
Or the date objects, as before, are by their instantiation validated.

PR #570, opened 2022-08-12
MET-4673 Added UISClient from Ecloud, to OrchestratorConfig, ProxiesService, ProxiesController.

adolfopeixinho:
Unit tests missing for UISClient


metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/service/ProxiesService.java
```diff
@@ -394,6 +399,52 @@ public RecordsResponse getListOfFileContentsFromPluginExecution(MetisUserView me
...
+   * resource</li>
+   * <li>{@link eu.europeana.metis.exception.UserUnauthorizedException} if the user is not
+   * authorized to perform this task</li>
+   * <li>{@link eu.europeana.metis.core.exceptions.NoWorkflowExecutionFoundException} if no workflow
+   * execution exists for the provided identifier</li>
+   * </ul>
+   */
+  public Record lookupIdFromUISClient(MetisUserView metisUserView, String workflowExecutionId , ExecutablePluginType pluginType, String searchId )
+      throws GenericMetisException {
+
```
adolfopeixinho, line 422: Don't know if this makes sense. But to call the `getRecord` method it's required an instance of some sort of plugin, so I added that. Which makes `authorizer` call superfluous

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/service/ProxiesService.java
```diff
@@ -394,6 +399,52 @@ public RecordsResponse getListOfFileContentsFromPluginExecution(MetisUserView me
...
+    final CloudId cloudId;
+    try {
+      cloudId = uisClient.getCloudId(searchId,searchId);
+
+
+      if (cloudId != null) {
+        return Optional.ofNullable(getRecord(executionAndPlugin.getRight(), cloudId.getId())).orElse(null);
+      }
+    } catch (CloudException e) {
+      throw new GenericMetisException(String.format("Failed to lookup cloudId for searchId: %s", searchId), e);
```
jochen-vermeulen, line 444: Should throw an instance of `ExternalTaskException` (which is a subclass of `GenericMetisException`).

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/service/ProxiesService.java
```diff
@@ -394,6 +399,52 @@ public RecordsResponse getListOfFileContentsFromPluginExecution(MetisUserView me
...
+   * <ul>
+   * <li>{@link MCSException} if an error occurred while retrieving the records from the external
+   * resource</li>
+   * <li>{@link eu.europeana.metis.exception.UserUnauthorizedException} if the user is not
+   * authorized to perform this task</li>
+   * <li>{@link eu.europeana.metis.core.exceptions.NoWorkflowExecutionFoundException} if no workflow
+   * execution exists for the provided identifier</li>
+   * </ul>
+   */
+  public Record lookupIdFromUISClient(MetisUserView metisUserView, String workflowExecutionId , ExecutablePluginType pluginType, String searchId )
```
jochen-vermeulen, line 420: The method name could be better chosen, as it does not just lookup an ID. Perhaps something like `lookupRecordById`?

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/service/ProxiesService.java
jochen-vermeulen: Makes sense indeed. Please remove the two commented-out lines.

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/service/ProxiesService.java
```diff
@@ -394,6 +399,52 @@ public RecordsResponse getListOfFileContentsFromPluginExecution(MetisUserView me
...
+          .format("No executable plugin of type %s found for workflowExecution with id: %s",
+              pluginType.name(), workflowExecutionId));
+    }
+    // Authorize
+//    authorizer.authorizeReadAllDatasets(metisUserView);
+
+    // lookup for eCloudId.
+    final CloudId cloudId;
+    try {
+      cloudId = uisClient.getCloudId(searchId,searchId);
```
jochen-vermeulen, line 437: I think there may be the case that the searchId is ALREADY a cloud ID. I think that in this case `uisClient.getCloudId()` would return null, but in that case, you should set `cloudId = searchId` and try with that.
Once deployed, we should test this.

metis-core/metis-core-service/src/test/java/eu/europeana/metis/core/service/TestProxiesService.java
```diff
@@ -333,6 +337,8 @@ void getExternalTaskStatistics() throws Exception {
     assertSame(recordStatistics, result);
   }
 
+  // TODO: add tests for lookupIdFromUISClient
```
jochen-vermeulen, line 340: Flagging that some unit tests are needed here.

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/service/ProxiesService.java
adolfopeixinho: I don't think this makes much sense. But perhaps I'm missing something.
`    CloudId cloudId = null;
    try {
      cloudId = uisClient.getCloudId(searchId,searchId);
      if (cloudId != null) {
        return Optional.ofNullable(getRecord(executionAndPlugin.getRight(), cloudId.getId())).orElse(null);
      }else {
        cloudId.setId(searchId);
        return Optional.ofNullable(getRecord(executionAndPlugin.getRight(), cloudId.getId())).orElse(null);
      }`

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/service/ProxiesService.java
```diff
@@ -394,9 +400,64 @@ public RecordsResponse getListOfFileContentsFromPluginExecution(MetisUserView me
...
+        ecloudId = searchId;
+      } else {
+        // Some other connectivity issue.
+        throw new ExternalTaskException(
+            String.format("Failed to lookup cloudId for searchId: %s", searchId), e);
+      }
+    }
+
+    // Try to retrieve the record.
+    return Optional.ofNullable(getRecord(executionAndPlugin.getRight(), ecloudId)).orElse(null);
```
JoanaCMS, line 457: Why do `.orElse(null)`? Doesn't `Optional.ofNullable` already handles in case `getRecord` returns null?
This makes it redundant. Either simply return `getRecord` as it is or just use the `Optional.ofNullable`

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/service/ProxiesService.java
```diff
@@ -394,9 +400,64 @@ public RecordsResponse getListOfFileContentsFromPluginExecution(MetisUserView me
     // Done.
     return new RecordsResponse(records);
   }
+  /**
+   * Get a record from the external resource based on o searchId, workflow execution and {@link PluginType}.
+   *
+   * @param metisUserView the user wishing to perform this operation
+   * @param workflowExecutionId the execution identifier of the workflow
+   * @param pluginType the {@link ExecutablePluginType} that is to be located inside the workflow
+   * @param searchId the searchId of the record we wish to obtain
```
JoanaCMS, line 409: What is the `searchId`? It is not clear to me what this is supposed to be, the `recordId`? `europeanaId`? `anyOtherId`?
If we are *not* supposed to know then perhaps change the name to `idToSearchWith`?

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/service/ProxiesService.java
jochen-vermeulen: Done.

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/service/ProxiesService.java
jochen-vermeulen: Done.

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/service/ProxiesService.java
jochen-vermeulen: Fixed.

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/service/ProxiesService.java
jochen-vermeulen: Fixed.

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/service/ProxiesService.java
jochen-vermeulen: Ah, I think what Adolfo meant here was to check for the nullity of `recordId` itself. I fixed it.

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/service/ProxiesService.java
jochen-vermeulen: OK, changed it to `idToSearch`. Is that better?

PR #573, opened 2022-08-30
feat/MET-4720_MET-4761 add skos notation date normalization data to mongo

jeortizquan:
**remark**: we will need to wait until corelib makes their official release. The pom needs to run with 2.15.8. At the moment  works with 2.15.8-SNAPSHOT version of corelib.

### DIFF
--- a/metis-indexing/src/main/java/eu/europeana/indexing/mongo/TimespanUpdater.java
+++ b/metis-indexing/src/main/java/eu/europeana/indexing/mongo/TimespanUpdater.java
@@ -23,5 +23,6 @@ public class TimespanUpdater extends AbstractIsolatedEdmEntityUpdater<TimespanIm
     propertyUpdater.updateMap("isPartOf", TimespanImpl::getIsPartOf);
     propertyUpdater.updateMap("dctermsHasPart", TimespanImpl::getDctermsHasPart);
     propertyUpdater.updateArray("owlSameAs", TimespanImpl::getOwlSameAs);
+    propertyUpdater.updateMap("notation", TimespanImpl::getSkosNotation);
   }
 }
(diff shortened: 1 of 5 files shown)


pom.xml
```diff
@@ -137,7 +137,7 @@
     <version.com.zoho.crm.java-sdk>3.0.1</version.com.zoho.crm.java-sdk>
     <version.apache.commons.text>1.9</version.apache.commons.text>
     <version.commonscompress>1.21</version.commonscompress>
-    <version.corelib>2.15.3</version.corelib>
+    <version.corelib>2.15.8</version.corelib>
```
jochen-vermeulen, line 140: I suppose that for now we'll have to use the snapshot version. Right now the build is broken.

metis-indexing/src/main/java/eu/europeana/indexing/mongo/TimespanUpdater.java
```diff
@@ -23,5 +23,6 @@ protected void update(MongoPropertyUpdater<TimespanImpl> propertyUpdater) {
     propertyUpdater.updateMap("isPartOf", TimespanImpl::getIsPartOf);
     propertyUpdater.updateMap("dctermsHasPart", TimespanImpl::getDctermsHasPart);
     propertyUpdater.updateArray("owlSameAs", TimespanImpl::getOwlSameAs);
+    propertyUpdater.updateMap("notation", TimespanImpl::getSkosNotation);
```
jochen-vermeulen, line 26: In order to get the skos notation from `TimespanImpl` you will first need to set it. See `TimespanFieldInput`.

pom.xml
jeortizquan: Updated to the snapshot version.

metis-indexing/src/main/java/eu/europeana/indexing/mongo/TimespanUpdater.java
jeortizquan: I see. Updated the ```TimespanFieldInput``` class

PR #584, opened 2022-10-21
MET-4785 HttpRecordIterator provides extracted directory

JoanaCMS:


metis-harvesting/src/main/java/eu/europeana/metis/harvesting/http/HttpHarvesterImpl.java
```diff
@@ -31,6 +31,7 @@
 import org.apache.commons.io.FileUtils;
 import org.apache.commons.io.FilenameUtils;
 import org.apache.commons.io.IOUtils;
+import org.apache.commons.lang3.StringUtils;
```
jeortizquan, line 34: Is not used. just remove.

jeortizquan (commented):
We are good this part, just need to remove unused imports.

metis-harvesting/src/main/java/eu/europeana/metis/harvesting/http/HttpHarvesterImpl.java
JoanaCMS: Removed

PR #586, opened 2022-10-26
debt/MET-4892_MET-4898 Allow setting mongo pool size

JoanaCMS:


metis-indexing/src/main/java/eu/europeana/indexing/IndexingSettings.java
```diff
@@ -96,6 +96,16 @@ public void setMongoApplicationName(String applicationName) {
     this.mongoProperties.setApplicationName(applicationName);
   }
 
+  /**
+   * Set the max number of connections pool size in mongo. Can be null, in which case a default generic
+   * application name is to be used.
```
stzanakis, line 101: It seems the javadoc is misleading

stzanakis (changes requested):

metis-indexing/src/main/java/eu/europeana/indexing/IndexingSettings.java
JoanaCMS: Oops, I'll fix that

PR #589, opened 2022-11-15
feat/MET-4865 Docker k8s metis-authentication

stzanakis:


metis-authentication/metis-authentication-common/src/main/java/eu/europeana/metis/authentication/utils/MetisZohoOAuthPSQLHandler.java
```diff
@@ -39,17 +37,22 @@ public class MetisZohoOAuthPSQLHandler implements TokenStore {
...
   private static String clientSecret;
 
-  static {
-    org.hibernate.cfg.Configuration configuration = new org.hibernate.cfg.Configuration();
-    configuration.addAnnotatedClass(MetisZohoOAuthToken.class);
-    configuration.configure();
-    ServiceRegistry serviceRegistry = new StandardServiceRegistryBuilder()
-        .applySettings(configuration.getProperties()).build();
-    sessionFactory = configuration.buildSessionFactory(serviceRegistry);
+  //  static {
```
JoanaCMS, line 44: Are these comments needed? If not please remove
Otherwise please add a descriptive comment about this

metis-authentication/metis-authentication-rest/README.MD
```diff
@@ -0,0 +1,98 @@
...
+This can be applied with an environment variable pointing to the appropriate .properties file
+`SPRING_CONFIG_LOCATION=file:///path/to/file.properties`
+It can be applied in Intellij from the:
+> Edit configurations -> Modify Options -> Environment Variables -> Add the SPRING_CONFIG_LOCATION variable
+
+### Environment variables
+
+Alternatively an environment variables file can be created.  
+Spring boot accepts environment variables that can be created using the same names as in the `.properties` file,
+with their names upper-cased and punctuation characters replaced with "_".
```
JoanaCMS, line 20: A small visualization example would be nice

metis-authentication/metis-authentication-rest/src/main/java/eu/europeana/metis/authentication/rest/config/ElasticConfig.java
```diff
@@ -0,0 +1,45 @@
...
+   *
+   * @return map of elastic apm agent properties
+   */
+  public Map<String, String> getApm() {
+    return this.apmProperties;
+  }
+
+  @Override
+  public String toString() {
+    return NEW_LINE_TAB + apmProperties.entrySet().stream().map(Objects::toString)
```
JoanaCMS, line 36: Why add the `NEW_LINE_TAB` in front of the string? Is it needed for some kind of setup?

metis-authentication/metis-authentication-common/src/main/java/eu/europeana/metis/authentication/utils/MetisZohoOAuthPSQLHandler.java
stzanakis: Indeed, leftovers I forgot to remove

metis-authentication/metis-authentication-rest/src/main/java/eu/europeana/metis/authentication/rest/config/ElasticConfig.java
stzanakis: You are right!
I copied this from sandbox in fact but I didn't think about it too much.
This is used for the debugging log below. I've updated it to a better approach.

metis-authentication/metis-authentication-rest/README.MD
stzanakis: I try to avoid adding binaries such as images to git.
If we want to do this, it's probably better to use git-lfs.
I can have a look at it today.

stzanakis: Or do you mean an example of how an environment variable would look like in this case?

JoanaCMS: How an environment variable would look like

stzanakis: I see, thanks for the clarification! I've added an example.

PR #594, opened 2023-01-16
Feat/met 4866 dockerize metis dereference

stzanakis:


metis-dereference/metis-dereference-rest/docker-compose.yml
```diff
@@ -0,0 +1,28 @@
...
+  metis-dereference-local:
+    image: europeana/metis-dereference:develop
+    container_name: metis-dereference-local
+    build:
+      context: ./
+      dockerfile: Dockerfile
+    ports:
+      - '8080:8080'
+    environment:
+      MONGO_HOSTS: metis-core-mongo
```
JoanaCMS, line 22: If I understand correctly, we create a container for mongo with name `metis-dereference-mongo` but then within `metis-dereference-local` we set up as a mongo host `metic-core-mongo`, shouldn't we connect to the previously mentioned container `metis-dereference-mongo`?

metis-dereference/metis-dereference-rest/src/main/java/eu/europeana/metis/dereference/rest/config/ApplicationConfiguration.java
```diff
@@ -33,59 +35,57 @@
...
+   * @param propertiesHolder the object that holds all boot configuration values
+   * @throws TrustStoreConfigurationException if the configuration of the truststore failed
    */
-  @Override
-  public void afterPropertiesSet() {
-    if (socksProxyEnabled) {
-      new SocksProxy(socksProxyHost, socksProxyPort, socksProxyUsername, socksProxyPassword).init();
+  public ApplicationConfiguration(ConfigurationPropertiesHolder propertiesHolder) throws TrustStoreConfigurationException {
+    mongoClientEntity = ApplicationConfiguration.initializeApplication(propertiesHolder);
+    mongoClientVocabulary = ApplicationConfiguration.initializeApplication(propertiesHolder);
```
JoanaCMS, line 55: Two different instances of "the same" MongoClient, is each instances used for something different?

metis-dereference/metis-dereference-rest/src/main/java/eu/europeana/metis/dereference/rest/config/SwaggerConfig.java
```diff
@@ -1,46 +1,30 @@
...
-import springfox.documentation.spi.DocumentationType;
-import springfox.documentation.spring.web.plugins.Docket;
-import springfox.documentation.swagger2.annotations.EnableSwagger2;
 
 /**
  * Config for Swagger documentation
  */
 @Configuration
-@EnableSwagger2
 public class SwaggerConfig {
```
JoanaCMS, line 13: Since Swagger will not be used, perhaps would be best to rename this class name to `OpenAPIConfig`

metis-dereference/metis-dereference-rest/docker-compose.yml
stzanakis: Yes absolutely! I fixed it.

metis-dereference/metis-dereference-rest/src/main/java/eu/europeana/metis/dereference/rest/config/ApplicationConfiguration.java
stzanakis: One is for the entities and the other one for the vocabularies.
I left it as such since it was there.
But it seems reasonable to fix now so I've reduced it to one client.

metis-dereference/metis-dereference-rest/src/main/java/eu/europeana/metis/dereference/rest/config/SwaggerConfig.java
stzanakis: Actually not.
Swagger is still used(check at the import packages above)
It is just the specification that changes to the newer model.
So I would keep it as such

PR #596, opened 2023-01-23
feat/MET-4866_Dockerize_metis_dereference_fixes

stzanakis:


metis-dereference/metis-dereference-rest/src/main/java/eu/europeana/metis/dereference/rest/controller/DereferencingController.java
```diff
@@ -50,14 +48,7 @@ public DereferencingController(DereferenceService dereferenceService) {
...
   @ResponseBody
-  @Operation(description = "Dereference a URI",
-      responses = {
-          @ApiResponse(responseCode = "200", content = {
-              @Content(
-                  mediaType = MediaType.APPLICATION_JSON_VALUE,
-                  schema = @Schema(implementation = EnrichmentResultList.class))
-          }),
-      })
+  @Operation(description = "Dereference a URI", responses = {@ApiResponse(responseCode = "200")})
```
JoanaCMS, line 51: Shouldn't we write to return results both in XML and JSON? Or like this it returns both by default?
The same comment applies for the rest of the changes

stzanakis: Yes indeed, if not mentioned it captures what it reads from the other annotations(as before essentially).

PR #598, opened 2023-01-26
MET-2018_MET-5051 add dereference entity found xslt produced no contextual class

jeortizquan:


metis-dereference/metis-dereference-service/src/test/java/eu/europeana/metis/dereference/service/MongoDereferenceServiceTest.java
```diff
@@ -150,17 +150,32 @@ void testDereference_InvalidUrl() throws JAXBException, URISyntaxException {
...
-    assertEquals(DereferenceResultStatus.ENTITY_FOUND_XML_XLT_ERROR, result.getDereferenceStatus());
+    assertEquals(DereferenceResultStatus.ENTITY_FOUND_XML_XSLT_ERROR, result.getDereferenceStatus());
+  }
+
+  @Test
+  void testDereference_XmlXsltProduceNoContextualClass() throws JAXBException, URISyntaxException {
+    final String entityId = "http://www.yso.fi/onto/yso/p105069";
+
+    // Mock the service
+    doReturn(new DereferenceResultWrapper(DereferenceResultStatus.ENTITY_FOUND_XML_XSLT_PRODUCE_NO_CONTEXTUAL_CLASS))
```
stzanakis, line 172: You are testing the service, why mock it also?
Doesn't this defeat the purpose of testing the service?
Normally all calls in the service layer should not be mocked and calls to "other" components etc. DAOs, http calls should be mocked instead.
I see the other tests are similar though, so perhaps we could move ahead(given also the timing) with a followup ticket that will be more thorough in the testing implememntation.
What do you say?

jeortizquan: Yes, I agree, we can think about refactoring the code, due some of the methods changed are inside of a private call that calls another private method. Here is the follow up ticket created. https://europeana.atlassian.net/browse/MET-5065

PR #603, opened 2023-03-15
Feat/MET-4872 contract for and mongo solr only indexing

jeortizquan:


metis-common/metis-common-base/src/main/java/eu/europeana/metis/common/SettingsHolder.java
```diff
@@ -0,0 +1,4 @@
+package eu.europeana.metis.common;
+
```
JoanaCMS, line 2: A suggestion to take advantage of this interface:
I  see that `MongoIndexingSettings` has `MongoProperties` and `SolrIndexingSettings` has `SolrProperties`, perhaps you could create a `DatabaseProperties` interface which `MongoProperties` and `SolrProperties` implement
Then, this interface `SettingsHolder` could have a method called `getDatabaseProperties`, which as a consequence `MongoIndexingSettings` and `SolrIndexingSettings` will implement that method and return their specific properties 
This is just a suggestion of course, what do you think?

metis-indexing/src/main/java/eu/europeana/indexing/FullBeanPublisher.java
```diff
@@ -168,52 +169,125 @@ public void publish(RdfWrapper rdf, Date recordDate, List<String> datasetIdsToRe
...
+   * <li>{@link IndexerRelatedIndexingException} In case an error occurred during publication.</li>
+   * <li>{@link SetupRelatedIndexingException} in case an error occurred during indexing setup</li>
+   * <li>{@link RecordRelatedIndexingException} in case an error occurred related to record
+   * contents</li>
+   * </ul>
+   */
+  public void publishSolr(RdfWrapper rdf, Date recordDate) throws IndexingException {
+    final FullBeanImpl fullBean = convertRDFToFullBean(rdf);
+    if (!preserveUpdateAndCreateTimesFromRdf){
+      final Date createdDate = recordDate;
```
JoanaCMS, line 222: Is this comment supposed to explain what is happening? Then could you remove the `TODO`? it could be misleading that something is missing here
Otherwise, if this comment is not needed at all, then just remove it please

jeortizquan: actually this TODO is still valid, I created the followup ticket in jira. https://europeana.atlassian.net/browse/MET-5160

jeortizquan: it will be implemented in this feature. will create a new branch, the rebase polluted the PR.

jeortizquan:
After rebase the branch poluted.

PR #608, opened 2023-04-05
bug/MET-4860 remove sonar notification for NullPointerException

jeortizquan:


metis-indexing/src/main/java/eu/europeana/indexing/mongo/property/MongoPropertyUpdaterImpl.java
```diff
@@ -273,22 +273,21 @@ public void updateWebResourceMetaInfo(Function<T, WebResourceMetaInfo> getter,
...
-    final UpdateOperator[] extraUpdateOperators = this.updateOperators
-        .subList(1, this.updateOperators.size()).toArray(UpdateOperator[]::new);
-    final Update<T> update = queryCreator.get().update(firstUpdateOperator, extraUpdateOperators);
+
+    final UpdateOperator[] extraUpdateOperators = this.updateOperators.toArray(UpdateOperator[]::new);
     final UpdateOptions updateOptions = new UpdateOptions().upsert(true).multi(true);
+
     try {
-      retryableExternalRequestForNetworkExceptions(() -> update.execute(updateOptions));
+      retryableExternalRequestForNetworkExceptions(() -> queryCreator.get().update(updateOptions, extraUpdateOperators));
```
stzanakis, line 287: I do not this the .get() should be in the try block, same for the catch block.
The try-catch here is meant for duplicate key exceptions and not other type of failures.

stzanakis (changes requested):

metis-indexing/src/main/java/eu/europeana/indexing/mongo/property/MongoPropertyUpdaterImpl.java
jeortizquan: oh I see. yes. modified to be outside the try-block.

PR #611, opened 2023-05-15
bug/MET-5110_fix_entity_status_when_no_entity_found_for_existing_vocabulary

jeortizquan:


metis-enrichment/metis-enrichment-client/src/main/java/eu/europeana/enrichment/rest/client/dereference/DereferencerImpl.java
```diff
@@ -33,315 +37,321 @@
...
+
+    private DereferencedEntities dereferenceOwnEntities(Set<ReferenceTerm> resourceIds,
+                                                        HashSet<Report> reports,
+                                                        Class<? extends AboutType> classType) {
+        if (entityResolver == null) {
+            return new DereferencedEntities(Collections.emptyMap(), new HashSet<>());
+        }
+        try {
+            Map<ReferenceTerm, List<EnrichmentBase>> result = new HashMap<>();
+            entityResolver.resolveById(resourceIds).forEach((key, value) -> result.put(key, List.of(value)));
```
jochen-vermeulen, line 249: Here you probably need to check against the europeana URL. 
Something like
```
var ownEntities = resourceIds.stream()
    .filter(id->EntityResolver.europeanaLinkPattern.matcher(id.getReference().toString()).matches())
    .collect(Collectors.toSet());
entityResolver.resolveById(ownEntities).forEach((key, value) -> result.put(key, List.of(value)));
ownEntities.stream().filter(id->!result.containsKey(id))
    .forEach(notFoundOwnId ->setDereferenceStatusInReport(notFoundOwnId.getReference().toString(), 
        reports, DereferenceResultStatus.UNKNOWN_EUROPEANA_ENTITY));
```

metis-dereference/metis-dereference-service/src/main/java/eu/europeana/metis/dereference/service/MongoDereferenceService.java
```diff
@@ -33,379 +43,398 @@
...
+                } else {
+                    result = new HashMap<>();
+                    result.put(resourceId, new ImmutablePair<>(resource.getEnrichmentBase(),
+                            resource.getDereferenceResultStatus()));
+                }
+                // Done
+                dereferenceResult = new DereferenceResult(
+                        result.values().stream().map(Pair::getLeft).collect(Collectors.toList()),
+                        result.values().stream().map(Pair::getRight).filter(Objects::nonNull).findFirst()
+                                .orElse(DereferenceResultStatus.UNKNOWN_EUROPEANA_ENTITY));
```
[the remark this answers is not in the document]
jeortizquan: I see. This is the case when no status was found after doing all processing so I wanted to report UKNOWN or not DEFINED. As you mention looking it from the perspective of the whole process we can use `DereferenceResultStatus.SUCCESS` then.

metis-enrichment/metis-enrichment-client/src/main/java/eu/europeana/enrichment/rest/client/dereference/DereferencerImpl.java
jeortizquan: I understand what this means now, because of the previous way the status was computed never thought to explicitly validate it. will proceed to update the code.

PR #613, opened 2023-05-24
bug/MET-5054 fix connectivity dereference from WARNING to ERROR types

jeortizquan:
Add network exceptions handling for dereference report


stzanakis (changes requested):
Building locally the client the dereference keeps repeating in a loop for very long.
And that is also visible here above with the build taking "Successful in 28m".
This is not acceptable I would say.
We have this issue with the retries in general, so for now I would say it is better to not use the test for this cases and create a ticket(perhaps we already have one), to make the retries more flexible for tests, so that we can reduce to the minimum the wait time between retries(for tests only).

PR #617, opened 2023-06-28
Feat/MET-5361 Capture tier calculation results for sandbox

JoanaCMS:


metis-indexing/src/test/java/eu/europeana/indexing/tiers/view/MediaResourceTechnicalMetadataBuilderTest.java
```diff
@@ -44,6 +44,8 @@ void objectCreationTest() {
     assertThrows(NullPointerException.class, mediaResourceTechnicalMetadataBuilder::build);
     final MediaTier mediaTier = MediaTier.T0;
     mediaResourceTechnicalMetadataBuilder.setMediaTier(mediaTier);
+    final MediaTier mediaTierBeforeLicense = MediaTier.T1;
```
stzanakis, line 47: This is not tested later down on the assertions?
Perhaps other tests are similar.

stzanakis (changes requested):
Seems fine.
There are some missing assertions on the tests.

metis-indexing/src/test/java/eu/europeana/indexing/tiers/view/MediaResourceTechnicalMetadataBuilderTest.java
JoanaCMS: You are right, I forgot to add the extra assertion. I just fixed that and I'll check other tests

PR #636, opened 2023-11-13
MET-4985_valid_eCloud_tasks_marked_failed_in_Metis

marcin-rp:
…while checking progress. The only left case when it does it  is AccessDeniedOrObjectDoesNotExistException.

### DIFF
--- a/metis-core/metis-core-common/src/main/java/eu/europeana/metis/core/workflow/plugins/AbstractExecutablePlugin.java
+++ b/metis-core/metis-core-common/src/main/java/eu/europeana/metis/core/workflow/plugins/AbstractExecutablePlugin.java
@@ -221,14 +221,10 @@ public abstract class AbstractExecutablePlugin<M extends AbstractExecutablePlugi
   }
 
   @Override
-  public MonitorResult monitor(DpsClient dpsClient) throws ExternalTaskException {
+  public MonitorResult monitor(DpsClient dpsClient) throws DpsException {
     LOGGER.info("Requesting progress information for externalTaskId: {}", getExternalTaskId());
-    TaskInfo taskInfo;
-    try {
-      taskInfo = dpsClient.getTaskProgress(getTopologyName(), Long.parseLong(getExternalTaskId()));
-    } catch (DpsException | RuntimeException e) {
-      throw new ExternalTaskException("Requesting task progress failed", e);
-    }
+    TaskInfo taskInfo =
+        dpsClient.getTaskProgress(getTopologyName(), Long.parseLong(getExternalTaskId()));
     LOGGER.info("Task information received for externalTaskId: {}", getExternalTaskId());
     updateExecutionProgress(taskInfo);
     return new MonitorResult(taskInfo.getState(), taskInfo.getStateDescription());
--- a/metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/execution/WorkflowExecutor.java
+++ b/metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/execution/WorkflowExecutor.java
@@ -1,10 +1,10 @@
 package eu.europeana.metis.core.execution;
 
-import static eu.europeana.metis.network.ExternalRequestUtil.UNMODIFIABLE_MAP_WITH_NETWORK_EXCEPTIONS;
 import static java.lang.Thread.currentThread;
 
 import eu.europeana.cloud.client.dps.rest.DpsClient;
 import eu.europeana.cloud.common.model.dps.TaskState;
+import eu.europeana.cloud.service.dps.exception.AccessDeniedOrObjectDoesNotExistException;
 import eu.europeana.cloud.service.dps.exception.DpsException;
 import eu.europeana.metis.core.dao.DataEvolutionUtils;
 import eu.europeana.metis.core.dao.ExecutedMetisPluginId;
@@ -27,7 +27,6 @@ import eu.europeana.metis.core.workflow.plugins.PluginStatus;
 import eu.europeana.metis.core.workflow.plugins.ThrottlingValues;
 import eu.europeana.metis.exception.BadContentException;
 import eu.europeana.metis.exception.ExternalTaskException;
-import eu.europeana.metis.network.ExternalRequestUtil;
 import java.util.Date;
 import java.util.List;
 import java.util.Optional;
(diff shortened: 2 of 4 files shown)


metis-core/metis-core-service/src/test/java/eu/europeana/metis/core/execution/TestWorkflowExecutor.java
```diff
@@ -232,11 +234,57 @@ void callNonMockedFieldValue_ConsecutiveMonitorFailures() throws Exception {
     workflowExecution.setId(new ObjectId());
     workflowExecution.setWorkflowStatus(WorkflowStatus.INQUEUE);
     workflowExecution.setMetisPlugins(abstractMetisPlugins);
+    workflowExecution.setStartedDate(new Date());
+
+    doReturn(oaipmhHarvestPluginMetadata).when(oaipmhHarvestPlugin).getPluginMetadata();
+
+    Throwable[] dpsException100Times=new Throwable[100];
+    Arrays.setAll(dpsException100Times, index -> new DpsException());
```
jeortizquan, line 242:
```
Arrays.setAll(dpsException100Times,
        index -> new ExternalTaskException("Requesting task progress failed", new DpsException()));
```

jeortizquan (changes requested):
Please have a look on the observations to the code, and update the unit tests accordingly to this latest changes.

metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/execution/WorkflowExecutor.java
```diff
@@ -408,36 +406,21 @@ private void periodicCheckingLoop(long sleepTime, AbstractExecutablePlugin plugi
...
-//        if (cause instanceof IllegalStateException) {
-//          //If the application has a forcible shutdown we might experience the JerseyClient in DpsClient
-//          // internally closing down without our consent even if we would have a synchronized close
-//          // implemented. This catch gives us a little more assurance of avoiding an execution
-//          // being marked as failed.
-//          LOGGER.warn("Application is probably shutting down at the moment and dpsClient has "
-//              + "closed without our consent, so we are ignoring this exception.", e);
-//          return;
-//        }
+      }catch (AccessDeniedOrObjectDoesNotExistException e){
```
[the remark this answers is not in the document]
marcin-rp: 1. We talked about it in our team and decided to not split AccessDeniedOrObjectDoesNotExistException.
Reasons:
-the user who does not have right to the object should not be able to check if the object exists. So both situations are indistinguishable. On this API level.
-we need to change eCloud api. 
-it is technically more difficult, we check acl now by annotations, and we need to check also if the object itself exists after it.

So, if you want to do the change anyway. Maybe lets meet in greater team together to discuss it.

2. I could not revert the whole code cause here is main change - Metis should not fail on any exception but the AccessDeniedOrObjectDoesNotExistException.
So I only changed the code to not catch AccessDeniedOrObjectDoesNotExistException directly but catch it wrapped by UnrecoverableExternalTaskException.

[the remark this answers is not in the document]
jeortizquan: * I pointed out the split `AccessDeniedOrObjectDoesNotExistException` because is not a good practice (Clean Code) at least in the naming (because indicates it does two things). We can leave it as it is 
* I understand now what you explain in point. no. 2 because I see `MessageBodyProviderNotFoundException` can still happen it comes from `org.glassfish.jersey.message.internal` dependency and the `IllegalStateException` comes from the `eu.europeana.cloud.service.mcs.persistent.uis.UISClientHandlerImpl` so in this cases we don't want metis to fail is that correct?

metis-common/metis-common-utils/src/main/java/eu/europeana/metis/exception/UnrecoverableExternalTaskException.java
```diff
@@ -0,0 +1,12 @@
+package eu.europeana.metis.exception;
+
+/**
+ * Type of ExternalTaskException indicating that there is no sense of retrying failed operation.
+ * Because the result will be always negative. For example when the task does not exist on server.
+ */
+public class UnrecoverableExternalTaskException extends ExternalTaskException{
```
jeortizquan, line 7: for this moment please extend from `GenericMetisException`

metis-core/metis-core-common/src/main/java/eu/europeana/metis/core/workflow/plugins/AbstractExecutablePlugin.java
```diff
@@ -226,6 +228,8 @@ public MonitorResult monitor(DpsClient dpsClient) throws ExternalTaskException {
     TaskInfo taskInfo;
     try {
       taskInfo = dpsClient.getTaskProgress(getTopologyName(), Long.parseLong(getExternalTaskId()));
+    }catch (AccessDeniedOrObjectDoesNotExistException e){
```
jeortizquan, line 231: watch out for the formatting apply, "format code" as mention above also to this file.

metis-common/metis-common-utils/src/main/java/eu/europeana/metis/exception/UnrecoverableExternalTaskException.java
```diff
@@ -0,0 +1,12 @@
+package eu.europeana.metis.exception;
+
+/**
+ * Type of ExternalTaskException indicating that there is no sense of retrying failed operation.
+ * Because the result will be always negative. For example when the task does not exist on server.
+ */
```
jeortizquan, line 6: add `@ResponseStatus(value = HttpStatus.BAD_REQUEST, reason = "Unrecoverable task")`

metis-core/metis-core-common/src/main/java/eu/europeana/metis/core/workflow/plugins/AbstractExecutablePlugin.java
jeortizquan: update `ExecutablePlugin.java` to support the exception and update the java documentation.
`  MonitorResult monitor(DpsClient dpsClient) throws ExternalTaskException, UnrecoverableExternalTaskException;`

update method in `AbstractExecutablePlugin.java` to support the exception.
`@Override
  public MonitorResult monitor(DpsClient dpsClient) throws ExternalTaskException, UnrecoverableExternalTaskException {`

jeortizquan (changes requested):
There are small changes requested, please apply formatting, and then we are good to go.

PR #645, opened 2024-01-09
MET-5632 update tier representation in mongo

jeortizquan:


metis-indexing/src/main/java/eu/europeana/indexing/fullbean/EuropeanaAggregationFieldInput.java
```diff
@@ -1,14 +1,12 @@
 package eu.europeana.indexing.fullbean;
```
stzanakis, line 1: Is then the field qualityAnnotations required?
Can we not remove it from the class now?

metis-indexing/src/test/resources/europeana_record_rdf.xml
```diff
@@ -0,0 +1,1060 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
stzanakis, line 1: When using example files for testing, I would say try to make them as small as possible to test exactly what you want to test. Coming back to it later to fix/update something will be possibly very difficult

metis-indexing/src/test/java/eu/europeana/indexing/fullbean/RdfToFullBeanConverterTest.java
```diff
@@ -0,0 +1,142 @@
...
+  private static void assertAbout(FullBeanImpl fullBean) {
+    assertEquals("/277/CMC_HA_1185", fullBean.getAbout());
+  }
+
+  private static void assertEuropeanaCollection(FullBeanImpl fullBean) {
+    assertEquals("277_local_09012024_1543", fullBean.getEuropeanaCollectionName()[0]);
+  }
+
+  private static void assertProxies(FullBeanImpl fullBean) {
+    assertEquals("/proxy/europeana/277/CMC_HA_1185", fullBean.getProxies().get(0).getAbout());
```
stzanakis, line 57: get[x] and/or getFirst does not seem like a sound check. The order might be different on a second execution. Probably it's better to get if it's part of the list.
This involves all the cases in this class

metis-indexing/src/main/java/eu/europeana/indexing/fullbean/RdfToFullBeanConverter.java
```diff
@@ -41,7 +41,7 @@ private static Date convertToDate(String dateString) {
     return Date.from(Instant.from(DateTimeFormatter.ISO_OFFSET_DATE_TIME.parse(dateString)));
   }
 
-  private static <S, T> List<T> convertList(List<S> sourceList, Function<S, T> converter,
+  protected static <S, T> List<T> convertList(List<S> sourceList, Function<S, T> converter,
```
stzanakis, line 44: Changing the visibility of a method just for the sake of creating a test for it is considered bad practise. They are implementation details and public methods are calling those as well.

stzanakis (changes requested):

metis-indexing/src/main/java/eu/europeana/indexing/fullbean/EuropeanaAggregationFieldInput.java
jeortizquan: true. is not required here any more. removed.

metis-indexing/src/main/java/eu/europeana/indexing/fullbean/RdfToFullBeanConverter.java
jeortizquan: Actually was not only for the sake of the test the intent was that the code needed to be testable, this indicates also that the abstraction is not located on the right place(file) a solution would be to abstract this methods to a separate class. After removing the unused code, there is no need to change the visibility so I am returning it to their previous state.

metis-indexing/src/test/resources/europeana_record_rdf.xml
jeortizquan: Reduced the size, by removed redundant information.

metis-indexing/src/test/java/eu/europeana/indexing/fullbean/RdfToFullBeanConverterTest.java
jeortizquan: Refactored.

metis-indexing/src/main/java/eu/europeana/indexing/fullbean/RdfToFullBeanConverter.java
stzanakis: It's arguable.
Generally public methods call private methods so by properly testing the public methods the private methods are tested as well. Unless those methods were not meant to be private or located in this class(e.g. also used by other classes)
Otherwise we would be making every method public so that we can test them individually which breaks it's encapsulation.

jeortizquan: Correct. this was just flagging an issue of design and you are right we don't want to break the encapsulation, and I agree that a proper testing will cover the private methods of a public method.

PR #651, opened 2024-01-31
Bug/met 5717 fix authentication startup zoho

jeortizquan:


stzanakis:
This is not necessary anymore.
Issue was the client account that had been deleted.

PR #657, opened 2024-02-16
MET-5796 remove register user feature and zoho usage

jeortizquan:


jeortizquan:
Not using this approach.

PR #661, opened 2024-02-26
feat/MET-5796_MET-5804 disconnect zoho and update unit tests

jeortizquan:


jeortizquan:
This is included on MET-5796 subtask MET-5803

PR #665, opened 2024-03-12
Feat/MET-5763 Add support for stl files in media processing

JoanaCMS:


jeortizquan (changes requested):
My suggestion is to add some unit tests in `MimeTypeDetectHttpClientTest` with the tests files to cover the 3 new cases.

metis-media-service/src/test/java/eu/europeana/metis/mediaprocessing/extraction/MediaExtractorImplTest.java
```diff
@@ -70,7 +71,7 @@ static void prepare() {
     textProcessor = mock(TextProcessor.class);
     media3dProcessor = mock(Media3dProcessor.class);
     mediaExtractor = spy(new MediaExtractorImpl(resourceDownloadClient, mimeTypeDetectHttpClient,
-        tika, imageProcessor, audioVideoProcessor, textProcessor, media3dProcessor));
+        new TikaWrapper(), imageProcessor, audioVideoProcessor, textProcessor, media3dProcessor));
```
jochen-vermeulen, line 74: `tika` is never used. I think perhaps here you'd be better off mocking TikaWrapper? Or otherwise just removing the tika value from this class.

jochen-vermeulen (changes requested):

metis-media-service/src/test/java/eu/europeana/metis/mediaprocessing/extraction/MediaExtractorImplTest.java
JoanaCMS: Forgot to fix that, it is done

PR #666, opened 2024-03-13
Feat/MET-5764 Add support gltf files for media processing

JoanaCMS:


jeortizquan (changes requested):
My suggestion is to add some unit tests in `MimeTypeDetectHttpClientTest` with tests files to cover the 2 new cases.

jeortizquan (approved):

PR #673, opened 2024-05-01
Debt/met 5966 align harvesting methods

jochen-vermeulen:


metis-harvesting/src/main/java/eu/europeana/metis/harvesting/http/HttpHarvesterImpl.java
```diff
@@ -258,24 +331,38 @@ public FileVisitResult preVisitDirectory(Path dir, BasicFileAttributes attrs) {
...
+    @Override
+    public ByteArrayInputStream getContent() {
+      return entryContent;
     }
 
     @Override
-    public String getEntryName() {
-      return entryName;
+    public boolean isDeleted() {
+      return false;
```
JoanaCMS, line 354: It always returns false, is this intentional?

metis-harvesting/src/main/java/eu/europeana/metis/harvesting/http/HttpHarvesterImpl.java
```diff
@@ -258,24 +331,38 @@ public FileVisitResult preVisitDirectory(Path dir, BasicFileAttributes attrs) {
...
     @Override
+    @Deprecated
     public ByteArrayInputStream getEntryContent() {
-      return entryContent;
+      return getContent();
+    }
+
+    @Override
+    public Instant getTimeStamp() {
+      return null;
```
JoanaCMS, line 365: Also here, it always returns null

metis-harvesting/src/main/java/eu/europeana/metis/harvesting/http/HttpHarvesterImpl.java
jochen-vermeulen: Yes. Other protocols (such as OAI) sometimes return records or record IDs that are deleted at source (with a flag saying that it is deleted). However, this is not the case here. In a zip file we can't make that distinction (we don't have such a flag) and we assume that all records are active and to be harvested.

metis-harvesting/src/main/java/eu/europeana/metis/harvesting/http/HttpHarvesterImpl.java
jochen-vermeulen: Similar story. Other protocols (can) maintain a timestamp for a record (i.e. last change time). In a zip file we don't have this.

PR #676, opened 2024-07-11
Feat/MET-6002 Save record deletion reason

JoanaCMS:


jeortizquan (changes requested):
There is a small change. see the comments.

PR #682, opened 2024-08-07
feat/MET-6035 Removing depublication reason from some API calls

JoanaCMS:


jeortizquan (approved):

metis-core/metis-core-service/src/test/java/eu/europeana/metis/core/dao/TestDepublishRecordIdDao.java
jeortizquan, line 1: just remove. `this unused import 'eu.europeana.metis.core.workflow.plugins.DepublicationReason`

PR #684, opened 2024-08-19
MET-6076 changes in OAI client

pWoz:
…aiClient)


pWoz:
Closed without merging since we found the source of this issue in the PoC project.

PR #685, opened 2024-08-21
Feat/MET-6035 Fixing depublication reason

JoanaCMS:


metis-core/metis-core-service/src/main/java/eu/europeana/metis/core/dao/DepublishRecordIdDao.java
```diff
@@ -328,7 +330,7 @@ private Query<DepublishRecordId> prepareQueryForDepublishRecordIds(String datase
    * {@link DepublicationStatus#PENDING_DEPUBLICATION}
    */
   public void markRecordIdsWithDepublicationStatus(String datasetId, Set<String> recordIds,
-      DepublicationStatus depublicationStatus, @Nullable Date depublicationDate) {
+      DepublicationStatus depublicationStatus, @Nullable Date depublicationDate, DepublicationReason depublicationReason) {
 
     // Check correctness of parameters
```
jochen-vermeulen, line 335: Perhaps add a check below? Similarly to where we check that the date must not be null when depublishing a record, we should check that the reason must not be null either?

PR #691, opened 2024-09-11
MET-6115 Refine fields for tombstone and index

stzanakis:


metis-indexing/src/main/java/eu/europeana/indexing/IndexerImpl.java
```diff
@@ -134,29 +140,69 @@ public boolean indexTombstone(String rdfAbout) throws IndexerRelatedIndexingExce
...
+    final Optional<ProxyImpl> providerProxy =
+        publishedFullbean.getProxies().stream().filter(not(ProxyImpl::isEuropeanaProxy)).findFirst();
+    providerProxy.ifPresent(proxy -> tombstoneFullbean.setProxies(List.of(prepareProxy(proxy))));
     return tombstoneFullbean;
   }
 
+  private static EuropeanaAggregation prepareEuropeanaAggregation(EuropeanaAggregation europeanaAggregation) {
+    final ChangeLog tombstoneChangeLog = new ChangeLogImpl();
+    tombstoneChangeLog.setType("Delete");
+    tombstoneChangeLog.setContext("http://data.europeana.eu/vocabulary/depublicationReason/noPermission");
```
jeortizquan, line 170: this has to be parametrizable? the de-publication reason I mean.

jeortizquan (commented):
I just have an observation over the de-publication.

metis-indexing/src/main/java/eu/europeana/indexing/IndexerImpl.java
stzanakis: Yes it should be. As I mentioned in the standup I realized that after looking at the upcoming ticket MET-6060, so I will handle that with that ticket.

PR #693, opened 2024-09-13
MET-6117 Remove deprecations from oai harvesting

stzanakis:


metis-harvesting/src/main/java/eu/europeana/metis/harvesting/oaipmh/OaiHarvestingIterator.java
```diff
@@ -0,0 +1,218 @@
...
+      return readCompleteListSizeFromXML(listIdentifiersResponse);
+    } catch (OAIRequestException | IOException e) {
+      throw new HarvesterException(String.format(
+          "Problem with counting records for endpoint %1$s because of: %2$s",
+          harvest.getRepositoryUrl(), e.getMessage()), e);
+    }
+  }
+
+  @Override
+  public void close() throws IOException {
```
jeortizquan, line 143: this method is not needed anymore I see `oaiClient` implements `Closable` interface which implements `AutoClosable`. So my suggestion is to remove it

jeortizquan (commented):
Looks good to me there is one observation of one method that can be removed now that we are using java 21.

metis-harvesting/src/main/java/eu/europeana/metis/harvesting/oaipmh/OaiHarvestingIterator.java
stzanakis: Can you please elaborate?
If I remove it then there is no wai to close the oaiClient..

PR #697, opened 2024-10-08
MET-6139 Update code according to the updates in metis-schema classes

stzanakis:
@jeortizquan 
This is a first step of the code migration. 
In this first iteration Tests(if any) are not migrated, packaging organization not optimized, reported issues not resolved.
With the first review, we just agree on the direction and then we continue the implementation on this PR.

@jochen-vermeulen In the end the todo I mentioned was actually from you in the end. So perhaps you have a say in this one as well?


stzanakis:
Based on this [metis-schema PR](https://github.com/europeana/metis-schema/pull/14) I am closing this to rework on a smaller update and perhaps some renaming on the media library.

PR #698, opened 2024-10-10
MET-6211 Remove deprecations in metis indexing and cleanup

stzanakis:


metis-authentication/metis-authentication-rest/src/main/java/eu/europeana/metis/authentication/rest/controller/AuthenticationController.java
```diff
@@ -208,15 +207,13 @@ public void updateUserToMakeAdmin(@RequestHeader("Authorization") String authori
     if (emailParameter == null || StringUtils.isBlank(emailParameter.getEmail())) {
       throw new BadContentException("userEmailToMakeAdmin is empty");
     }
-    String accessToken = authenticationService
-        .validateAuthorizationHeaderWithAccessToken(authorization);
+    String accessToken = authenticationService.validateAuthorizationHeaderWithAccessToken(authorization);
```
jeortizquan, line 210: suggestion to use `final` here.

metis-common/metis-common-utils/src/main/java/eu/europeana/metis/utils/CommonStringValues.java
```diff
@@ -24,8 +24,19 @@ public final class CommonStringValues {
...
+
+  /**
+   * Sanitized input value from Logging injection attacks(javasecurity:S5145).
+   * <p>Replaces CR and LF characters with a safe value e.g. '_'.</p>
+   *
+   * @param input the input
+   * @return the sanitized input, safe for logging
+   */
+  public static String sanitizeCRLF(String input) {
+    return input == null ? null : CRLF_PATTERN.matcher(input).replaceAll("");
```
jeortizquan, line 40: suggestion: do returning an empty string would be better? `return input == null ? "" : CRLF_PATTERN.matcher(input).replaceAll("");` instead of `null`?

metis-common/metis-common-utils/src/main/java/eu/europeana/metis/utils/CommonStringValues.java
```diff
@@ -24,8 +24,19 @@ public final class CommonStringValues {
...
 
-  public static final Pattern CRLF_PATTERN = Pattern.compile(CommonStringValues.REPLACEABLE_CRLF_CHARACTERS_REGEX);
+  public static final Pattern CRLF_PATTERN = Pattern.compile(REPLACEABLE_CRLF_CHARACTERS_REGEX);
 
   private CommonStringValues() {
   }
+
+  /**
+   * Sanitized input value from Logging injection attacks(javasecurity:S5145).
+   * <p>Replaces CR and LF characters with a safe value e.g. '_'.</p>
```
jeortizquan, line 34: the comment doesn't correspond to what the function does.

metis-common/metis-common-utils/src/main/java/eu/europeana/metis/utils/DepublicationReason.java
```diff
@@ -2,28 +2,39 @@
...
-  REMOVED_DATA_AT_SOURCE("Removed data at source", "http://data.europeana.eu/vocabulary/depublicationReason/sourceRemoval"),
-  GENERIC("Generic", "http://data.europeana.eu/vocabulary/depublicationReason/generic"),
-  UNKNOWN("Unknown", "http://data.europeana.eu/vocabulary/depublicationReason/unknown");
+  BROKEN_MEDIA_LINKS("Broken media links", "contentTier0"),
+  GDPR("GDPR", "gdpr"),
+  PERMISSION_ISSUES("Permission issues", "noPermission"),
+  SENSITIVE_CONTENT("Sensitive content", "sensitiveContent"),
+  REMOVED_DATA_AT_SOURCE("Removed data at source", "sourceRemoval"),
+  GENERIC("Generic", "generic"),
+  UNKNOWN("Unknown", "");
```
jeortizquan, line 19: `"Unknown", "unknown"` ? or is ok `empty string`?

metis-common/metis-common-utils/src/main/java/eu/europeana/metis/utils/CommonStringValues.java
```diff
@@ -24,8 +24,19 @@ public final class CommonStringValues {
...
+   *
+   * @param input the input
+   * @return the sanitized input, safe for logging
+   */
+  public static String sanitizeCRLF(String input) {
```
jeortizquan, line 39: do we want `american english` or `british english` :smile: ? `sanitizeCRLF` or `sanitiseCRLF`, also in the comment `sanitised`

jeortizquan (commented):
Some observations...

metis-authentication/metis-authentication-rest/src/main/java/eu/europeana/metis/authentication/rest/controller/AuthenticationController.java
stzanakis: Done. I guess additional suggestion, since I did not add/change the code, just put it in one line.

metis-common/metis-common-utils/src/main/java/eu/europeana/metis/utils/CommonStringValues.java
stzanakis: :D For this specific difference I always use 'z', it makes more sense to me and in code we usually use US local. In fact almost everywhere we use 'z', although it tends to be 's' in some documents.

metis-common/metis-common-utils/src/main/java/eu/europeana/metis/utils/CommonStringValues.java
stzanakis: Indeed, corrected.

metis-common/metis-common-utils/src/main/java/eu/europeana/metis/utils/CommonStringValues.java
stzanakis: I think not, because it's about sanitizing the field(of bad characters) rather than correcting it. Alternatively I would have thrown an exception, but I thought it's the responsibility of the client to have a valid value.
What do you say?

metis-common/metis-common-utils/src/main/java/eu/europeana/metis/utils/DepublicationReason.java
stzanakis: It has to be empty, I had it "unknown", but after discussions with colleagues this url is not valid according to the design and only the enum value is fine for historical data as described in the javadoc.

jeortizquan: okey

metis-common/metis-common-utils/src/main/java/eu/europeana/metis/utils/CommonStringValues.java
jeortizquan: mmm. :thinking: I was looking it from the perspective of not returning a `null` to make the function "safe" and getting always `string` the empty string doesn't harm was my thought.

jeortizquan: leave it  as it is
afbeelding

jeortizquan (approved):

PR #702, opened 2024-11-13
MET-6252 Add additional fields to tombstone

stzanakis:


metis-indexing/src/main/java/eu/europeana/indexing/TombstoneUtil.java
```diff
@@ -0,0 +1,214 @@
...
+public class TombstoneUtil {
+
+  /**
+   * Prepare the tombstone fullbean.
+   *
+   * @param publishedFullbean the published fullbean
+   * @param depublicationReason the depublication reason
+   * @return the newly created tombstone
+   */
+  public FullBeanImpl prepareTombstoneFullbean(FullBeanImpl publishedFullbean, DepublicationReason depublicationReason) {
```
jeortizquan, line 41: my sugestion would be to use a `static` method and update the rest of private also to `static`

metis-indexing/src/main/java/eu/europeana/indexing/IndexerImpl.java
```diff
@@ -43,6 +34,7 @@ public class IndexerImpl implements Indexer {
 
   private final IndexingSupplier<StringToFullBeanConverter> stringToRdfConverterSupplier;
 
+  private final TombstoneUtil tombstoneUtil = new TombstoneUtil();
```
jeortizquan, line 37: use methods from static class? what do you think? I am seeing this from the perspective every indexer will be needed to create this object, than using one static class is just one time on the ram.

jeortizquan (approved):

PR #710, opened 2025-01-28
Feat/met 6260 reindex depublish records from datasets

jeortizquan:


metis-indexing/src/main/java/eu/europeana/indexing/solr/SolrDocumentPopulator.java
```diff
@@ -198,15 +198,15 @@ private void populateWithDateRanges(SolrInputDocument document, RdfWrapper rdfWr
     final ProxyType europeanaProxy = rdfWrapper.getProxies().stream().filter(RdfWrapper::isEuropeanaProxy).findFirst()
                                                .orElseThrow();
 
-    final List<String> proxyChoiceLinks = europeanaProxy.getChoiceList().stream()
+    final List<String> proxyChoiceLinks = europeanaProxy.getChoiceList()!=null? europeanaProxy.getChoiceList().stream()
```
stzanakis, line 201: I cannot see why the choice list is empty here. Is this happening during indexing or tombstoning?

metis-indexing/src/main/java/eu/europeana/indexing/IndexerPool.java
```diff
@@ -108,11 +122,12 @@ public void indexRdf(RDF stringRdfRecord, IndexingProperties indexingProperties)
...
+   * @return the boolean result of the record removal.
    * @throws IndexingException In case something went wrong.
    */
-  public void remove(String stringRdfRecord) throws IndexingException {
-    indexRecord(indexer -> indexer.remove(stringRdfRecord));
+  public boolean remove(String stringRdfRecord) throws IndexingException {
+    return indexBoolTask(indexer -> indexer.remove(stringRdfRecord));
   }
 
   private void indexRecord(IndexTask indexTask) throws IndexingException {
```
stzanakis, line 133: You can change this to return a value instead of duplicating the method(and also avoid creating the IndexBoolTask) since it is private.
And adapt the public methods accordingly.

jeortizquan: done.

metis-indexing/src/main/java/eu/europeana/indexing/solr/SolrDocumentPopulator.java
jeortizquan: It happened during indexing, I was receiving the message `unable to publish in solr server`. I reverted the change and tested again and it didn't happened anymore. done.

PR #711, opened 2025-01-31
MET-6370-Make-HttpHarvester-step-method-public

marcin-rp:
… and used independently.

### DIFF
--- a/metis-harvesting/src/main/java/eu/europeana/metis/harvesting/http/HttpHarvesterImpl.java
+++ b/metis-harvesting/src/main/java/eu/europeana/metis/harvesting/http/HttpHarvesterImpl.java
@@ -85,15 +85,9 @@ public class HttpHarvesterImpl implements HttpHarvester {
   }
 
   private Path extractArchive(Path archiveFile) throws HarvesterException {
-    final Path extractedDirectory = archiveFile.toAbsolutePath().getParent();
-    extractArchive(archiveFile, extractedDirectory);
-    // Return the extracted directory
-    return extractedDirectory;
-  }
 
-  @Override
-  public void extractArchive(Path archiveFile, Path extractedDirectory) throws HarvesterException{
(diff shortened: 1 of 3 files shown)


metis-harvesting/src/main/java/eu/europeana/metis/harvesting/http/HttpHarvesterImpl.java
```diff
@@ -108,12 +114,10 @@ private Path extractArchive(Path archiveFile) throws HarvesterException {
...
       throw new HarvesterException("Problem correcting directory rights.", e);
     }
-
-    // Return the extracted directory
-    return extractedDirectory;
   }
 
-  private Path downloadFile(String archiveUrlString, Path downloadDirectory) throws IOException, URISyntaxException {
+  @Override
+  public Path downloadFile(String archiveUrlString, Path downloadDirectory) throws IOException, URISyntaxException {
```
stzanakis, line 120: Since this now becomes a public method we'd need to improve it by throwing HarvesterException instead of these two, similarly to the other methods.

metis-harvesting/src/main/java/eu/europeana/metis/harvesting/http/HttpHarvesterImpl.java
```diff
@@ -177,7 +181,7 @@ public void forEachFiltered(ReportingIteration<FullRecord> action, Predicate<Pat
     }
   }
 
-  private static class PathIterator extends AbstractHttpHarvestIterator<Path> {
+  public static class PathIterator extends AbstractHttpHarvestIterator<Path> {
```
stzanakis, line 184: I suppose this now is a reusable class outside of the containing. Should we convert it to a top level class with javadoc?

stzanakis (changes requested):

PR #712, opened 2025-02-06
MET-6143 Fix runtime exception being caught and not propagated

Ruufuus:
Could anyone take a look why this changes could cause those issues in build? 
As far as I see tests are failing when setting up Mongo db for test. I cannot recreate those issues locally but maybe I am not aware of something.

Also could you tell me if that change is appropriate? We need to not catch runtime exceptions like Interruption Exceptions or Cancellation Exceptions for Flink to handle task cancellation properly without any work-arounds.

### DIFF
--- a/metis-enrichment/metis-enrichment-client/src/main/java/eu/europeana/enrichment/rest/client/dereference/DereferencerImpl.java
+++ b/metis-enrichment/metis-enrichment-client/src/main/java/eu/europeana/enrichment/rest/client/dereference/DereferencerImpl.java
@@ -47,7 +47,7 @@ import org.springframework.web.client.HttpClientErrorException.BadRequest;
 public class DereferencerImpl implements Dereferencer {
 
   private static final Logger LOGGER = LoggerFactory.getLogger(DereferencerImpl.class);
-  public static final String CANCELLATION_EXCEPTION_WARN_MESSAGE = "Cancellation exception occurred while trying to perform dereferencing, rethrowing.";
+  private static final String CANCELLATION_EXCEPTION_WARN_MESSAGE = "Cancellation exception occurred while trying to perform dereferencing, rethrowing.";
 
   private final EntityMergeEngine entityMergeEngine;
   private final EntityResolver entityResolver;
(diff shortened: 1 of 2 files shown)


stzanakis:
@Ruufuus Please rebase your branch, you have branched off from an earlier commit in develop. This has already been fixed.

jeortizquan (commented):
from my point of view looks good to me, I see that this will allow that the `CancellationException` goest all the way up. it won't be logged by the dereferencer. Please check with Jochen if we are fine with this, and let's hear what is Simon opinion of the PR.

metis-enrichment/metis-enrichment-client/src/main/java/eu/europeana/enrichment/rest/client/dereference/DereferencerImpl.java
```diff
@@ -249,6 +250,8 @@ public DereferencedEntities dereferenceOwnEntities(Set<ReferenceTerm> resourceId
                    result.putIfAbsent(notFoundOwnId, Collections.emptyList());
                  });
       return new DereferencedEntities(result, reports, classType);
+    } catch (CancellationException e){
```
stzanakis, line 253: This looks reasonable to me, although I would add a LOGGER.warn when this happens as well.
Applies for all locations.

Ruufuus:
@stzanakis Something like this? 
Wasn't really sure where to put constants with log message (it has to be used 3 times so I guess constant is appropriate) and I couldn't find place to put it so I left it in class, Kinda make sense to leave it there since it is dereferencer specific log constant and I doubt it will be reused anyway.

stzanakis:
Yes like that is fine.

metis-enrichment/metis-enrichment-client/src/main/java/eu/europeana/enrichment/rest/client/dereference/DereferencerImpl.java
```diff
@@ -46,6 +47,7 @@
 public class DereferencerImpl implements Dereferencer {
 
   private static final Logger LOGGER = LoggerFactory.getLogger(DereferencerImpl.class);
+  public static final String CANCELLATION_EXCEPTION_WARN_MESSAGE = "Cancellation exception occurred while trying to perform dereferencing, rethrowing.";
```
stzanakis, line 50: Shouldn't this be private?

stzanakis:
Thank you, that is perfect.
To comment on you note "Also could you tell me if that change is appropriate? We need to not catch runtime exceptions like Interruption Exceptions or Cancellation Exceptions for Flink to handle task cancellation properly without any work-arounds.".
You are absolutely right, the code in fact catches Exception class which in my opinion is not correct(it should be a specific known set of exceptions) and that brought the side effect you had in Flink and any other app that hosts this code in fact.
InterruptedException could also be a problem indeed.
So perhaps we should rework this code in the future a bit. I've created a ticket MET-6451

PR #732, opened 2025-06-27
MET-6616-Record-became-a-blank-file-after-media-processing

marcin-rp:


jeortizquan:
this needs to be implemented in the metis-schema.

PR #741, opened 2025-08-05
MET-6713 use try-resources for entity-api-client and refactor constructor

jeortizquan:
…ctors


jeortizquan:
pending sonar issues addressed by MET-6708

PR #768, opened 2025-12-23
MET-6921: Give TIKA detection from on-drive resource access to all so…

jochen-vermeulen:
…urce information.

### DIFF
--- a/metis-media-service/src/main/java/eu/europeana/metis/mediaprocessing/extraction/MediaExtractorImpl.java
+++ b/metis-media-service/src/main/java/eu/europeana/metis/mediaprocessing/extraction/MediaExtractorImpl.java
@@ -1,7 +1,6 @@
 package eu.europeana.metis.mediaprocessing.extraction;
 
 import static eu.europeana.metis.utils.SonarqubeNullcheckAvoidanceUtils.performThrowingAction;
-import static org.apache.tika.metadata.HttpHeaders.CONTENT_TYPE;
 
 import eu.europeana.metis.mediaprocessing.MediaExtractor;
 import eu.europeana.metis.mediaprocessing.exception.MediaExtractionException;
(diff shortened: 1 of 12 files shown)


metis-media-service/src/main/java/eu/europeana/metis/mediaprocessing/extraction/MediaExtractorImpl.java
```diff
@@ -219,19 +217,17 @@ String detectAndVerifyMimeType(Resource resource, ProcessingMode mode)
     }
 
     // Detect the mime type.
-    final String providedMimeType = resource.getProvidedMimeType();
     final String detectedMimeType;
-    final boolean hasContent;
     try {
-      hasContent = resource.hasContent();
-      detectedMimeType = hasContent ? detectType(resource.getContentPath(), providedMimeType)
+      detectedMimeType = resource.hasContent() ? detectType(resource.getContentPath(), resource)
```
jeortizquan: I would suggest then to merge the parameters since it is using the same object and just update the method signature.
from `detectType(resource.getContentPath(), resource)` to `detectType(resource)`

metis-media-service/src/main/java/eu/europeana/metis/mediaprocessing/wrappers/TikaWrapper.java
```diff
@@ -22,6 +28,82 @@ public TikaWrapper() {
...
+        extractedResourceName =
+            slashIndex < 0 ? resourcePath : resourcePath.substring(slashIndex + 1);
+      }
+    }
+    return extractedResourceName;
+  }
+
+  private static Metadata getMetadataFromResource(RemoteResourceMetadata resource) {
+    final Metadata metadata = new Metadata();
+    final String resourceName = getResourceNameFromContentDispositionOrFromActualURI(
```
jeortizquan: suggestion, same here the parameter also can be simplified to `getResourceNameFromContentDispositionOrFromActualURI(resource)`

jeortizquan (commented):
I have some small suggestions, also update branch with latest updates.

metis-media-service/src/main/java/eu/europeana/metis/mediaprocessing/wrappers/TikaWrapper.java
jochen-vermeulen: OK, here it makes sense to do this, indeed.

metis-media-service/src/main/java/eu/europeana/metis/mediaprocessing/extraction/MediaExtractorImpl.java
jochen-vermeulen: Indeed, that is better here.

jeortizquan (approved):

PR #772, opened 2026-01-27
MET-6921 add unit integration tests

jeortizquan:


metis-media-service/src/test/java/eu/europeana/metis/mediaprocessing/extraction/MediaExtractorImplITTest.java
```diff
@@ -33,4 +40,25 @@ void testIIFExtraction() throws Exception {
     }
   }
 
+  @Disabled("enable imagemagick first on metis-actions for the build.")
+  @ParameterizedTest(name = "{index} => record={0}, expected={1}")
+  @MethodSource
+  void testCube3dExtraction_MimeType(String resourcePath, Set<String> expectedMimeTypes) throws Exception {
```
stzanakis: This is practically identical to the above method. I would conbine them.

metis-media-service/src/test/java/eu/europeana/metis/mediaprocessing/extraction/MediaExtractorImplITTest.java
```diff
@@ -33,4 +40,25 @@ void testIIFExtraction() throws Exception {
...
+    rdfDeserializer = new RdfConverterFactory().createRdfDeserializer();
+    mediaExtractor = new MediaProcessorFactory().createMediaExtractor();
+    InputStream inputRdf = getClass().getClassLoader().getResourceAsStream(resourcePath);
+    List<RdfResourceEntry> resourceEntryList = rdfDeserializer.getRemainingResourcesForMediaExtraction(inputRdf);
+    Set<String> actualMimeTypes = HashSet.newHashSet(2);
```
stzanakis: Why presized to 2? Is the size known? and if it is it is not used, it is now static.

metis-media-service/src/test/java/eu/europeana/metis/mediaprocessing/extraction/MediaExtractorImplITTest.java
```diff
@@ -33,4 +40,25 @@ void testIIFExtraction() throws Exception {
...
+    for (RdfResourceEntry resourceEntry : resourceEntryList) {
+      ResourceExtractionResult extractionResult = mediaExtractor.performMediaExtraction(resourceEntry, true);
+      assertNotNull(extractionResult.getMetadata());
+      actualMimeTypes.add(extractionResult.getMetadata().getMimeType());
+    }
+    assertEquals(expectedMimeTypes, actualMimeTypes);
+  }
+
+  private static Stream<Arguments> testCube3dExtraction_MimeType() {
+    return Stream.of(Arguments.of("__files/record-MET-6921.xml", Set.of("application/xhtml+xml", "model/gltf+json")),
```
stzanakis: I would not name the files as such, something more descriptive, a ticket number would mean nothing in 2 weeks from now.
In fact a single file could accomodate more than one cases(even including the "iiif" file from the first method).

Perhaps we should try merging files as much as possible. Having a separate file for just one different parameter  would create too many unncessary files.

metis-media-service/src/test/java/eu/europeana/metis/mediaprocessing/extraction/MediaExtractorImplITTest.java
```diff
@@ -33,4 +40,25 @@ void testIIFExtraction() throws Exception {
...
+    mediaExtractor = new MediaProcessorFactory().createMediaExtractor();
+    InputStream inputRdf = getClass().getClassLoader().getResourceAsStream(resourcePath);
+    List<RdfResourceEntry> resourceEntryList = rdfDeserializer.getRemainingResourcesForMediaExtraction(inputRdf);
+    Set<String> actualMimeTypes = HashSet.newHashSet(2);
+    for (RdfResourceEntry resourceEntry : resourceEntryList) {
+      ResourceExtractionResult extractionResult = mediaExtractor.performMediaExtraction(resourceEntry, true);
+      assertNotNull(extractionResult.getMetadata());
+      actualMimeTypes.add(extractionResult.getMetadata().getMimeType());
+    }
+    assertEquals(expectedMimeTypes, actualMimeTypes);
```
stzanakis: Isn't this order specific? The order is not predictable or necessary is it?

metis-media-service/src/test/java/eu/europeana/metis/mediaprocessing/extraction/MediaExtractorImplITTest.java
```diff
@@ -33,4 +40,25 @@ void testIIFExtraction() throws Exception {
     }
   }
 
+  @Disabled("enable imagemagick first on metis-actions for the build.")
```
stzanakis: Only imagemagick or other applications as well?
Hmm, perhaps we should re-evaluate this type of tests at some point..

metis-media-service/src/test/java/eu/europeana/metis/mediaprocessing/extraction/MediaExtractorImplITTest.java
jeortizquan: done.

metis-media-service/src/test/java/eu/europeana/metis/mediaprocessing/extraction/MediaExtractorImplITTest.java
jeortizquan: Removed. updated to be dynamic.

metis-media-service/src/test/java/eu/europeana/metis/mediaprocessing/extraction/MediaExtractorImplITTest.java
jeortizquan: done. I renamed the test files and also merged the test.

metis-media-service/src/test/java/eu/europeana/metis/mediaprocessing/extraction/MediaExtractorImplITTest.java
jeortizquan: The order is not predictable. that is why the use of a Set.

metis-media-service/src/test/java/eu/europeana/metis/mediaprocessing/extraction/MediaExtractorImplITTest.java
jeortizquan: mm. you are right, probably there are more dependencies to update on the metis-actions. At that moment only image magick was needed. The intention of these tests are to have a complete round of the media extraction. I updated the comment.

metis-media-service/src/test/java/eu/europeana/metis/mediaprocessing/extraction/MediaExtractorImplITTest.java
```diff
@@ -1,36 +1,77 @@
...
+  }
 
+  @Disabled("Enable media-processing tools (image-magick, ghostscript, etc) first on metis-actions for the build.")
+  @ParameterizedTest(name = "{index} => record={0}, expected-resource{1}, expected-mimetype={2}")
+  @MethodSource
+  void testMediaExtraction_ResourceAndMimeType(String resourcePath,
+      Set<Pair<String, RdfResourceKind>> expectedResource,
+      Set<String> expectedMimeTypes) throws Exception {
+    // Given
     rdfDeserializer = new RdfConverterFactory().createRdfDeserializer();
```
stzanakis: Since the test is repeated isn't it better the instances to be created once for the class?

metis-media-service/src/test/java/eu/europeana/metis/mediaprocessing/extraction/MediaExtractorImplITTest.java
stzanakis: Sorry, my sentence made little sense :D. I meant, don't we want to differentiate from where a resource came from e.g., isShownAt, isShowBy, etc.? Now it will overwrite it, I think.

metis-media-service/src/test/java/eu/europeana/metis/mediaprocessing/extraction/MediaExtractorImplITTest.java
```diff
@@ -1,36 +1,77 @@
...
   private RdfDeserializer rdfDeserializer;
   private MediaExtractor mediaExtractor;
 
-  @Disabled("enable imagemagick first on metis-actions for the build.")
-  @Test
-  void testIIFExtraction() throws Exception {
+  private static Stream<Arguments> testMediaExtraction_ResourceAndMimeType() {
+    return Stream.of(
+        Arguments.of("__files/rdf_with_iiif_sample.xml",
+            Set.of(Pair.of("https://stacks.stanford.edu/image/iiif/zw031pj2507/zw031pj2507_0001/full/full/0/default.jpg", IIIF),
```
stzanakis: If we care about those specific urls, should we also care what mimetype exactly they produce and not compare two different sets? Therefore, we maybe need a Triple (or even better a small record class in the test)
(This comment is relevant to the last comment of assertEquals on the sets)

jeortizquan: done. I updated the test so that it will assertion to consider specific the mimetype and urlType.

metis-media-service/src/test/java/eu/europeana/metis/mediaprocessing/extraction/MediaExtractorImplITTest.java
jeortizquan: done.

PR #780, opened 2026-04-08
MET-7084: Provider references are only dereferenced if they are Europ…

jochen-vermeulen:
…eana links, otherwise only coreferenced (and perhaps replaced).

### DIFF
--- a/metis-enrichment/metis-enrichment-client/src/main/java/eu/europeana/enrichment/rest/client/dereference/Dereferencer.java
+++ b/metis-enrichment/metis-enrichment-client/src/main/java/eu/europeana/enrichment/rest/client/dereference/Dereferencer.java
@@ -20,10 +20,7 @@ public interface Dereferencer extends AutoCloseable {
     EUROPEANA_ENTITY,
 
     /** Only external (non-Europeana) entities are permitted. **/
-    EXTERNAL_ENTITY,
-
-    /** All entities are permitted. **/
-    ANY_ENTITY
+    EXTERNAL_ENTITY
   }
 
   /**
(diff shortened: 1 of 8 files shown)


metis-enrichment/metis-enrichment-client/src/main/java/eu/europeana/enrichment/rest/client/dereference/Dereferencer.java
```diff
@@ -3,14 +3,29 @@
...
 
 /**
  * Instances of this class can perform dereferencing.
  */
 public interface Dereferencer extends AutoCloseable {
 
+  /**
+   * The various entity types permitted for a resource reference.
+   */
+  enum PermittedEntityType {
```
stzanakis, line 17: To me, it would be better to remove the `ANY_ENTITY` and use a set to define permitted types.

In code, you are now doing
`final Set<PermittedEntityType> europeanaTypes = EnumSet.of(PermittedEntityType.EUROPEANA_ENTITY,
PermittedEntityType.ANY_ENTITY);`

and then 
`.filter(entry -> europeanaTypes.contains(entry.getValue()))`

Looks a bit confusing at first while if you "move" the Set and your map was:
`Map<String, PermittedEntityType> resourceIds` then you could simply do:
`.filter(entry -> entry.getValue().contains(EUROPEANA_ENTITY)`
That also makes it a bit clearer to my eyes.

stzanakis (approved):
Have a look at one comment I have about the PermittedEntityType and see if that is of any value.
Please go ahead in either case, the rest looks fine.

metis-enrichment/metis-enrichment-client/src/main/java/eu/europeana/enrichment/rest/client/dereference/Dereferencer.java
jochen-vermeulen: Just so that I understand: you are suggesting to change the map type to `Map<String, Set<PermittedEntityType>>`, is that correct?

stzanakis (approved):
Build is failing because of the recent cleanup on metis-common libs.
Should be fine to merge.
Alternatively, If you want to be sure you'd have to re-base the branch.

